PR #389, opened 2014-07-27
Cause a Content-Security-Policy: sandbox header to unregister the Service Worker at that scriptURL.

jyasskin:
Joshua Peek suggested that this should work
(http://lists.w3.org/Archives/Public/public-webappsec/2014Jul/0109.html)
because `sandbox` gives the resource a unique origin, which combines with
Service Workers' same-origin policy to disallow execution.

See slightlyoff/ServiceWorker#113 and slightlyoff/ServiceWorker#224.


slightlyoff:
Do we have a decision here?

annevk:
I don't think we have sandboxing for workers yet or decided what that would mean. As far as unregistering goes I think we decided upon something else, no?

PR #616, opened 2015-02-04
Expanded on the update() Note:

jeffposnick:
@jakearchibald @slightlyoff:

Here's my take on elaborating the `update()` Note, to clarify that, by default, the browser's HTTP cache is consulted for an unexpired copy of the service worker script, and that `update()` skips the HTTP cache.

I hadn't previously been clear on whether it was the browser HTTP cache that was consulted or some service worker-specific cache—the impression that I got was that service workers scripts were always cached for 24 hours, and the only way to achieve a quicker cache expiration was to call `update()`.

### DIFF
--- a/spec/service_worker/index.html
+++ b/spec/service_worker/index.html
@@ -336,7 +336,7 @@ interface <dfn id="service-worker-registration-interface" title="ServiceWorkerRe
     <spec-section id="service-worker-registration-update">
       <h1><code>update()</code></h1>
 
-      <p class="note"><code>update()</code> pings the server for an updated version of this script without consulting caches. This is conceptually the same operation that UA does maximum once per every 24 hours.</p>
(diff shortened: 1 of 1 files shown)


jakearchibald:
> When loading a registered service worker script, the UA normally checks the HTTP cache for an unexpired copy

This is done when loading a client request occurs, rather than when the service worker script is loaded. Eg, two controlled document loads would cause two updates, even if the SW didn't shut down in between, therefore only loaded once.

> `update()`'s behavior is conceptually the same as what the UA does a maximum once every 24 hours on its own, to prevent service worker scripts from being cached for longer than a day.

Can't decide if "maximum" is misleading. I mean, the UA can check for updates more than once every 24hrs.

jungkees:
@jeffposnick thanks for having suggested a better wording. Recently after the latest f2f, it has been sorted out that update() obeys the same browser cache rule as other update paths: https://github.com/slightlyoff/ServiceWorker/issues/514#issuecomment-130491914. And the note has been removed from the text as such.

For
> I hadn't previously been clear on whether it was the browser HTTP cache that was consulted or some service worker-specific cache

The fetched service worker script resource is stored apart from the browser cache and has no expiration in and of itself until it's replaced by a new version through one of the update paths.

PR #663, opened 2015-03-26
Remove "parsed" state mention

ferjm:
`parsed` is no longer listed as a possible state at http://www.w3.org/TR/service-workers/#service-worker-state


jungkees:
_parsed_ is an internal state that cannot be accessed from the script surface. That is to say, `ServiceWorker.state` does never return "parsed". Thanks for reviewing the spec.

PR #686, opened 2015-04-22
You don't need any flags in Chrome today :)

karolklp:


karolklp:
@slightlyoff  merge me :)

PR #733, opened 2015-08-17
Mention self.skipWaiting() instead of event.replace()

zuzusik:
As far as I understand, `event.replace()` is abandoned in favor of `self.skipWaiting()`


jakearchibald:
Thank you!

Issue #684, opened 2015-04-22
Interception of the fallback requests initiated from a SW for cross-origin resources


jungkees:
While discussing use cases for [`navigator.connect()`](https://github.com/mkruisselbrink/navigator-connect/blob/gh-pages/explainer.md), we've encountered a case where interception of the fallback requests to a third-party resources and serving them through a cross-origin SW would greatly help the overall resource management. Here's the use case and the proposed extension API that enables the communication between a _client_ SW and a _service_ SW: 
https://github.com/mkruisselbrink/navigator-connect/blob/gh-pages/explainer.md#solving-the-fonts-problem

A _service_ opts in to handle _clients_' fallback requests by calling `e.handleFallThroughRequests` from within its `oninstall` event listener. Then, the requests initiated from a _client_ SW are intercepted, and fetch events tagged with `e.isFallThrough` set to true are dispatched to the _service_ SW instead of going out to the network.

It was originally discussed as part of the `navigator.connect()` spec, but agreed to be proposed as an extension to the SW API (v2): https://github.com/mkruisselbrink/navigator-connect/issues/28

Comments would be appreciated.

benfrancis:
I'd be interested to hear peoples' thoughts on whether another use case for this could be for a cross-origin Service Worker to provide an HTTP web service which works even when offline https://github.com/mkruisselbrink/navigator-connect/issues/23

This has been discussed as a use case on Mozilla's dev-webapi mailing list https://groups.google.com/forum/#!msg/mozilla.dev.webapi/2NWlXa9SlqY/mQ0zBI1SChUJ

This could provide an avenue to replace many of Gecko's current proprietary APIs with potentially more webby solutions.

jakearchibald:
You could have:

```js
// https://foo/sw.js
self.oninstall = event => {
  event.waitUntil(
    event.registerExternal('//font-service/sw.js')
  );
};
```

`//font-service/sw.js` would require a `Service-Worker-Scope` header, which would be the scope on `//font-service` that it would be registered to.

```js
// https://font-service/sw.js
self.oninstall = event => {
  event.handleFallThroughRequests(['/resources', '/font']);
  // …
};
```

URLs passed to `handleFallThroughRequests` must be within the SW's scope.

My questions/concerns about this model:

* The number of workers that may be needed to handle requests to a single page - is this a problem?
* How/when will `//font-service/sw.js` be checked for updates?

mkruisselbrink:
[...]

PR #751, opened 2015-09-22
Foreign fetch

mkruisselbrink:
Sorry for the delay, here is my foreign fetch pull request for issue #684 again, with some of the changes @jungkees made to it already applied on top.
This updates the [spec](http://mkruisselbrink.github.io/ServiceWorker/spec/service_worker/) to add support for foreign fetch, and adds a separate [foreign fetch explainer document](https://github.com/mkruisselbrink/ServiceWorker/blob/foreign-fetch/foreign_fetch_explainer.md).

### DIFF
--- a/spec/service_worker/index.html
+++ b/spec/service_worker/index.html
@@ -1449,23 +1449,28 @@
         <li><a href="#wait-until-method"><span class="secno">4.4.1</span> <span class="content"><code class="idl"><span>event.waitUntil(f)</span></code></span></a>
        </ol>
       <li>
-       <a href="#fetch-event-section"><span class="secno">4.5</span> <span class="content"><code class="idl"><span>FetchEvent</span></code></span></a>
+       <a href="#install-event-section"><span class="secno">4.5</span> <span class="content"><code class="idl"><span>InstallEvent</span></code></span></a>
        <ol class="toc">
-        <li><a href="#fetch-event-request"><span class="secno">4.5.1</span> <span class="content"><code class="idl"><span>event.request</span></code></span></a>
-        <li><a href="#fetch-event-clientid"><span class="secno">4.5.2</span> <span class="content"><code class="idl"><span>event.clientId</span></code></span></a>
-        <li><a href="#fetch-event-isreload"><span class="secno">4.5.3</span> <span class="content"><code class="idl"><span>event.isReload</span></code></span></a>
-        <li><a href="#fetch-event-respondwith"><span class="secno">4.5.4</span> <span class="content"><code class="idl"><span>event.respondWith(r)</span></code></span></a>
+        <li><a href="#register-foreign-fetch-method"><span class="secno">4.5.1</span> <span class="content"><code class="idl"><span>event.registerForeignFetch(options)</span></code></span></a>
        </ol>
       <li>
-       <a href="#extendablemessage-event-section"><span class="secno">4.6</span> <span class="content"><code class="idl"><span>ExtendableMessageEvent</span></code></span></a>
+       <a href="#fetch-event-section"><span class="secno">4.6</span> <span class="content"><code class="idl"><span>FetchEvent</span></code></span></a>
        <ol class="toc">
-        <li><a href="#extendablemessage-event-data"><span class="secno">4.6.1</span> <span class="content"><code class="idl"><span>event.data</span></code></span></a>
-        <li><a href="#extendablemessage-event-origin"><span class="secno">4.6.2</span> <span class="content"><code class="idl"><span>event.origin</span></code></span></a>
-        <li><a href="#extendablemessage-event-lasteventid"><span class="secno">4.6.3</span> <span class="content"><code class="idl"><span>event.lastEventId</span></code></span></a>
-        <li><a href="#extendablemessage-event-source"><span class="secno">4.6.4</span> <span class="content"><code class="idl"><span>event.source</span></code></span></a>
-        <li><a href="#extendablemessage-event-ports"><span class="secno">4.6.5</span> <span class="content"><code class="idl"><span>event.ports</span></code></span></a>
+        <li><a href="#fetch-event-request"><span class="secno">4.6.1</span> <span class="content"><code class="idl"><span>event.request</span></code></span></a>
+        <li><a href="#fetch-event-clientid"><span class="secno">4.6.2</span> <span class="content"><code class="idl"><span>event.clientId</span></code></span></a>
+        <li><a href="#fetch-event-isreload"><span class="secno">4.6.3</span> <span class="content"><code class="idl"><span>event.isReload</span></code></span></a>
+        <li><a href="#fetch-event-respondwith"><span class="secno">4.6.4</span> <span class="content"><code class="idl"><span>event.respondWith(r)</span></code></span></a>
        </ol>
-      <li><a href="#execution-context-events"><span class="secno">4.7</span> <span class="content">Events</span></a>
+      <li>
+       <a href="#extendablemessage-event-section"><span class="secno">4.7</span> <span class="content"><code class="idl"><span>ExtendableMessageEvent</span></code></span></a>
+       <ol class="toc">
+        <li><a href="#extendablemessage-event-data"><span class="secno">4.7.1</span> <span class="content"><code class="idl"><span>event.data</span></code></span></a>
+        <li><a href="#extendablemessage-event-origin"><span class="secno">4.7.2</span> <span class="content"><code class="idl"><span>event.origin</span></code></span></a>
+        <li><a href="#extendablemessage-event-lasteventid"><span class="secno">4.7.3</span> <span class="content"><code class="idl"><span>event.lastEventId</span></code></span></a>
+        <li><a href="#extendablemessage-event-source"><span class="secno">4.7.4</span> <span class="content"><code class="idl"><span>event.source</span></code></span></a>
+        <li><a href="#extendablemessage-event-ports"><span class="secno">4.7.5</span> <span class="content"><code class="idl"><span>event.ports</span></code></span></a>
+       </ol>
+      <li><a href="#execution-context-events"><span class="secno">4.8</span> <span class="content">Events</span></a>
      </ol>
     <li>
      <a href="#cache-objects"><span class="secno">5</span> <span class="content">Caches</span></a>
@@ -1540,6 +1545,7 @@
       <li><a href="#run-service-worker-algorithm"><span class="secno"></span> <span class="content">Run Service Worker</span></a>
       <li><a href="#terminate-service-worker-algorithm"><span class="secno"></span> <span class="content">Terminate Service Worker</span></a>
       <li><a href="#on-fetch-request-algorithm"><span class="secno"></span> <span class="content">Handle Fetch</span></a>
+      <li><a href="#on-foreign-fetch-request-algorithm"><span class="secno"></span> <span class="content">Handle Foreign Fetch</span></a>
       <li><a href="#handle-functional-event-algorithm"><span class="secno"></span> <span class="content">Handle Functional Event</span></a>
       <li><a href="#on-client-unload-algorithm"><span class="secno"></span> <span class="content">Handle Service Worker Client Unload</span></a>
       <li><a href="#on-user-agent-shutdown-algorithm"><span class="secno"></span> <span class="content">Handle User Agent Shutdown</span></a>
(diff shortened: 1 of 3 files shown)


spec/service_worker/index.html
```diff
@@ -1372,6 +1373,48 @@ <h2 id="status">Status of This Document</h2>
...
+            <li>Let <var>subScopeString</var> be the <a href="https://url.spec.whatwg.org/#concept-url-serializer">serialized</a> <var>subScopeURL</var>.</li>
+            <li>If <var>subScopeString</var> does not start with <var>scopeString</var>, throw a <code>TypeError</code> and abort these steps.</li>
+            <li>Add <var>subScopeURL</var> to <var>subScopeURLs</var>.</li>
+          </ol>
+        </li>
+        <li>Set this <a href="#dfn-service-worker">service worker</a>'s <a href="#dfn-foreign-fetch-scopes">list of foreign fetch scopes</a> to <var>subScopeURLs</var>.</li>
+      </ol>
+      </spec-algorithm>
+
+      <p class="fixme">Maybe subscopes should be parsed using the registrations scope URL instead of the workers API base url. Although that might actually be more surprising/confusing.</p>
```
mkruisselbrink: Any opinions on this? Since we require foreign fetch scopes to be "inside" the scope of the service worker it superficially seems to maybe make sense to parse relative urls relative to the scope. But parsing relative urls relative to the location of the sw script also seems like it would make sense, and might be less surprising.

spec/service_worker/index.html
```diff
@@ -3589,6 +3632,148 @@ <h2 id="status">Status of This Document</h2>
     </spec-algorithm>
   </spec-section>
 
+  <spec-section id="foreign-fetch-scope-match-algorithm">
+    <h1>Match Service Worker for Foreign Fetch</h1>
```
mkruisselbrink: I'm not entirely convinced that the behavior I described here is the behavior we want. In particular I wonder what would be less surprising behavior in the following situations:

- Two SW registrations, SW1 with scope "/", and SW2 with scope "/subscope/"
- SW1 registers for foreign fetch with scope "/"
- Do cross origin resource requests to "/subscope/foo/bar" get intercepted by SW1?
- SW2 now registers for foreign fetch with scope "/subscope/"
- Now which worker gets to handle foreign fetch events for "/subscope/foo/bar"?
- A new version of SW1 now registers for foreign fetch with scope "/subscope/foo"
- Now where do foreign fetch events for "/subscope/foo/bar" go?

In my proposed algorithm as long as SW2 is around foreign fetch events for "/subscope/foo/bar" will never go to SW1. I think that might be the easiest to understand, except in the last case it does mean that SW1 effectively registered for foreign fetch with a scope that can never receive events as long as SW2 is around, which might be surprising.

spec/service_worker/index.html
```diff
@@ -3589,6 +3632,148 @@ <h2 id="status">Status of This Document</h2>
...
+          <li>If <var>requestString</var> starts with <var>scopeString</var> return <var>worker</var>.</li>
+        </ol>
+      </li>
+      <li>Return null.</li>
+    </ol>
+    </spec-algorithm>
+  </spec-section>
+
+  <spec-section id="on-foreign-fetch-request-algorithm">
+    <h1>Handle Foreign Fetch</h1>
```
mkruisselbrink: It would be nice if somehow some of the duplication between this and the regular fetch event handling could be shared. But maybe they will end up being too different for that not to matter much?

spec/service_worker/index.html
bsittler: I think your proposal is probably the saner resolution. This way foreign fetch interception is strictly subsidiary to same-origin fetch interception interception and it's always trivial to share handler logic for requests where the calling origin is not important.

bsittler:
I think SW-relative is much more sensible and makes it easier to ensure registrations match for foreign and same-origin interceptions for the same URL in cases where the logic will in fact be identical.

spec/service_worker/index.html
jungkees: +1 to using settings object's API base URL. Having a consistency with other APIs seems better for devs so they don't have to look up a document for this usage.

spec/service_worker/index.html
jungkees: > This way foreign fetch interception is strictly subsidiary to same-origin fetch interception interception and it's always trivial to share handler logic for requests

Agreed. I think using Match Service Worker Registration in Match Service Worker for Foreign Fetch to organize the relationship of those multiple registrations makes good sense. So the behavior in the example will be like the following, right?

> Do cross origin resource requests to "/subscope/foo/bar" get intercepted by SW1?

No. It'll just fallback to the network again when returning to fetch.

> SW2 now registers for foreign fetch with scope "/subscope/"
> Now which worker gets to handle foreign fetch events for "/subscope/foo/bar"?

SW2 will do.

>A new version of SW1 now registers for foreign fetch with scope "/subscope/foo"
> Now where do foreign fetch events for "/subscope/foo/bar" go?

SW1 will still do.

spec/service_worker/index.html
jungkees: Let's see how we can sort this out with fetch when we settle on the desirable behavior.

wanderview:
FYI, I intend to review this before TPAC, but likely won't get to it for a couple weeks.

mkruisselbrink:
I updated the [spec](http://mkruisselbrink.github.io/ServiceWorker/spec/service_worker/) to add a (poorly spec-ed) 'withCredentials' option to foreign fetch (sort of as briefly discussed at TPAC). I also added a separate [foreign fetch explainer document](https://github.com/mkruisselbrink/ServiceWorker/blob/foreign-fetch/foreign_fetch_explainer.md), since that seemed potentially helpful.

mkruisselbrink:
@wanderview or @annevk (or anybody else): any comments on this? Things that I should change/improve etc? The currently suggested spec is definitely not perfect, but I'm not entirely sure how to better spec the credentials part (and if what it is trying to say is even the correct behavior, or if there is more we could/should do minimize the risks of badly written service workers exposing data they didn't intend to expose).

annevk:
1. I think it's a very bad idea to reuse the `fetch` event. I don't think we should do that.
2. We shouldn't use `withCredentials` since that is XMLHttpRequest terminology.
3. If the request is without credentials, this also needs to affect the used service worker in some way. Otherwise there's still ambient authority leakage.

annevk:
@ehsan you probably want to look at the links in https://github.com/slightlyoff/ServiceWorker/pull/751#issuecomment-161458787 to review foreign fetch.

mkruisselbrink:
> I think it's a very bad idea to reuse the fetch event. I don't think we should do that.

Okay, changed it to use a separate foreignfetch event (although it was my (apparently mistaken) understanding from our conversation at TPAC that using the same event would probable be okay). I don't have a strong opinion on same vs. different event, and most people seem to have at least the initial reaction that this should be a separate event, so making it a separate event is probably the way to go.

> We shouldn't use withCredentials since that is XMLHttpRequest terminology.

Agreed; not sure what the best name is. Things like "dontOmitCredentials" would be even worse (with the double negation etc), but I guess naming this also somewhat depends on what exactly it is going to do.

> If the request is without credentials, this also needs to affect the used service worker in some way. Otherwise there's still ambient authority leakage.

Agreed here as well, although unfortunately I'm not quite familiar enough with all the details/intricacies around the fetch spec, cors and authority leakage to really come up with what exactly the desired effect here is going to be. Hopefully something we can discuss at the SW F2F.

annevk:
So:

1. The service worker itself is always fetched with credentials included.
1. Any foreign fetch from A to B may or may not have credentials included.
  1. Whether credentials are included is exposed on the `Request` class, though in a somewhat confusing way in the context of foreign fetch. E.g., what does "same-origin" mean?
  1. Although A makes the foreign fetch, only B can turn this into a network fetch. This means that to B's server, it looks like a same-origin fetch.

Perhaps the answer is that B cannot give basic responses to A. Only responses that are synthetic (generated by a service worker that knows this was a foreign fetch due to the new event), CORS responses that approve of A in some manner, or responses that are turned into CORS responses through some new kind of API. This should protect B's service worker and server quite a bit.

mkruisselbrink:
I updated the explainer with what I think were the resolutions from the F2F. Also updated the spec to get rid of the withCredentials flag, and instead have a list of origins when registering for foreign fetch.
I didn't specify any kind of `makeVisible` method yet, but did change it to now turn all responses in opaque responses.

spec/service_worker/index.html
```diff
@@ -1478,6 +1485,59 @@ <h2 id="status">Status of This Document</h2>
...
+      <ol>
+        <li>If the <a href="https://dom.spec.whatwg.org/#dispatch-flag">dispatch flag</a> is unset, then:
+          <ol>
+            <li><a href="http://heycam.github.io/webidl/#dfn-throw">Throw</a> an "<code><a href="http://heycam.github.io/webidl/#invalidstateerror">InvalidStateError</a></code>" exception.</li>
+            <li>Abort these steps.</li>
+          </ol>
+        </li>
+        <li>If <var>origins</var> is empty <a href="http://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>TypeError</code> and abort these steps.</li>
+        <li>Let <var>originURLs</var> be an empty list of <a href="https://url.spec.whatwg.org/#concept-url">URLs</a>.</li>
+        <li>If the value of <var>origins</var> is not a single string equal to a single U+002A ASTERISK character (*):
```
mkruisselbrink: Question: Should the valid values for the origin parameter be `'*'`, `['*']`, `'https://some.origin/'`, `['https://some.origin', 'https://other.origin']` or should `['*']` not be in that list? And more in general I'm not sure how to best deal with a string-or-sequence-of-strings paremeter in spec algorithms. I guess adding something like "if origins is a string, set origins to a list containing origins` might do it, but is probably not the best/clearest phrasing.

annevk: 1. Make `registerForeignFetchScopes()` accept a dictionary. That'll be better going forward.
1. Both _subScopes_ and _origins_ should just take a sequence I think. No need to get fancy. Wrapping the string in [ and ] is not hard.

mkruisselbrink: > Make registerForeignFetchScopes() accept a dictionary. That'll be better going forward.

Not sure I agree with this. Sure, every method in every API could just have a single dictionary argument with possibly required fields in the dictionary, but that is not how most APIs are written. Many APIs do have an optional dictionary parameter with optional extra arguments, where I fully agree that a dictionary is the way to go, I just don't think that here with an API where the arguments are in no way interchangeable (no subscope should be a valid origin, and no origin would be a valid subscope), and are always required would really be better if it used a dictionary.

> Both subScopes and origins should just take a sequence I think. No need to get fancy. Wrapping the string in [ and ] is not hard.

Okay, made that change.

jungkees:
- Handle Foreign Fetch step 11.1 (in html file):
```
s/<var>originMatches to true.</var>/<var>originMatches</var> to true./
```

- It seems better to place _Handle Foreign Fetch_ right after _Handle Fetch_ and _Match Service Worker for Foreign Fetch_ right after _Match Service Worker Registration_.

spec/service_worker/index.html
annevk: When you already have two arguments and will likely add more in the future it seems better to just have a dictionary from the start since the eventual code will become more readable.

mkruisselbrink: > When you already have two arguments and will likely add more in the future it seems better to just have a dictionary from the start since the eventual code will become more readable.

Okay, changed it to a dictionary.

mkruisselbrink:
> Handle Foreign Fetch step 11.1 (in html file):
> s/<var>originMatches to true.</var>/<var>originMatches</var> to true./

Done

> It seems better to place Handle Foreign Fetch right after Handle Fetch and Match Service Worker for Foreign Fetch right after Match Service Worker Registration.

Done

mkruisselbrink:
Unless somebody has strong objections against it, I think I'd like to go ahead and merge this change (and iterate/fix things as they come up afterwards). So, any objections?

jungkees:
> merge this change (and iterate/fix things

+1

annevk:
So, makeVisible is not defined yet, there are no issues filed against Fetch for it, what's the story there? How do we want to approach that and who is going to maintain the security boundary?

mkruisselbrink:
> So, makeVisible is not defined yet, there are no issues filed against Fetch for it, what's the story there? How do we want to approach that and who is going to maintain the security boundary?

Yes, that's correct. Currently foreign fetch is specced to always return responses as opaque filtered responses, which of course isn't very useful, so this is definitely something that still needs to be addressed.

I think there are still some open questions/concerns around the exact shape of some kind of makeVisible API, so my plan was to file a separate (service worker) issue for that after landing this change (it seemed odd to file issues for features not technically part of the spec yet), to at least make sure it's well documented how/why we ended up with whatever makeVisible style API we'll end up with.

annevk:
I recommend filing all deferred issues now and have them link here before landing. That way it's much more clear what is going on (and you could have avoided my question).

mkruisselbrink:
Okay, filed #841 for the makeVisible part and whatwg/fetch#223 to add a step to fetch to call handle foreign fetch.

PR #809, opened 2016-01-06
It notes that this is very tightened to the use of promises

titoBouzout:
Aim to alert developers and people reading that they must understand the use of Promises. 

App Cache may be a douchebag as you note here, but at least it does not force developers to new programming concepts.


jungkees:
Apologies it was so late. The changes have been manually merged in https://github.com/w3c/ServiceWorker/commit/31d1543b1e0d0cbfb0a14d950745f7a1ab5eef78 to avoid conflicts against other PRs in time. Thanks for your contribution!

PR #818, opened 2016-01-15
Use “service worker” consistently

mathiasbynens:
To reduce confusion, this patch also ensures that sentence case is used consistently for titles.

Note: this only affects the Markdown files.

Ref. https://github.com/slightlyoff/ServiceWorker/issues/705.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,22 @@
-## What's Going On Here?
+## What’s going on here?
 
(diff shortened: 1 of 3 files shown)


mathiasbynens:
*taps mic*

Is this thing on?

annevk:
Seems I still have some power here to make things happen.

annevk:
Thanks @mathiasbynens for actually making it happen, of course.

Issue #685, opened 2015-04-22
A header-based SW installation


jungkees:
It's been discussed as part of the `navigator.connect()` and decided to be proposed as an extension to SW API (v2).

Assuming a communication between a _client_ SW and a _service_ SW is enabled through the efforts in `navigator.connect()`'s [persistent channel messaging](https://github.com/mkruisselbrink/navigator-connect/blob/gh-pages/explainer.md#inter-app-rpc) and the [fallThrough requests proposal](https://github.com/slightlyoff/ServiceWorker/issues/684), the _client_ SWs would need to use readily installed _service_ SWs whenever possible.

Here's a proposal to allow _services_ to install their SWs using HTTP headers when any of their resources are requested: https://github.com/mkruisselbrink/navigator-connect/blob/gh-pages/explainer.md#mommy-where-do-service-workers-come-from

jakearchibald:
You could have:

```js
// https://foo/sw.js
self.oninstall = event => {
  event.waitUntil(
    event.registerExternal('//font-service/sw.js')
  );
};
```

`//font-service/sw.js` would require a `Service-Worker-Scope` header, which would be the scope on `//font-service` that it would be registered to.

```js
// https://font-service/sw.js
self.oninstall = event => {
  event.handleFallThroughRequests(['/resources', '/font']);
  // …
};
```

URLs passed to `handleFallThroughRequests` must be within the SW's scope.

My questions/concerns about this model:

* The number of workers that may be needed to handle requests to a single page - is this a problem?
* How/when will `//font-service/sw.js` be checked for updates?

jakearchibald:
On #684 I pitch the idea of:

```
// https://foo/sw.js
self.oninstall = event => {
  event.waitUntil(
    event.registerExternal('//font-service/sw.js')
  );
};
```

This would only require `'//font-service/sw.js` to have a header indicating scope. It doesn't allow page/subresource headers to register a ServiceWorker, but is there any benefit in doing this?

mkruisselbrink:
I don't really like the idea of having in this case the `foo` service worker specify the exact path to the `font-service` service worker script. It seems to me that conceptually the script URL of a service worker should be an implementation detail of that site, and not part of its public API. But maybe that's just my mental model being wrong. 

[...]

PR #828, opened 2016-02-03
Specify serviceworker link type for link element and link header.

mkruisselbrink:
This is my [attempt](http://mkruisselbrink.github.io/ServiceWorker/spec/service_worker/#link-type-serviceworker) to specify the serviceworker link element and header type to address issue #685.

### DIFF
--- a/spec/service_worker/index.bs
+++ b/spec/service_worker/index.bs
@@ -49,6 +49,7 @@ spec: dom
         text: Event
         text: EventTarget
 
+spec: html; type: element; text: link
 spec: html; type: dfn
     text: about:blank
     text: active document
@@ -58,6 +59,7 @@ spec: html; type: dfn
     text: auxiliary browsing context
     text: browsing context
     text: discard a document
+    text: document base url
     text: effective script origin
     text: entry settings object
     text: environment settings object
@@ -203,6 +205,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: StructuredCloneWithTransfer
         urlPrefix: interaction.html
             text: has focus steps
+        urlPrefix: semantics.html
+            text: external resource link
         urlPrefix: webappapis.html
             text: classic script
             text: creation url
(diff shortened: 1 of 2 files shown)


mkruisselbrink:
Currently I wrote this to only trigger a register when a <link rel=serviceworker> element is inserted into a document, but maybe it makes more sense to be more consisted with other link elements, and also trigger whenever one of the attributes changes?

And thinking about this more, I need to somehow modify the Register algorithm to make it work without a client, while still keeping the origin checks that are currently done based on the client.

annevk:
@mkruisselbrink you should probably figure out what bits the algorithm needs from client and require those to be passed in, instead.

mkruisselbrink:
> @mkruisselbrink you should probably figure out what bits the algorithm needs from client and require those to be passed in, instead.

Yeah, it seems the client is used for two (well, currently three) things:

1. the origin is needed to make sure the script/scope are on the origin as the client that called register
2. the full client is used to set the client property of the fetch request that retrieves the worker script
3. Issue #829, the client is for some reason associated with the returned `ServiceWorkerRegistration`

This third one I'll ignore for now, since it's broken anyway. For the second one it is probably correct to have a null client, just like in some update checks. So that leaves the first one; either I can add the "client" origin as a separate field for a job (which would only be used by register jobs), or maybe it makes sense to move those checks to before the job is scheduled. It's not entirely clear to me why some origin checks are done before a register job is created, and some are done when the register job is executed. It seems like it should be fine to do all parameter validation before potentially scheduling a register job.

mkruisselbrink:
Okay, I changed it to move the origin checks to be done before a job is created. Also made the link element processing more in line with how other specs process link elements.

One thing I was wondering is if processing of a Link: header should somehow make sure that the response on which the header is set is the response to a request that originated in a secure context. I think having such a check would make sense, but specifying it is slightly tricky since the whole Link: header processing isn't very precisely defined currently. Ideally there would just be a step somewhere in the fetch spec where Link: headers are processed, at which point the algorithm would have access to the client associated with the request. Maybe currently it's enough to be hand-wavy about where such a client might come from.

annevk:
We could definitely add a hook in Fetch to make it more explicit that's what's happening.

wanderview:
Gecko bug to implement: https://bugzilla.mozilla.org/show_bug.cgi?id=1247141

spec/service_worker/index.bs
```diff
@@ -1849,6 +1843,74 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
...
+      </pre>
+    </div>
+  </section>
+
+  <section>
+    <h3 id="link-element-interface-section">Link element interface extensions</h3>
+
+    <pre class="idl">
+      partial interface HTMLLinkElement {
+        attribute USVString scope;
```
domenic: Can you add [CEReactions] to this, per https://github.com/w3c/webcomponents/issues/186 ?

spec/service_worker/index.bs
```diff
@@ -1849,6 +1843,74 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
...
+      };
+    </pre>
+
+    <p>The <dfn id="link-scope-attribute">scope</dfn> attribute returns the value of this element's <dfn element-attr for="link">scope</dfn> attribute.</p>
```
domenic: This should use reflection instead: https://html.spec.whatwg.org/multipage/infrastructure.html#reflect

PR #835, opened 2016-02-16
Convert spec to bikeshed

mkruisselbrink:
I'm fairly confident about the accuracy of this conversion, but a few comments:
- (Obviously) I fixed all fatal errors bikeshed reported while processing the spec. This fixed a few links that were broken in the old spec (yay bikeshed checking internal links for validity)
- I ran the resulting HTML (at https://mkruisselbrink.github.io/ServiceWorker/spec/service_worker_1/) through the W3C link checker; all remaining errors there seem to be pretty benign.
  - In fixing these I discovered a number of broken links to concepts in html and/or fetch that need fixing. I filed bugs for some, and added all of them to the "Ignored Terms" section in the bikeshed metadata, which should make it relatively straight forward to figure out what things need fixing.
- I diffed the list of IDs defined by the bikeshed version with those in the pre-bikeshed version. A few things stand out from this:
  - sections that are auto-generated by bikeshed have different IDs (so #about, #dependencies and #introduction no longer exist)
  - the pre-bikeshed version had unique IDs auto-generated for every single paragraph in the spec (of the form "header-id.1", "header-id.2", etc. Bikeshed doesn't generate these IDs and thus these IDs no longer exist and any links to them would be broken. No idea if anybody actually linked directly to a specific paragraph. Fixing this would require implementing this functionality in bikeshed, but I personally don't think doing so is necessary/worth it.
- Some references now link to equivalent definitions in a different spec (e.g. before some worker references pointed to the w3c workers spec, while other pointed to whatwg html5. Now all point to the whatwg version. Some other references we pointed to the html5 spec pointing to some other spec, now those mostly point directly to the spec referenced by html5, maybe some more I forgot).
- Chrome has a bug (http://crbug.com/431867) which causes it to incorrectly render the `<th>` elements with the default bikeshed stylesheet (text is centered despite the stylesheet requiring left/start text-align). Really only a cosmetic issue with tables in the spec.
- ~~Bikeshed has a bug (tabatkins/bikeshed#595) that results in bikeshed crashing on this spec. So you'll need to apply tabatkins/bikeshed#596 to actually be able to generate HTML.~~
- I think I'd prefer to just squash all these commits together and push the resulting commit rather than merging this pull request as is, since the individual commits aren't terribly meaningful (most of them result in bikeshed failing with fatal errors). 
- There is still room for improvement in how things are dealt with within the spec, but I think this conversion is good enough to replace the spec and any other improvements can be made later.

@jungkees not sure how carefully you want to review this/how much you trust I didn't make any mistakes; I guess let me know if/when you want me to merge this (squashed or not).


jungkees:
> through the W3C link checker; all remaining errors there seem to be pretty benign.

Nit: 1 occurrence of `<af for="url">origin</a>` in the source files.

> added all of them to the "Ignored Terms" section in the bikeshed metadata, which should make it relatively straight forward to figure out what things need fixing.

Great. It'd be useful when fixing the related parts.

> sections that are auto-generated by bikeshed have different IDs (so #about, #dependencies and #introduction no longer exist)

#about and #dependencies got new section ids in conformance sections, so no problem. For #introduction, we might want to make the section name and the section id be consistent. I think "Motivations" is more relevant here.

> the pre-bikeshed version had unique IDs auto-generated for every single paragraph in the spec (of the form "header-id.1", "header-id.2", etc.

I don't think we need them any more.

> Some references now link to equivalent definitions in a different spec

That's fine and better for now I think. We might have to sort them out when going for a CR publication later though.

> Chrome has a bug (http://crbug.com/431867) which causes it to incorrectly render the <th> elements with the default bikeshed stylesheet

Not an issue. Let's defer to the resolution of the issue there.

> I think I'd prefer to just squash all these commits together and push the resulting commit

Agreed.

> There is still room for improvement in how things are dealt with within the spec, but I think this conversion is good enough to replace the spec and any other improvements can be made later.

Sure.

> @jungkees not sure how carefully you want to review this/how much you trust I didn't make any mistakes; I guess let me know if/when you want me to merge this (squashed or not).

LGTM. Please merge them (squashed)!

Great to have you as a co-editor. Thanks a lot!

mkruisselbrink:
Okay, fixed those nits and pushed the resulting squashed commit.

annevk:
Thank you @mkruisselbrink. This makes the specification so much easier to deal with.

PR #876, opened 2016-04-11
Change "validate the response" to "process the response" to match HTML

domenic:
See https://github.com/whatwg/html/pull/1003. DO NOT MERGE until that is merged; I will ping this thread when that happens.


annevk:
I think this should be good to merge now.

Issue #849, opened 2016-03-15
Spec uses incorrect settings object when fetching a module worker


mkruisselbrink:
In [Update](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#update-algorithm) the algorithm uses "job's client" as "settings object" passed to "Fetch a module script tree". This client is whatever context called "register" or "update", and could even be null if this is a regular "soft update" check. Fetch a module script tree then uses this settings object to find the module map in which to look up modules to fetch. Obviously when client is null that wouldn't work at all, but even when client isn't null we shouldn't reuse the module map of the client that installed the service worker, rather we should be using an module map that is owned by the service worker (similar to how regular workers have their own module map).

Other than for the module map this client is also used as the client for the requests for actual scripts, both in classic and module scripts. This effects things like the referrer of the request. Not sure what the correct thing to do there is, but the answer to that might influence to correct fix for the module map issue.

wanderview:
Also note the job's client is null when created from Soft Update.  Not sure storing the client on the job is a good idea at all.  We use its origin in some places, but otherwise its only used by the script loading for environment settings (which seems wrong as you point out).

mkruisselbrink:
Yeah, the job's client was used in a few more places (by associating it with `ServiceWorker` and `ServiceWorkerRegistation` instances), but I removed those in 08a506e9d5a25bc18dd6d16c1dfb11ef6a8914d2 because those client associations weren't actually needed. So a possible solution for this bug could indeed be to just get rid of the client of a job entirely. 

I wonder if current implementations use the client as referrer of requests to fetch a serviceworker, or if this something where implementations already don't follow the spec (looking at the chrome implementation I don't think we use the client as referrer anywhere). If spec already don't set the referrer of these requests to the client who called register/update, I don't see any reason not to make the spec match the implementations, and get rid of the client in jobs/updating entirely.

Of course we'd then still need a way to pass a module map to "fetch a module script tree" separate from the (possibly null) settings object, or create a dummy settings object just for the purpose of fetching updates.

yoshisatoyanagisawa:
Triage: still valid but too cumbersome to resolve.

PR #889, opened 2016-04-22
Don't associate a job with a client, instead just have a referrer.

mkruisselbrink:
The client of a job was used for two purposes:

1) To set as client on the requests for fetching the service worker script.

   This was replaced with setting the referrer of the request, while leaving
   the client null. This is consistent with how regular workers and shared
   workers fetch their source (with the difference that there the client of
   the request is the worker itself, and the referrer is the page/worker that
   created the worker).

   This is still somewhat problematic for module type workers though. The
   algorithm for fetching a module script tree requires a client to store
   for a module map to store the modules on (and also doesn't have a
   separate referrer, which might or might not be intentional). Not sure
   what to do here, but the previous spec was wrong anyway, since it could
   pass a null client and should be passing something representing the
   service worker. This is issue #849.

2) To verify that the client from which a service worker was registered or
   unregistered was same origin with the service worker. For this we now
   just use the origin of the job's referrer.


spec/service_worker/index.bs
```diff
@@ -3109,10 +3109,11 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
       <li>Switching on <var>job</var>'s <a>worker type</a>, run these substeps with the following options:
         <dl>
         <dt><em>"<code>classic</code>"</em></dt>
-        <dd><p><a>Fetch a classic worker script</a> given <var>job</var>’s <a>serialized</a> <a href="#dfn-job-script-url">script url</a>, <var>job</var>’s <a href="#dfn-job-client">client</a>, and "<code>serviceworker</code>".</p></dd>
+        <dd><p><a>Fetch a classic worker script</a> given <var>job</var>’s <a>serialized</a> <a href="#dfn-job-script-url">script url</a>, <var>job</var>’s <a for=job>referrer</a>, null, and "<code>serviceworker</code>".</p></dd>
         <dt><em>"<code>module</code>"</em></dt>
-        <dd><p><a>Fetch a module script tree</a> given <var>job</var>’s <a>serialized</a> <a href="#dfn-job-script-url">script url</a>, "<code>omit</code>", "<code>serviceworker</code>", and <var>job</var>’s <a href="#dfn-job-client">client</a>.</p></dd>
+        <dd><p><a>Fetch a module script tree</a> given <var>job</var>’s <a>serialized</a> <a href="#dfn-job-script-url">script url</a>, "<code>omit</code>", "<code>serviceworker</code>", and some settings object.</p></dd>
```
mkruisselbrink: @domenic any suggestions on how to best address this? For dedicated and shared workers a settings object is created for the worker before "fetch a module script tree" is invoked, and then that settings object is used to get a module map to store the fetched modules on. For service workers we generally don't create a settings object until the worker actually is started. 

Also I would somewhat expect the main script of a module worker to be fetched with the document/worker that created the worker as referrer, similar to how class workers use that as referrer. With the current spec of "Fetch a module script tree" that isn't possible/what happens though. All modules, including the main module of a worker are fetched with the worker itself as referrer.

domenic: You can't start a worker until its global scope/settings object have been created; it doesn't really make sense. Why do you think it's necessary in this case?

It sounds like there's a bug in the referrer handling though; filing an issue at whatwg/html would be much appreciated!

mkruisselbrink: > You can't start a worker until its global scope/settings object have been created; it doesn't really make sense. Why do you think it's necessary in this case?

Currently service workers do: `fetch, create settings object, start`, while other workers do: `create settings object, fetch, start`. So in both cases start happens after creating a settings object, it's just the fetching that happens differently.

The way the algorithms are currently written we don't create a settings object until after an update check has decided that a new script was indeed downloaded and we really need to start the worker. But yeah, it might actually be a lot clearer/easier to understand if we always create a settings object/global scope for the service worker before we initiate any kind of fetch/update check. If the update check fails we just throw the settings object away, and if the update check succeeded we can then reuse it to fire the install event at etc. Additionally at the end of a successful update check we can store the settings object's module map, similarly to how we currently store the workers script (and when we have to restart the same worker we can prepopulate the settings object's module map with that same module map). That would also address a couple of other places where the spec currently doesn't make sense around module type workers.

Actually rewriting things like that would be a fairly major undertaking, but probably the best way to actually make this all make sense...

> It sounds like there's a bug in the referrer handling though; filing an issue at whatwg/html would be much appreciated!

Will do.

mkruisselbrink:
Actually don't think that this was the right way to go, as jobs will still need a client.

PR #933, opened 2016-07-22
Run the latest Bikeshed

yutakahirano:
Run Bikeshed in order to keep the subsequent change clean.

### DIFF
--- a/spec/service_worker/index.html
+++ b/spec/service_worker/index.html
@@ -11,7 +11,7 @@
  *
  * Indices
  *   - .toc for the Table of Contents (<ol class="toc">)
- *     -> <span class="secno"> for the section numbers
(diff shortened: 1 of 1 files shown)


yutakahirano:
Should I run Bikeshed for /spec/service_worker_1 as well?

domenic:
It seems like some collaborators are not using the latest Bikeshed so this gets reset every other commit, which is frustrating.

annevk:
Probably @jungkees has not updated his copy (in the bikeshed directory: git pull; bikeshed update).

jungkees:
Sorry. It was me. I didn't realize I'm doing it wrong. I'll do bikeshed update before processing documents from this point on. Thanks for letting me know.

Issue #922, opened 2016-06-29
Eliminate usage of entry and incumbent settings objects


domenic:
See https://readable-email.org/list/public-script-coord/topic/multiple-globals-and-you. We'd like to stop using these in new specs. In addition, it seems to be the case that most places that use them, are not actually implemented that way in browsers.

It would be good to do an audit of this spec and see which globals are actually implemented in browsers, then update the spec to use those. In the rare cases where incumbent and entry are actually implemented as specced, it would be ideal to change implementations if possible.

For an example of how you can test implementations, see e.g. https://github.com/whatwg/html/pull/1472 and https://github.com/whatwg/html/pull/1473

jungkees:
I'd need to check them thoroughly but presume many of them actually implemented entry and incumbent. I changed `navigator.serviceWorker.register(url)` and `navigator.serviceWorker.getRegistration(url)` to use entry for parsing url at least.

For that change, I was referencing the example code in https://www.w3.org/TR/2016/WD-html51-20160503/webappapis.html#calling-scripts section the corresponding part of which in WHATWG version is now replaced with your recent patches that clarify those concepts.

Having tried the example code with `location.assign()`, it turned out Chrome and IE (didn't try with Edge yet) use entry for parsing the url and Firefox uses current. (I saw you filed the `Location` object behavior in https://github.com/whatwg/html/issues/1431)

To sort them out in SW spec, I'd want to check once again if the goal is to make them all use current instead of entry and incumbent for both parsing urls and security checks? And don't we need any implementer/developer feedback on the behavior change? (I agree we should sort them out sooner than later but somewhat worry about the impact.)

domenic:
> To sort them out in SW spec, I'd want to check once again if the goal is to make them all use current instead of entry and incumbent for both parsing urls and security checks?

Sometimes it may be more appropriate to use relevant. Boris has argued that for URL parsing in particular that is better. (Note that inside constructors, the relevant settings object of the object being created _is_ the current settings object, so we just use the latter.)

> And don't we need any implementer/developer feedback on the behavior change?

So far in this multi-globals work, it seems to not impact developers at all, since they mostly don't do weird multi-global things on purpose. Implementers have been largely indifferent as well, apart from the fact that only Firefox seems to have a perfectly-to-spec version of entry and incumbent.

[...]

PR #950, opened 2016-08-10
Tidy up global object usage

domenic:
This commit is part of the effort in #922 to eliminate the use of entry and incumbent settings objects. It mostly just changes cases where all of relevant/current/entry/incumbent are equivalent, to use relevant. It also removes mention of "default action" for events, since that is not supposed to exist, and does some other tidying in the postMessage algorithms.

There are no normative changes in this commit.


jungkees:
@domenic, thanks for the work! I went through this change and it looks good to me. I'm wondering whether you'd want me to merge this and to deal with the URL parsing changes in a new PR (after having feedback around the WPT experiment?) Or, are you planning to add those changes in this PR? Please let me know.

domenic:
Yeah, I was planning on a separate PR for any normative changes, so merging this now is good :)

PR #961, opened 2016-08-24
Change the rest of the links to the new repo

anssiko:
@jungkees This will complement https://github.com/w3c/ServiceWorker/commit/eb7b112ac7055ff6afb4bb5a97ed72c4ac6c5c23 and updates the rest of the links (`s/slightlyoff/w3c/`).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,11 +2,11 @@
 
 ## Spec and API development
 
-For the nitty-gritty of the API, the [draft W3C specification](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/) is authoritative. For implementers and developers who seek a more stable version, [Service Workers 1](https://slightlyoff.github.io/ServiceWorker/spec/service_worker_1/) is a right document with which the contributors only focus on fixing bugs and resolving compatibility issues.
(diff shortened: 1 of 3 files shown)


jungkees:
Thanks @anssiko! Could you s/issued/issues/ in "resolved issued marked needs spec" (originally mis-typed) and make it to the master branch?

anssiko:
Thanks!

Assuming you've set the `master` branch as the GitHub Pages Source (via Settings > Options > GitHub Pages), I suggest you remove the `gh-pages` branch, and note in CONTRIBUTING.md the branch that should be used for contributions.

marcoscaceres:
If we use the W3C's echidna, the spec can be auto-published to TR. We are trying to put an end to using gh-pages to see if we can make /TR/ be less garbage.

annevk:
You're basically making TR/ more confusing then, though. Since then nobody can tell when it stands for trash and when it's useful. Would've been better to use a new namespace.

marcoscaceres:
@annevk, I agree. Though canonical will be whatever Google, and other search enginesm (not the company) decide to return when people search for a spec. By killing the options available to one,  a new namespace might not matter as much because there will only be one source of truth (in a new namespace or on TRash).

PR #963, opened 2016-08-26
Parse URLs relative to the relevant settings object, not entry

### LINKED ISSUE #922: Eliminate usage of entry and incumbent settings objects

domenic:
Fixes #922. Web platform tests at https://github.com/w3c/web-platform-tests/pull/3449. Part of the overall effort in https://github.com/whatwg/html/issues/1431.

In the process, fixes some ambiguity about how to parse URLs in the Link header, as previous it parsed relative to the entry settings object, which did not exist during this algorithm.


spec/service_worker/index.bs
```diff
@@ -2125,29 +2123,31 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
       <li>Let <var>request</var> be the <a for="fetch">request</a> for which this header was received in the response.</li>
       <li>If <var>request</var>'s <a for="request">client</a> is not a <a>secure context</a>, abort these steps.</li>
       <li>Let <var>scriptURL</var> be the result of <a for="url">parsing</a> the <a>target IRI</a> of the <code>Link</code> header.</li>
-      <li>Let <var>scopeURL</var> be the "<code>scope</code>" <a>target attribute</a> of the <code>Link</code> header, or null if no such attribute is present.</li>
+      <li>Let <var>scopeURL</var> null.
```
jungkees: "be" is missing.

spec/service_worker/index.bs
```diff
@@ -2125,29 +2123,31 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
...
       <li>Let <var>promise</var> be a new <a>promise</a>.</li>
       <li>Invoke [[#start-register-algorithm]] with <var>scopeURL</var>, <var>scriptURL</var>, <var>promise</var>, <var>client</var>, <var>client</var>'s <a>creation URL</a> and <var>workerType</var>.</li>
       <li>Run the following substeps <a>in parallel</a>:
         <ol>
           <li>Wait until <var>promise</var> settles.</li>
-          <li>If <var>promise</var> rejected, <a>queue a task</a> to <a>fire a simple event</a> named <code>error</code> at the <{link}> element.</li>
-          <li>If <var>promise</var> resolved, <a>queue a task</a> to <a>fire a simple event</a> named <code>load</code> at the <{link}> element.</li>
+          <li>If <var>promise</var> rejects, <a>queue a task</a> to <a>fire a simple event</a> named <code>error</code> at the <{link}> element.</li>
+          <li>If <var>promise</var> fulfills, <a>queue a task</a> to <a>fire a simple event</a> named <code>load</code> at the <{link}> element.</li>
```
jungkees: "resolves" is wrong here? We used "resolve" in other places.

jungkees:
@domenic, thanks for sorting out the behavior and working on this PR!

LGTM % the above nits

spec/service_worker/index.bs
domenic: Yeah, resolves is wrong. https://github.com/domenic/promises-unwrapping/blob/master/docs/states-and-fates.md

domenic: I will revert this and let you do a separate cleanup if you want

jungkees: I will do.

PR #970, opened 2016-08-31
Replace e.g. with i.e.

ricea:
"e.g." means "for example". "i.e." means "in other words". In this context, "i.e." makes more sense than "e.g."


ylafon:

PR #973, opened 2016-09-07
add redirect from /SW/ to /spec/s_w/

ylafon:


ylafon:

PR #976, opened 2016-09-09
Make cache example easier to understand

lukechilds:
I personally think that:

```js
return cachedResponse || fetch(event.request);
```
 makes it a lot more obvious what's going on as opposed to:

```js
return response || fetch(event.request);
```

I had to do a double take with `response` as at first glance it looked like the request was being made twice.


ylafon:

PR #983, opened 2016-10-05
(WIP) Navigation preload

jakearchibald:
Needs review.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2889,7 +2889,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
               1. Let |preloadRequestHeaders| be |preloadRequest|'s [=request/header list=].
               1. Let |preloadResponseObject| be a new {{Response}} object and a new associated {{Headers}} object whose [=guard=] is "`immutable`".
               1. [=header list/Append=] to |preloadRequestHeaders| a new [=header=] whose [=header/name=] is \`<code>Service-Worker-Navigation-Preload</code>\` and [=header/value=] is |registration|'s [=navigation preload header value=].
-              1. Set |preloadRequest|'s [=use-service-workers value=] to "`none`".
+              1. Set |preloadRequest|'s [=service-workers mode=] to "`none`".
               1. Run the following substeps [=in parallel=]:
                   1. [=Fetch=] |preloadRequest|.
                     
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -205,7 +205,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker registration=] has an associated <dfn export>navigation preload enabled flag</dfn>. It is initially unset.
 
-    A [=/service worker registration=] has an associated <dfn export>navigation preload header value</dfn>, which is a [=DOMString=]. It is initially set to "<code>true</code>".
+    A [=/service worker registration=] has an associated <dfn export>navigation preload header value</dfn>, which is a DOMString. It is initially set to "<code>true</code>".
 
     <section>
       <h4 id="service-worker-registration-lifetime">Lifetime</h4>
@@ -435,7 +435,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     </section>
 
     <section algorithm>
-      <h4><dfn>{{ServiceWorkerRegistration/navigationPreload}}</dfn></h4>
+      <h4 id="service-worker-registration-navigationpreload"><dfn>{{ServiceWorkerRegistration/navigationPreload}}</dfn></h4>
 
       The {{navigationPreload}} attribute's getter *must* return the [=/service worker registration's=] {{NavigationPreloadManager}} object.
     </section>
@@ -745,7 +745,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 </section>
 
 <section>
-    <h3>{{NavigationPreloadManager}}</h3>
+    <h3 id="navigation-preload-manager">{{NavigationPreloadManager}}</h3>
 
     <pre class="idl">
       [SecureContext, Exposed=(Window,Worker)]
@@ -763,7 +763,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     </pre>
 
     <section algorithm>
-      <h4><dfn>{{NavigationPreloadManager/enable()}}</dfn></h4>
+      <h4 id="navigation-preload-manager-enable"><dfn>{{NavigationPreloadManager/enable()}}</dfn></h4>
 
       The {{NavigationPreloadManager/enable()}} method, when invoked, *must* return a new [=promise=] |promise| and run the following steps [=in parallel=]:
 
@@ -774,7 +774,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     </section>
 
     <section algorithm>
-      <h4><dfn>{{NavigationPreloadManager/disable()}}</dfn></h4>
+      <h4 id="navigation-preload-manager-disable"><dfn>{{NavigationPreloadManager/disable()}}</dfn></h4>
 
       The {{NavigationPreloadManager/disable()}} method, when invoked, *must* return a new [=promise=] |promise| and run the following steps [=in parallel=]:
 
@@ -785,9 +785,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     </section>
 
     <section algorithm>
-      <h4><dfn>{{NavigationPreloadManager/setHeaderValue(value)}}</dfn></h4>
+      <h4 id="navigation-preload-manager-setheadervalue"><dfn>{{NavigationPreloadManager/setHeaderValue(value)}}</dfn></h4>
 
-      The {{NavigationPreloadManager/setHeaderValue(value)}} method, when invoked, *must* return [=a new promise=] |promise| and run the following steps [=in parallel=]:
+      The <a method for="NavigationPreloadManager"><code>setHeaderValue(|value|)</code></a> method, when invoked, *must* return [=a new promise=] |promise| and run the following steps [=in parallel=]:
 
         1. Let |registration| be the [=context object=]'s associated [=/service worker registration=].
         1. If |registration|'s [=active worker=] is null, [=reject=] |promise| with a <code>TypeError</code>, and abort these steps.
@@ -796,7 +796,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     </section>
 
     <section algorithm>
-      <h4><dfn>{{NavigationPreloadManager/getState()}}</dfn></h4>
+      <h4 id="navigation-preload-manager-getstate"><dfn>{{NavigationPreloadManager/getState()}}</dfn></h4>
 
       The {{NavigationPreloadManager/getState()}} method, when invoked, *must* return [=a new promise=] |promise| and run the following steps [=in parallel=]:
 
@@ -1405,7 +1405,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     </section>
 
     <section>
-      <h4>{{FetchEvent/preloadResponse|event.preloadResponse}}</h4>
+      <h4 id="fetch-event-preloadresponse">{{FetchEvent/preloadResponse|event.preloadResponse}}</h4>
 
       <dfn attribute for="FetchEvent"><code>preloadResponse</code></dfn> attribute *must* return the value it was initialized to.
     </section>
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -129,8 +129,10 @@ spec: csp2; urlPrefix: https://w3c.github.io/webappsec-csp/2/
 
 spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
     type: dfn
+        text: append to header list; url: concept-header-list-append
         text: basic filtered response; url: concept-filtered-response-basic
         text: cancel a ReadableStream; url: concept-cancel-readablestream
+        text: cloning the request; url: concept-request-clone
```
jakearchibald: I get the feeling this is the wrong way to do this, but I'm not sure what the right way is

docs/index.bs
```diff
@@ -3608,6 +3732,29 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
...
+              <li><a>Append to header list</a> <var>preloadRequestHeaders</var> a new <a>header</a> whose <a>name</a> is `<code>Service-Worker-Navigation-Preload</code>` and <a>value</a> is <var>registration</var>'s <a>navigation preload header value</a></li>
+              <li>Set <var>preloadRequest</var>'s <a>skip-service-worker flag</a></li>
+              <li>Set <var>navigationPreload</var> to a new <a>promise</a>.</li>
+              <li>Run the following substeps <a>in parallel</a>:
+                <ol>
+                  <li>
+                    <p><a>Fetch</a> <var>preloadRequest</var>.</p>
+                    <p>To <a>process response</a> for <var>preloadResponse</var>, run these substeps:</p>
+                    <ol>
+                      <li><p>If <var>preloadResponse</var>'s <a>type</a> is "<code>error</code>", reject <var>navigationPreload</var> with a <code>TypeError</code> and terminate these substeps.</li>
```
jakearchibald: `type` isn't linking up properly, but I'm not sure what the right way to do this is.

docs/index.bs
```diff
@@ -3608,6 +3732,29 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
           <li>Set <var>registration</var> to the result of running <a href="#scope-match-algorithm">Match Service Worker Registration</a> algorithm passing <var>request</var>'s <a for="request">url</a> as the argument.</li>
           <li>If <var>registration</var> is null or <var>registration</var>'s <a href="#dfn-active-worker">active worker</a> is null, return null.</li>
           <li>Set <var>client</var>'s <a href="#dfn-service-worker-client-active-worker">active worker</a> to <var>registration</var>'s <a href="#dfn-active-worker">active worker</a>.</li>
+          <li>If <var>request</var> is a <a>navigation request</a> and <var>registration</var>'s <a>navigation preload enabled flag</a> is set, and <var>request</var>'s <a>method</a> is `<code>GET</code>`, then:
```
jakearchibald: @annevk this is the section I'm least confident in - it's fetch for the preload. I don't think redirects are a problem since the mode will be `manual`. When the redirect is sent back to the navigate algorithm it'll be associated with the *original* request, which won't have the header.

Having said that, that may be unexpected for same-origin redirects.

docs/index.bs
```diff
@@ -877,6 +890,116 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
...
+        Promise&lt;void&gt; enable();
+        Promise&lt;void&gt; disable();
+        Promise&lt;void&gt; setHeaderValue(DOMString value);
+        Promise&lt;NavigationPreloadState&gt; getState();
+      };
+    </pre>
+    <pre class="idl" id="navigation-preload-state">
+      dictionary NavigationPreloadState {
+        boolean enabled = false;
+        DOMString headerValue;
```
jakearchibald: Yea, it should be

docs/index.bs
annevk: It's likely not a problem. Should we prefix the header so that it cannot be set through script?

docs/index.bs
```diff
@@ -1714,6 +1837,7 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
       [Constructor(DOMString type, FetchEventInit eventInitDict), Exposed=ServiceWorker]
       interface FetchEvent : ExtendableEvent {
         [SameObject] readonly attribute Request request;
+        readonly attribute Promise&lt;Response?&gt; preloadResponse;
```
annevk: This is wrong if you want to resolve it with undefined.

docs/index.bs
```diff
@@ -1723,6 +1847,7 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
     <pre class="idl" id="fetch-event-init-dictionary">
       dictionary FetchEventInit : ExtendableEventInit {
         required Request request;
+        required Promise&lt;Response?&gt; preloadResponse;
```
annevk: Why would you allow null as a value?

docs/index.bs
```diff
@@ -3608,6 +3740,29 @@ spec: webidl; urlPrefix: https://heycam.github.io/webidl/
...
+              <li>Let <var>preloadRequest</var> be the result of <a>cloning the request</a> <var>request</var></li>
+              <li>Let <var>preloadRequestHeaders</var> be <var>preloadRequest</var>'s <a>header list</a></li>
+              <li>Let <var>preloadResponseObject</var> be a new {{Response}} object and a new associated {{Headers}} object whose <a>guard</a> is "<code>immutable</code>"</li>
+              <li><a>Append to header list</a> <var>preloadRequestHeaders</var> a new <a>header</a> whose <a>name</a> is `<code>Service-Worker-Navigation-Preload</code>` and <a>value</a> is <var>registration</var>'s <a>navigation preload header value</a></li>
+              <li>Set <var>preloadRequest</var>'s <a>skip-service-worker flag</a></li>
+              <li>Run the following substeps <a>in parallel</a>:
+                <ol>
+                  <li>
+                    <p><a>Fetch</a> <var>preloadRequest</var>.</p>
+                    <p>To <a>process response</a> for <var>navigationPreloadResponse</var>, run these substeps:</p>
```
jakearchibald: I was using https://fetch.spec.whatwg.org/#fetch-method has a guide, and *response* seems to come out of nowhere there too. What's the right thing to do here, and why is fetch right to do it this way?

docs/index.bs
jakearchibald: I'm trying to show that `preloadResponse` is a promise that resolves with a response or undefined. What's the right IDL in this case? "any" seems less informative.

docs/index.bs
jakearchibald: Again the intent is for this to be a promise that resolves with a response or undefined.

docs/index.bs
annevk: Ah right, I guess this is fine. I should refactor this at some point to make it more clear...

jakearchibald:
@jungkees I'm happy with this now. Could you review?

docs/index.bs
```diff
@@ -2760,6 +2847,21 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Set |registration| to the result of running <a>Match Service Worker Registration</a> algorithm passing |request|'s [=request/url=] as the argument.
           1. If |registration| is null or |registration|'s <a>active worker</a> is null, return null.
           1. If |request|'s [=request/destination=] is not {{RequestDestination/"report"}}, set |reservedClient|'s <a>active service worker</a> to |registration|'s <a>active worker</a>.
+          1. If |request| is a [=navigation request=] and |registration|'s [=navigation preload enabled flag=] is set, and |request|'s [=request/method=] is "`GET`", then:
+              1. Let |preloadRequest| be the result of [=request/cloning=] the request |request|
+              1. Let |preloadRequestHeaders| be |preloadRequest|'s [=request/header list=]
+              1. Let |preloadResponseObject| be a new {{Response}} object and a new associated {{Headers}} object whose [=guard=] is "`immutable`"
+              1. [=header list/Append=] to |preloadRequestHeaders| a new [=header=] whose [=header/name=] is "`Service-Worker-Navigation-Preload`" and [=header/value=] is |registration|'s [=navigation preload header value=]
```
jungkees: A header's [name](https://fetch.spec.whatwg.org/#concept-header-name) and [value](https://fetch.spec.whatwg.org/#concept-header-value) are also byte sequences. So, using backticks without qutotation marks would be more correct.

docs/index.bs
```diff
@@ -201,6 +201,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
 
+    A [=/service worker registration=] has an associated <dfn export>{{NavigationPreloadManager}}</dfn> object.
+
+    A [=/service worker registration=] has an associated <dfn export>navigation preload enabled flag</dfn>. It is initially unset.
+
+    A [=/service worker registration=] has an associated <dfn export>navigation preload header value</dfn>, which is a DOMString. It is initially set to "<code>true</code>".
```
jungkees: I think the type should be a [byte sequence](https://infra.spec.whatwg.org/#byte-sequence) instead of a DOMString here. Also the initial value should be \\\`\<code>true\</code>\\`.

docs/index.bs
```diff
@@ -724,6 +737,71 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
 
 <section>
+    <h3 id="navigation-preload-manager">{{NavigationPreloadManager}}</h3>
+
+    <pre class="idl">
+      [SecureContext, Exposed=(Window,Worker)]
+      interface NavigationPreloadManager {
+        Promise&lt;void&gt; enable();
+        Promise&lt;void&gt; disable();
+        Promise&lt;void&gt; setHeaderValue(DOMString value);
```
jungkees: The param type should be ByteString instead of DOMString.

docs/index.bs
```diff
@@ -724,6 +737,71 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+    <section algorithm>
+      <h4 id="navigation-preload-manager-enable"><dfn>{{NavigationPreloadManager/enable()}}</dfn></h4>
+
+      The {{NavigationPreloadManager/enable()}} method, when invoked, *must* return a new [=promise=] |promise| and run the following steps [=in parallel=]:
+
+        1. Let |registration| be the [=context object=]'s associated [=/service worker registration=].
+        1. If |registration|'s [=active worker=] is null, [=reject=] |promise| with an "{{InvalidStateError}}" exception, and abort these steps.
+        1. Set |registration|'s [=navigation preload enabled flag=].
+        1. [=Resolve=] |promise|.
```
jungkees: This should resolve with _undefined_ I guess.

docs/index.bs
jakearchibald: Damnit it @mattto mentioned this too - sorry for forgetting

jakearchibald:
@jungkees excellent review, thank you. I've updated the branch.

docs/index.bs
```diff
@@ -724,6 +737,71 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+        1. Let |registration| be the [=context object=]'s associated [=/service worker registration=].
+        1. If |registration|'s [=active worker=] is null, [=reject=] |promise| with an "{{InvalidStateError}}" exception, and abort these steps.
+        1. Unset |registration|'s [=navigation preload enabled flag=].
+        1. [=Resolve=] |promise|.
+    </section>
+
+    <section algorithm>
+      <h4 id="navigation-preload-manager-setheadervalue"><dfn>{{NavigationPreloadManager/setHeaderValue(value)}}</dfn></h4>
+
+      The <a method for="NavigationPreloadManager"><code>setHeaderValue(|value|)</code></a> method, when invoked, *must* return [=a new promise=] |promise| and run the following steps [=in parallel=]:
```
jungkees: Ah.. That's right. I got it.

docs/index.bs
```diff
@@ -2760,6 +2847,21 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Set |registration| to the result of running <a>Match Service Worker Registration</a> algorithm passing |request|'s [=request/url=] as the argument.
           1. If |registration| is null or |registration|'s <a>active worker</a> is null, return null.
           1. If |request|'s [=request/destination=] is not {{RequestDestination/"report"}}, set |reservedClient|'s <a>active service worker</a> to |registration|'s <a>active worker</a>.
+          1. If |request| is a [=navigation request=] and |registration|'s [=navigation preload enabled flag=] is set, and |request|'s [=request/method=] is "`GET`", then:
+              1. Let |preloadRequest| be the result of [=request/cloning=] the request |request|
+              1. Let |preloadRequestHeaders| be |preloadRequest|'s [=request/header list=]
+              1. Let |preloadResponseObject| be a new {{Response}} object and a new associated {{Headers}} object whose [=guard=] is "`immutable`"
```
jungkees: I think this one hasn't been addressed yet? I missed that the value of the guard is just a string, so it seems it should be "\<code>immutable\</code>".

jungkees:
@jakearchibald, LGTM with the following points!

* https://github.com/w3c/ServiceWorker/pull/983/files#diff-27b79860afe28f01aed4f1f6228367faR2853
  - It seems the comment for this line hasn't been addressed yet
* https://github.com/w3c/ServiceWorker/pull/983/files/1e58712f505366a5c260213e297d1ead43f5874a..245b62e45dc0cef0d2f30e6fb1fdec9b1fbc7c4b#diff-27b79860afe28f01aed4f1f6228367faR2855
  - Maybe you can land this with https://github.com/whatwg/fetch/pull/435 (to reference service-workers mode.) Or just keep the current text (skip-service-worker flag) for this PR and change this line in https://github.com/w3c/ServiceWorker/pull/1025.

Please include the corresponding html file when you merge it. Thank you!

docs/index.bs
jungkees: @jakearchibald, I was confused with the markdown syntax here. If "\`immutable\`" works fine, that'd be good as-is.

jakearchibald:
@jungkees the guard setting is correct I think. `"\`foo\`"` is markdown shorthand for `"<code>foo</code>"`.

I'll revert the "service-workers mode" change before I merge.

jungkees:
@jakearchibald, with the first bullet point in the previous comment, I tried to say the following sentence might have to be changed,
From:
"Let preloadResponseObject be a new Response object and a new associated Headers object whose guard is "immutable"."
To:
"Let preloadResponseObject be a new Response object associated with a new Headers object whose guard is "immutable"."

jungkees:
@jakearchibald, sorry I found some places that need editorial changes. Noticed them through reviewing the html output. Addressed them with this follow-up commit: https://github.com/w3c/ServiceWorker/commit/a6f6a32dfdaabf0417e94dc39c18fc6c4f4aa9fb.

PR #992, opened 2016-10-18
Update link to caching API

gnowoel:
The current one leads to a 404 page.


ylafon:

PR #995, opened 2016-10-24
DO NOT MERGE: using better link defaults

domenic:
Hoping to get @tabatkins help on a couple things that are still not going right, and also HTML needs to add a few more terms so we can get rid of the pre class=anchors section for HTML.

Error 1:

```
[1;33mLINK ERROR:[0m Multiple possible 'dfn' refs for 'browsing context'.
Arbitrarily chose the one in html.
If this is wrong, insert one of the following lines into a <pre class=link-defaults> block:
spec:html; type:dfn; text:browsing context
spec:html; type:dfn; for:Document; text:browsing context
```

This is confusing since I was hoping that without any for="", it would pick the normal browsing context one. It didn't, which is sad, but I vaguely remember that being a policy design. So instead I added an explicit link default. But Bikeshed is still complaining.

Error 2:

For some reason it really wants to link `<a>origin</a>` to ForeignFetchEvent's origin, despite me changing that `<dfn>` to have a for="".


domenic:
And, I can't remove the `StructuredCloneWithTransfer` entry from the anchor block, despite HTML having 

```html
<h4 id="structuredclonewithtransfer" data-dfn-type="abstract-op" data-lt="StructuredCloneWithTransfer" data-export="">2.9.3 <dfn>StructuredCloneWithTransfer</dfn> ( <var>input</var>,
  <var>transferList</var>, <var>targetRealm</var> )</h4>
```

domenic:
Similarly I can't remove the link for DOM manipulation task source despite HTML having

```html
<dfn id="dom-manipulation-task-source" data-export="">DOM manipulation task source</dfn>
```

annevk:
When there is no `for` I think you need to write `for=/` to disambiguate. It's a little bit confusing to me why it doesn't pick that by default, but I guess it's to get you to make a conscious choice.

domenic:
I've split this up into safe changes, which are #996, and broken changes, which are here in the second commit. The problems are:

- `<a>browsing context</a>` is not working despite setting a link default for it like the error message asks me to
- `<a>DOM manipulation task source</a>` is not working despite HTML exporting that term
- `<a>task queues</a>` is linking to the local definition, not HTML's, despite HTML's being marked up with no `for=""`, and the local one being marked up with `for="service worker registration"`.
- `<a>origin</a>` is linking to the local definition, not HTML's, despite HTML's being marked up with no `for=""`, and the local one being marked up with `for="ForeignFetchEvent"`.

Some other problems reported above are user error; I've edited them to reflect that.

tabatkins:
> `<a>browsing context</a>` is not working despite setting a link default for it like the error message asks me to

Oh shoot, that's just a terrible error message on my part. The first option should include a `for: /;`, like Anne suggests. Tracking in <https://github.com/tabatkins/bikeshed/issues/851>.

> `<a>DOM manipulation task source</a>` is not working despite HTML exporting that term

Ah, bleh, this is a misfire of Shepherd's auto-detection. Note the value in the db:

```
$ bikeshed refs --text="DOM manipulation task source"
status:    snapshot
for:       []
level:     1
url:       https://html.spec.whatwg.org/multipage/webappapis.html#dom-manipulation-task-source
type:      interface
export:    True
text:      DOM manipulation task source
shortname: html
normative: True
spec:      html
```

It's type "interface", not "dfn". This is because, lacking other information, Shepherd introspects on the ID, and it so happens that the "dom-foo" pattern is used to detect interfaces and other IDL types.  To fix, HTML needs to mark it up with an explicit `data-dfn-type=dfn`.

> `task queues` and `origin`

Both of these are the same thing. I've tried to help you with these before: autolinking attributes filter the possible choices (and lacking an attribute means you don't filter on that aspect), and then locals are always preferred above foreign anchors.

So in both of these cases, without a `for` value on the link, we don't filter by `for` value. A single matching anchor is found locally, so it's chosen automatically; we never see the foreign anchor at all.  This is by design and extremely intentional; without it, it would be very annoying to link to local terms sometimes (if a foreign anchor happens to also match). Having a local anchor for a given term/type almost always means you want to link to that. For the rare cases you want to get around this, you either specify the `for` value or the `spec` value, so the local anchor gets filtered away and we move on to checking foreign anchors.

domenic:
Thanks! For the last case, I still think it's surprising that putting for on the local dfn doesn't punt it down in the precedence order, when using a non-for link.

Regardless, can it be worked around with a single line in link-defaults to tell Bikeshed to prefer HTML's origin definition for any un-for'ed instances? It would be a huge shame if we had to put spec=html on all instances of origin; I was hoping link defaults could be used to set such defaults.

domenic:
To explain my surprise a bit more, I think it's strange that you could be writing a spec with `<a>origin</a>` all over the place and getting nice links to HTML. Then one day you introduce `a foo's <dfn for=foo>origin</dfn> is an <a>origin</a>` and suddenly all your previously-working HTML links switch over and the origin you just typed refers back to foo's origin.

Maybe this is because in my mental model, any for values at the dfn site must also be used at the link site, whereas it doesn't look like Bikeshed subscribes to that model.

tabatkins:
Right, that's explicitly not Bikeshed's model; autolinking is a series of filters, and you can omit any filters that aren't necessary to narrow down the set of possible refs.

That particular situation you describe can indeed be frustrating, but it also appears to be very rare, whereas people wanting to (a) provide a `for` value for disambiguating in the global namespace, while (b) just typing simple, easy autolinks in their local spec is very common, at least in specs I edit.  

In particular, being able to just say `''foo''` to link to the "foo" CSS value, because it only appears on one property in the spec, is great; same with `'desc'` to link to the "desc" descriptor for the at-rule defined in the spec, rather than always having to type `'@foo/desc'` everywhere.  Or in IDL, being able to just say {{appendChild()}} or similar, rather than always having to specify the interface.

domenic:
I think this scenario will become increasingly common as we try to integrate HTML more into the Bikeshed ecosystem. HTML defines a lot of important terms, and then specs often say "the foo's X" as a local instance of the term.

In any case, I don't think I got an answer: can I use link-defaults to override the default, so it goes to HTML's?

tabatkins:
Yes, you can. Use `for: /;` if necessary to specify that you want the one *without* a `for` value.

tabatkins:
(I just pushed some fixes a few minutes ago that make this actually work correctly; make sure you update before trying it.)

domenic:
@tabatkins could you look at the newest commit and tell me what's going wrong? Origin and task queue are still going to the wrong place, despite my declaration of them in link-defaults with `for: /;`. Similarly browsing context is not going anywhere, despite adding `for: /;`. I tested with the Bikeshed API in addition to updating all my stuff locally.

tabatkins:
I don't see either of those terms in the latest commit's link-defaults. Can you point me to which file, specifically, you're having trouble with, and which instances of the terms are linking incorrectly?

docs/index.bs
```diff
@@ -42,7 +42,12 @@ Markup Shorthands: css no
 
 <pre class=link-defaults>
 spec: dom; type: interface; text: Document
+
 spec: html; type: element; text: link
+spec: html; type: dfn; text: origin; for: /
+spec: html; type: dfn; text: task queue; for: /
+spec: html; type: dfn; text: browsing context; for: /
```
annevk: @tabatkins they are right here, no?

domenic:
@tabatkins indeed, they are in the file docs/index.bs lines 47-49, as @annevk pointed out.

Almost all instances of the terms are linking incorrectly. As an example, [line 236 in the Bikeshed file](https://github.com/w3c/ServiceWorker/pull/995/files#diff-27b79860afe28f01aed4f1f6228367faR236) says `<a>origin</a>`, but [line 1752 in the HTML file](https://github.com/domenic/ServiceWorker/blob/3a17bab47531ad52e31735b73e0a6dbe857c8c86/docs/index.html#L1752) ends up with `<a data-link-type="dfn" href="#foreignfetchevent-origin" id="ref-for-foreignfetchevent-origin-1">origin</a>`, despite the line

```
spec: html; type: dfn; text: origin; for: /
```

in link-defaults.

docs/index.bs
tabatkins: OH! I was looking at the actual last commit in the repo, not in this PR. ^_^

tabatkins:
Yes, as I keep telling you, Bikeshed first tries to link to local things, and `<a>origin</a>` is an autolink that filters for `text=origin; type=dfn;`, which successfully matches against the local `<dfn for="ForeignFetchEvent">origin</dfn>` you have in the spec.  `<a for=/>origin</a>`, or `[=/origin=]`, additionally specifies that you want the `for=/;`, which makes the local no longer match, and then it finds the foreign dfn from HTML.

domenic:
OK. I asked earlier, is there any way to use link-defaults to make `<a>origin</a>` default to HTML? And you said yes. Is that incorrect? If it's incorrect, can you please add something, e.g. "link-strong-defaults"?

tabatkins:
I said yes, not realizing that you also define "origin" locally. I don't have any facility at the moment to do link-defaulting that ignores local definitions.

domenic:
Great, filed https://github.com/tabatkins/bikeshed/issues/853 to track that request. When a feature like that is in place, I'll continue working on integrating HTML into the Bikeshed ecosystem.

tabatkins:
And now (a) Bikeshed warns if your for-less autolink selected a for-full local definition, when there's a for-less cross-spec definition available, and (b) you can use `Assume Explicit For: yes` to make for-less autolinks act like they have `for=/`.

domenic:
Let's close this as @jakearchibald's recent work has changed a lot for the better.

PR #996, opened 2016-10-24
Editorial: tidy up cross-linking

domenic:
- Removes various link defaults and anchors that are no longer necessary
- Fixes ECMAScript spec link
- Fixes "fire a simple event" to simply be "fire an event" per recent DOM spec changes
- Removes links for "trusted event" since that is longer defined; a follow-up will be needed for that cleanup in general

This is a mergeable version of #995.


jungkees:
@domenic, thanks a lot for tidying up the links. I noticed you put the second commit to https://github.com/w3c/ServiceWorker/pull/995 in which `for="/"` to browsing context links were removed again (maybe with further cleanups in the link-defaults and the anchors sections I suppose.) Let me merge this PR before I change other part of the spec today. I'll not merge https://github.com/w3c/ServiceWorker/pull/995 until you clarify whether that's mergeable.

PR #1002, opened 2016-11-03
Editorial: more link-defaults and anchors cleanup

domenic:
I extracted out some more safe changes from #995. As you can see from the .html diff, there's not much difference; it's mostly just removing anchors blocks.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -23,13 +23,6 @@ Markup Shorthands: css no
 
 <pre class=biblio>
 {
-  "promises-guide": {
-    "href": "https://www.w3.org/2001/tag/doc/promises-guide",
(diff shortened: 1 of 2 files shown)


jungkees:
@domenic, thanks for the cleanup!

Re: request's request and response's response, SW defines neither of them on its own. They actually reference fetch's Request object's request and Response object's response. So the current links are right. So that part of the anchors section can be removed too I think. I'll do that in the next commit.

annevk:
Note that you don't always want to refer to the objects. You sometimes want the concepts.

jungkees:
For the concepts, shouldn't I use for=/ ?

annevk:
Yes, if you do that then you remove those lines and nothing should error.

PR #1006, opened 2016-11-17
Add more Soft Update on functional events

### LINKED ISSUE #905: should UA check for updates on navigate when there is no fetch event?
Currently we trigger updates for events fired at the SW.  In particular, navigates always trigger updates.

Now that we allow the browser to avoid firing fetch events if there is no fetch handler, should the update still be triggered?

Per the spec, I think the answer is no.  I think for dev UX, though, we should still trigger updates.  It would be nice to be consistent across browsers here.

jungkees:
This changes Handle Fetch and Handle Functional Event to invoke Soft
Update even if the service worker doesn't have any event listener
registered for the event. (Note that the cases where the request falls
back to network still do not trigger Soft Update.)

Fixes #905.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3414,7 +3414,12 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
         </ol>
       </li>
       <li>Let <var>activeWorker</var> be <var>registration</var>'s <a href="#dfn-active-worker">active worker</a>.</li>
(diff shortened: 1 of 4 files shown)


jakearchibald:
LGTM!

mkruisselbrink:
LGTM

jungkees:
Thanks for the reviews!

PR #1008, opened 2016-11-21
Editorial: export Handle Functional Event

annevk:
Also fix a reference to HTML.


ylafon:

Issue #931, opened 2016-07-18
Does accessing ExtendableEvent's extend lifetime promises race?


jungkees:
@annevk pointed out there can be a race in accessing ExtendableEvent’s extend lifetime promises:
https://github.com/slightlyoff/ServiceWorker/commit/72315b913d8f2b0f639b668056f2f58275e7abd2#commitcomment-18283685

I think it's a fair point as now `e.waitUntil(f)` can be called async.

jakearchibald:
Feels like `Set <var>extendLifetimePromises</var> to a copy of` should go before the parallel steps, but `If the length of <var>extendLifetimePromises</var> does not equal` also needs to be in the service worker thread, so I guess a task needs to be queued for that?

jakearchibald:
F2F:

* `Set <var>extendLifetimePromises</var> to a copy of` happening before parallel doesn't seem to be an issue, since it's only ever going to be more promises, which we want to catch
* We need to look at the spec language for "waiting for all" to see if it queues a microtask

jungkees:
Having discussed with @annevk, I will first have to move the "When dispatching an event e that uses the ExtendableEvent interface" steps to right after the actual (dispatching) actions. I'll define some general wrapper that dispatches the event and run these steps, and make it invoked from Handle Fetch and Handle Functional Events.

The OP still should be addressed here.

jungkees:
@domenic suggested we use a reference count approach over the current promise-copying-and-checking one. Will try to come up with it.

jungkees:
Addressed https://github.com/w3c/ServiceWorker/issues/931#issuecomment-248233338: https://github.com/w3c/ServiceWorker/commit/c2a518b3b01dd77ad0d0b47b6eeff54b6cd34ba7.

TODO: https://github.com/w3c/ServiceWorker/issues/931#issuecomment-248570852.

PR #1010, opened 2016-11-22
Introduce Extend Service Worker Lifetime algorithm

jungkees:
This replaces the steps that monkeypatched the extension of the service
worker's lifetime with an algorithm that is explicitly invoked after the
steps that dispatch the extendable events.

This also changes to use "create an event" algorithm in DOM spec.


jungkees:
@annevk, could you review it?

docs/index.bs
```diff
@@ -168,7 +168,7 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
...
     <p>A <a href="#dfn-service-worker">service worker</a> has an associated <dfn id="dfn-service-worker-id">id</dfn> (an opaque string), which uniquely identifies itself during the lifetime of its <a href="#dfn-containing-service-worker-registration">containing service worker registration</a>.</p>
-    <p>A <a href="#dfn-service-worker">service worker</a> is dispatched a set of <dfn id="dfn-lifecycle-events">lifecycle events</dfn>, <a href="#service-worker-global-scope-install-event">install</a> and <a href="#service-worker-global-scope-activate-event">activate</a>, and <dfn id="dfn-functional-events">functional events</dfn> including <a href="#service-worker-global-scope-fetch-event">fetch</a>.</p>
+    <p>A <a href="#dfn-service-worker">service worker</a> is dispatched a set of <dfn id="dfn-lifecycle-events">lifecycle events</dfn>, <a href="#service-worker-global-scope-install-event">install</a> and <a href="#service-worker-global-scope-activate-event">activate</a>, and <dfn id="dfn-functional-events" for="">functional events</dfn> including <a href="#service-worker-global-scope-fetch-event">fetch</a>.</p>
```
annevk: This should not need `for=""`.

annevk: You should also try to migrate away from relying on `href="#..."` and instead use the linking mechanism of Bikeshed that does not require `href` but uses `lt` (or just the contents).

docs/index.bs
```diff
@@ -3173,8 +3158,10 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
       <li>Invoke <a href="#run-service-worker-algorithm">Run Service Worker</a> algorithm with <var>installingWorker</var> as the argument.</li>
       <li><a>Queue a task</a> <var>task</var> to run the following substeps:
         <ol>
-          <li>Create a trusted event <var>e</var> that uses the {{InstallEvent}} interface, with the event type <code><a href="#service-worker-global-scope-install-event">install</a></code>, which does not bubble and is not cancelable.</li>
+          <li><a>Create an event</a> <var>e</var> with the {{InstallEvent}} interface.</li>
```
annevk: I think this phrasing is a bit confusing. "Let _e_ be the result of ..." would be clearer.

docs/index.bs
```diff
@@ -3173,8 +3158,10 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
...
+          <li>Initialize <var>e</var>’s {{Event/type}} attribute to "<code>install</code>".</li>
           <li><a>Dispatch</a> <var>e</var> at <var>installingWorker</var>'s <a>environment settings object</a>'s <a for="environment settings object">global object</a> <var>globalObject</var>.</li>
```
annevk: Shouldn't this be "relevant global object" rather than "environment settings object's global object"?

docs/index.bs
```diff
@@ -3359,6 +3348,31 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
...
+        <dd><var>event</var>, an {{ExtendableEvent}} object</dd>
+      <dt>Output</dt>
+        <dd>None</dd>
+    </dl>
+    <ol>
+      <li>If <var>event</var>'s <a>extend lifetime promises</a> is empty, unset <var>event</var>'s <a>extensions allowed flag</a> and abort these steps.</li>
+      <li>Let <var>extendLifetimePromises</var> be an empty array.</li>
+      <li>Run the following substeps <a>in parallel</a>:
+        <ol>
+          <li><em>SetupPromiseArray</em>: Set <var>extendLifetimePromises</var> to a copy of <var>event</var>'s <a>extend lifetime promises</a>.</li>
```
annevk: I thought we had a better way of doing this? Just letting the promises decrease a counter.

docs/index.bs
```diff
@@ -3426,13 +3440,16 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
       <li>Invoke <a href="#run-service-worker-algorithm">Run Service Worker</a> algorithm with <var>activeWorker</var> as the argument.</li>
       <li><a>Queue a task</a> <var>task</var> to run the following substeps:
         <ol>
-          <li>Create a trusted event <var>e</var> that uses the {{FetchEvent}} interface, with the event type <code><a href="#service-worker-global-scope-fetch-event">fetch</a></code>, which does not bubble.</li>
-          <li>Let the {{FetchEvent/request}} attribute of <var>e</var> be initialized to <var>r</var>.</li>
+          <li><a>Create an event</a> <var>e</var> with the {{FetchEvent}} interface.</li>
+          <li>Initialize <var>e</var>’s {{Event/type}} attribute to "<code>fetch</code>".</li>
+          <li>Initialize <var>e</var>’s {{Event/cancelable}} attribute to true.</li>
```
annevk: Why do you make it cancelable? That seems like a change.

You're not handling the return value of the dispatch algorithm though, so how does it matter?

docs/index.bs
```diff
@@ -3509,10 +3526,13 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
       <li>Invoke [[#run-service-worker-algorithm]] algorithm with <var>activeWorker</var> as the argument.</li>
       <li><a>Queue a task</a> <var>task</var> to run the following substeps:
         <ol>
-          <li>Create a trusted event <var>e</var> that uses the {{ForeignFetchEvent}} interface, with the event type <code><a href="#service-worker-global-scope-foreignfetch-event">foreignfetch</a></code>, which does not bubble.</li>
-          <li>Let the {{ForeignFetchEvent/request}} attribute of <var>e</var> be initialized to <var>r</var>.</li>
-          <li>Let the {{ForeignFetchEvent/origin}} attribute of <var>e</var> be initialized to the  <a lt="Unicode serialization of an origin">Unicode serialization</a> of <var>request</var>'s <a for=request>origin</a>.</li>
+          <li><a>Create an event</a> <var>e</var> with the {{ForeignFetchEvent}} interface.</li>
+          <li>Initialize <var>e</var>’s {{Event/type}} attribute to "<code>foreignfetch</code>".</li>
+          <li>Initialize <var>e</var>’s {{Event/cancelable}} attribute to true.</li>
```
annevk: Same here.

docs/index.bs
```diff
@@ -3582,8 +3602,9 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
     <ol>
       <li><a>Assert</a>: a <a>Record</a> with the \[[value]] equals to <var>registration</var> is contained in <a href="#dfn-scope-to-registration-map">scope to registration map</a>.</li>
       <li><a>Assert</a>: <var>registration</var>'s <a href="#dfn-active-worker">active worker</a> is not null.</li>
+      <li>Let <var>event</var> be the <a for="/">functional event</a> for which this algorithm was invoked.</li>
```
annevk: Shouldn't this be passed in as variable?

docs/index.bs
jungkees: Without the `for=""`, it generates no dfn found error. So I fixed this specific dfn in the follow-up patch by removing `href="#..."` entirely.

> migrate away from relying on href="#..."

Yes. I'll do this cleanup soon as a separate work.

docs/index.bs
jungkees: _installingWorker_ is not a platform object but a service worker. So it seems we can't get a "relevant global object" from it?

docs/index.bs
```diff
@@ -168,7 +168,7 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
...
     <p>A <a href="#dfn-service-worker">service worker</a> has an associated <dfn id="dfn-service-worker-id">id</dfn> (an opaque string), which uniquely identifies itself during the lifetime of its <a href="#dfn-containing-service-worker-registration">containing service worker registration</a>.</p>
-    <p>A <a href="#dfn-service-worker">service worker</a> is dispatched a set of <dfn id="dfn-lifecycle-events">lifecycle events</dfn>, <a href="#service-worker-global-scope-install-event">install</a> and <a href="#service-worker-global-scope-activate-event">activate</a>, and <dfn id="dfn-functional-events" for="">functional events</dfn> including <a href="#service-worker-global-scope-fetch-event">fetch</a>.</p>
+    <p>A <a href="#dfn-service-worker">service worker</a> is dispatched a set of <dfn id="dfn-lifecycle-events">lifecycle events</dfn>, <a href="#service-worker-global-scope-install-event">install</a> and <a href="#service-worker-global-scope-activate-event">activate</a>, and <dfn export>functional events</dfn> including <a href="#service-worker-global-scope-fetch-event">fetch</a>.</p>
```
annevk: You should keep the ID, no? Otherwise you change the ID to "functional-events".

docs/index.bs
jungkees: Right. That's what we discussed during the TPAC. My plan is to do https://github.com/w3c/ServiceWorker/issues/931#issuecomment-248570852 as a separate work. (This particular PR is for https://github.com/w3c/ServiceWorker/issues/931#issuecomment-248233338).

docs/index.bs
annevk: I guess not. Does a service worker have an environment settings object as slot? This should probably link to it.

docs/index.bs
jungkees: Explained above.

docs/index.bs
jungkees: While I agree with you, _event_ here is apparent here, and it needs changes to caller if we'd want to pass the event as an argument. Would it be better?

docs/index.bs
jungkees: Yes yes. Indeed, I was worrying about the refs being changed by the bikeshed cleanups. Now I understand how you'd not break the x-refs from the existing contents. Just missed this simple thing..

jungkees: Will get it back in the follow-up patch on your additional comments.

docs/index.bs
annevk: It just seems that per the original text they were not cancelable. Shouldn't that change be separate or pointed out somehow?

docs/index.bs
annevk: Okay. I guess moving it and then changing the approach is fine.

docs/index.bs
annevk: DOM always defaulted to not cancelable.

docs/index.bs
annevk: It would be clearer I think. In code you would also have to pass it.

docs/index.bs
jungkees: Ah.. It doesn't. It seems better to define a slot for the global object rather than the environment settings object. I don't have any other use of service worker's environment settings object apart from getting the global object. Also what'd be a good name for it? The service worker's ___ global object. Incumbent? But I think just "global object" would also be fine as a service worker won't have multiple global objects at the same time. Any suggestion?

docs/index.bs
jungkees: OK. The previous text was wrong. It didn't specify it's cancelable.

docs/index.bs
jungkees: Agreed.

docs/index.bs
annevk: Yeah, just do `<dfn for="service worker">global object</dfn>` and set it somewhere when creating the service worker.

jungkees: Great. Was working on exactly that way.

docs/index.bs
jungkees: Added an argument, event. Please update the notifications API accordingly when it's landed.

docs/index.bs
jungkees: Done.

jungkees:
@annevk, do you think it's good to merge now?

annevk:
I think so.

PR #1012, opened 2016-11-22
Fix arguments to "fetch a classic worker script"

domenic:
This follows the analogous fix in https://github.com/whatwg/html/pull/2077 for HTML.

Ideally this should be merged after the HTML pull request, but in practice it should be fine either way, as the HTML PR is pretty straightforward and will likely be merged forthwith.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3025,11 +3025,12 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
       <li>Let <var>httpsState</var> be "<code>none</code>".</li>
       <li>Let <var>referrerPolicy</var> be the empty string.</li>
       <li>Switching on <var>job</var>'s <a>worker type</a>, run these substeps with the following options:
+        <!-- TODO: reorganize algorithm so that the worker environment is created before fetching happens -->
         <dl>
         <dt><em>"<code>classic</code>"</em></dt>
-        <dd><p><a>Fetch a classic worker script</a> given <var>job</var>’s <a lt="URL serializer">serialized</a> <a href="#dfn-job-script-url">script url</a>, <var>job</var>’s <a href="#dfn-job-client">client</a>, and "<code>serviceworker</code>".</p></dd>
+        <dd><p><a>Fetch a classic worker script</a> given <var>job</var>’s <a lt="URL serializer">serialized</a> <a href="#dfn-job-script-url">script url</a>, <var>job</var>’s <a href="#dfn-job-client">client</a>, "<code>serviceworker</code>", and the to-be-created <a>environment settings object</a> for this service worker.</p></dd>
         <dt><em>"<code>module</code>"</em></dt>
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -3025,11 +3025,12 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
       <li>Let <var>httpsState</var> be "<code>none</code>".</li>
       <li>Let <var>referrerPolicy</var> be the empty string.</li>
       <li>Switching on <var>job</var>'s <a>worker type</a>, run these substeps with the following options:
+        <!-- TODO: reorganize algorithm so that the worker environment is created before fetching happens -->
         <dl>
         <dt><em>"<code>classic</code>"</em></dt>
-        <dd><p><a>Fetch a classic worker script</a> given <var>job</var>’s <a lt="URL serializer">serialized</a> <a href="#dfn-job-script-url">script url</a>, <var>job</var>’s <a href="#dfn-job-client">client</a>, and "<code>serviceworker</code>".</p></dd>
+        <dd><p><a>Fetch a classic worker script</a> given <var>job</var>’s <a lt="URL serializer">serialized</a> <a href="#dfn-job-script-url">script url</a>, <var>job</var>’s <a href="#dfn-job-client">client</a>, "<code>serviceworker</code>", and the to-be-created <a>environment settings object</a> for this service worker.</p></dd>
```
jungkees: An environment settings object for each service worker execution changes. [Run Service Worker](https://w3c.github.io/ServiceWorker/#run-service-worker-algorithm) step 7 creates one before run the script in step 14. So, I guess we should set the script's settings object slot to the created environment settings object in Run Service Worker. And just pass null to Fetch a classic worker script algorithm's script settings object param here, if that makes sense. For that, additional guards would be needed in _create a classic script_ I guess.

domenic: I'd prefer to fix this in a separate issue. It seems really hard to me to fix. You can't just mutate a script's settings object slot like that; the settings object determines how the script is parsed and some other important things, which happens during fetching. You can't pass null in that case as otherwise it's impossible to parse the script.

jungkees: I see. Having looked at those _fetch/run worker script_ algorithms, in the case of module script, the parsing is all done in _creating a module script_ which is called during the fetching phase. The problem we have with SW script execution model is that fetching and running execute in separate places with a different lifecycle from that of other workers. I don't think I fully understand the module script execution model, but would it make sense to do something like:
- SW: Create a (concrete but initialized with minimal things) environment settings object before [Update](https://w3c.github.io/ServiceWorker/#update-algorithm) step 7 and give it to _fetch a module worker script graph_.
- HTML: This above steps compose the module map as expected and it's stored in the _script_ asynchronously returned as a result of the fetch.
- SW: That _script_ is stored in the [service worker](https://w3c.github.io/ServiceWorker/#dfn-service-worker)'s [script resource](https://w3c.github.io/ServiceWorker/#dfn-script-resource) slot.
- SW: When Run Service Worker is invoked, it's given the service worker and the _script_ in the script resource slot. In this algorithm, when a new environment settings object is created, copy necessary states from the given _script_'s settings object. Or even just reuse this _script_'s settings object as-is to run the script? (Not sure this scenario makes sense. Can a realm created during the fetching phase can be reused in all the subsequent Run Service Worker script executions?)

Anyway, agreed that we'd better fix it in a separate issue. I'll merge this to be aligned with HTML changes in the meantime.

jungkees: Remaining issue filed: https://github.com/w3c/ServiceWorker/issues/1013.

PR #1018, opened 2016-12-02
Add client.type

### LINKED ISSUE #1005: consider exposing a type attribute on Client
Currently `Clients.matchAll()` considers three different types of Client objects:

1. Window
2. DedicatedWorker
3. SharedWorker

Currently its only possible to differentiate the Window client using `instanceof WindowClient`.  Since the algorithm operates on the type, though, it would be nice to expose it for dedicated and shared workers as well.

So, something like:

```
interface Client
{
  readonly attribute DOMString type;
}
```

The value would match those used in `Clients.matchAll()`.

As a side note, I find it annoying that "worker" is mapped to DedicatedWorker.  I think it would have been nicer to use "dedicatedworker" and make "worker" map to all Worker types.

jungkees:
This adds a type attribute to Client interface. Before this, we could
only differentiate the WindowClient objects from Client objects. With
the type attribute, a client can be identified as one of window,
dedicated worker, and shared worker.

Fixes #1005.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -866,6 +866,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       interface Client {
         readonly attribute USVString url;
         readonly attribute DOMString id;
+        readonly attribute ClientType type;
         readonly attribute boolean reserved;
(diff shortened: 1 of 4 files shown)


jakearchibald:
LGTM!

Issue #894, opened 2016-05-09
potential http cache issues with importScripts()


wanderview:
Currently the spec says to use RequestCache `reload` when updating the service worker script if its been over 24 hours since the last update check.  I don't however, see anything about bypassing the http cache for files fetched for `importScripts()`.  I think this could lead to http cache coherence problems when loading and evaluating a service worker script thats split across multiple files.

Edit: My statement about the default http cache behavior in gecko was not accurate.

jakearchibald:
Hmm, I agree. We should use the same cache mode for `importScripts` as we do for the service worker script itself.

wanderview:
Maybe this is something chrome is already doing, but I don't think we do it in firefox right now.  Kind of a sharp edge waiting to catch someone.

jakearchibald:
Pre F2F notes: The `reload` behaviour in SW avoids lock-in. We don't need it for `importScripts` requests, since you can cache-bust urls in the SW script, but is this difference a gotcha?

jakearchibald:
F2F:

* If the main script is skipping cache, all imported modules should do the same

wanderview:
Gecko bug:  https://bugzilla.mozilla.org/show_bug.cgi?id=1290939

Sounds like chrome already does this.

jungkees:
> If the main script is skipping cache, all imported modules should do the same

I'd like to double-check this behavior. The main script "skipping cache" is different from the main script "changed". I believe the model has been a service worker version holds until the main script itself is changed. In this proposed change, do we want to include all the to-be-imported-scripts for calculating the byte-by-byte match decision for short-circuiting a registration? Then, if an imported script has been changed, will we see this set of scripts as a different service worker version? Advancing to Install in this case seems weird to me.

jungkees:
Note that in the current spec behavior, the imported scripts are served from the cache without hitting the network once the installation is successfully complete for a service worker version.

jungkees:
Sorry. In my previous two comments, I completely missed this history: https://github.com/w3c/ServiceWorker/issues/839#issuecomment-236256162. I'll dig into it again with that context and consensus considered.

jungkees:
Closed by https://github.com/w3c/ServiceWorker/commit/7deb2380a2b8fc327d73636f7c185a22f85f11d5.

Issue #753, opened 2015-09-29
how does SoftUpdate "force bypass cache flag" get set?


wanderview:
Currently the SoftUpdate algorithm takes a "force bypass cache flag".  How does this ever get set?  Is it needed any more?

jungkees:
Currently no callers in the spec sets the _force bypass cache flag_ as pointed. I just left it there for some potential use from implementers (e.g. calls from devtools) as of now.

There's another aspect. Algorithms in other specs are supposed to dispatch their functional events by invoking [Handle Functional Event](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#handle-functional-event-algorithm) which calls Soft Update at its last step without setting the flag. I'm not sure whether it makes sense but some algorithms may want to call Soft Update directly and with the flag set on their own need.

Does the above usage make sense? Or should be drop the flag?

wanderview:
That makes sense to me, but a note might help other people reading the spec.  Thanks.

PR #1020, opened 2016-12-06
Fetch service worker scripts with "no-cache" by default

### LINKED ISSUE #893: consider fetching service worker scripts with no-cache by default
Lately I've found myself recommending to people that they use `Cache-control: no-cache` or `max-age: 0` on their service worker script resources.  It seems best practice in order for service worker updates to be detected as devs expect.

Should Update algorithm just fetch the service worker script with `Request.cache='no-cache'` by default?  One could argue that an "update check" is semantically similar to pressing the refresh button.

Edit: Removed the erroneous statement about the firefox http cache.

jungkees:
This introduces service worker registration's use cache field and its
related APIs: options.useCache to register method and
registration.useCache for ServiceWorkerRegistration objects. This
changes the default cache mode of fetching SW scripts to "no-cache".
After the change, 24 hours limit and job's force bypass cache flag rules
are still enforced. register() method's options value set to
{ useCache: true } re-enables the previous default behavior provided
before this change.

Fixes #893 #894.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -187,8 +187,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker registration=] has an associated <dfn export id="dfn-scope-url">scope url</dfn> (a [=/URL=]).
 
-    A [=/service worker registration=] has an associated <dfn export id="dfn-registration-script-url">registering script url</dfn> (a [=/URL=]).
-
     A [=/service worker registration=] has an associated <dfn export id="dfn-installing-worker">installing worker</dfn> (a [=/service worker=] or null) whose [=service worker/state=] is *installing*. It is initially set to null.
 
     A [=/service worker registration=] has an associated <dfn export id="dfn-waiting-worker">waiting worker</dfn> (a [=/service worker=] or null) whose [=service worker/state=] is *installed*. It is initially set to null.
@@ -196,6 +194,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     A [=/service worker registration=] has an associated <dfn export id="dfn-active-worker">active worker</dfn> (a [=/service worker=] or null) whose [=service worker/state=] is either *activating* or *activated*. It is initially set to null.
 
     A [=/service worker registration=] has an associated <dfn export id="dfn-last-update-check-time">last update check time</dfn>. It is initially set to null.
+
+    A [=/service worker registration=] has an associated <dfn export id="dfn-use-cache">use cache</dfn> (a boolean). It is initially set to false.
 
     A [=/service worker registration=] has an associated <dfn export id="dfn-uninstalling-flag">uninstalling flag</dfn>. It is initially unset.
 
@@ -391,6 +391,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         readonly attribute ServiceWorker? active;
 
         readonly attribute USVString scope;
+        readonly attribute boolean useCache;
 
         [NewObject] Promise&lt;void&gt; update();
         [NewObject] Promise&lt;boolean&gt; unregister();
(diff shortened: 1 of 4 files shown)


docs/index.bs
```diff
@@ -187,8 +187,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker registration=] has an associated <dfn export id="dfn-scope-url">scope url</dfn> (a [=/URL=]).
 
-    A [=/service worker registration=] has an associated <dfn export id="dfn-registration-script-url">registering script url</dfn> (a [=/URL=]).
```
jakearchibald: Ohh, good catch!

docs/index.bs
```diff
@@ -1636,7 +1644,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. If the "<code>scope</code>" <a>target attribute</a> of the <code>Link</code> header is present, set |scopeURL| to the result of <a lt="URL parser">parsing</a> the "<code>scope</code>" <a>target attribute</a> with |scriptURL|.
       1. Let |workerType| be the "<code>workertype</code>" <a>target attribute</a> of the <code>Link</code> header, or "<code>classic</code>" if no such attribute is present.
       1. If |workerType| is not a valid {{WorkerType}} value, abort these steps.
-      1. Invoke [=Start Register=] with |scopeURL|, |scriptURL|, a new <a>promise</a>, null, |contextURL| and |workerType|.
+      1. Let |useCache| be the "<code>usecache</code>" <a>target attribute</a> of the <code>Link</code> header, or false if no such attribute is present.
+      1. If |useCache| is not a valid boolean value, abort these steps.
```
jakearchibald: What would be an invalid value?

jungkees: Any other value except true or false. I copied this from other target attribute's value sanitization steps above. @mkruisselbrink, could you check if this step is needed and done in a right way?

jungkees:
> it needs to request imported scripts no-cache too.

Addressed. Note that job's force bypass cache flag is effective for the script executions in both Update and Install for now. Depending on https://github.com/w3c/ServiceWorker/issues/1021, there's a chance that the script execution in Install won't need this bypass cache flag.

jakearchibald:
I don't think updates are doing the right thing.

"Soft update" is never called with *force bypass cache flag*. "Update" 9.6 will only set *force bypass cache for importscripts flag* if *force bypass cache flag* is set, so it ignores *use cache* on the registration.

jakearchibald:
Same goes for "Install" - it doesn't pay attention to *use cache* on the registration when it comes to `importScripts`.

jungkees:
@jakearchibald, _force bypass cache flag_ is for dev tools, debugging purposes, potential calls from other specs, etc. (See https://github.com/w3c/ServiceWorker/issues/753.) The registration's use cache field is being checked in Update and `importScripts()` as one of the conditions for "no-cache". In this change, I made the registration's use cache is set in Set Registrations algorithm given job's use cache that's passed from Register requests.

docs/index.bs
```diff
@@ -2500,7 +2527,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
           Else, continue the rest of these steps after the algorithm's asynchronous completion, with |script| being the asynchronous completion value.
 
-      1. If |newestWorker| is not null, |newestWorker|'s [=service worker/script url=] [=url/equals=] |job|'s [=job/script url=] with the *exclude fragments flag* set, and |script| is a byte-for-byte match with |newestWorker|'s <a>script resource</a>, then:
+      1. If |newestWorker| is not null, |newestWorker|'s [=service worker/script url=] [=url/equals=] |job|'s [=job/script url=] with the *exclude fragments flag* set, and |script|'s [=source text=] is a byte-for-byte match with |newestWorker|'s [=script resource=]'s [=source text=], if |script| is a [=classic script=], and |script|'s [=module record=]'s \[[ECMAScriptCode]] is a byte-for-byte match with |newestWorker|'s [=script resource=]'s [=module record=]'s \[[ECMAScriptCode]] otherwise, then:
```
jungkees: Addressed: https://github.com/whatwg/html/pull/2157. And put module script specifier for module records: https://github.com/w3c/ServiceWorker/pull/1020/commits/d2047e1cc3cd37955e2fd15cd73e688b77f62e99.

jakearchibald:
Ah, I think I was looking at a commit before `Apply cache mode "no-cache" to importScripts() requests`. LGTM!

Issue #1013, opened 2016-11-24
Sort out settings object for service worker module scripts


jungkees:
SW has different script execution model from other workers. SW's script fetching and running execute in separate places with a different lifecycle.

Fetch a module worker script graph requires a settings object in the fetching phase to pare the scripts. (Fetch a classic worker script also requires this argument as well but seems not using for parsing in the fetching phase.)

We need to fix it such that module scripts parsed during the fetching phase can be used in subsequent Run Service Worker script executions.

Issue reported in https://github.com/w3c/ServiceWorker/pull/1012#pullrequestreview-9860300.

jungkees:
@Domenic, @annevk,

I've been thinking about the environment settings object for service workers. Currently we create a settings object for a service worker whenever [Run Service Worker](https://w3c.github.io/ServiceWorker/#run-service-worker-algorithm) needs a cold start. So, it's basically 1:1:1 to the realm and the global object as for other environments. However, having thought about it more, I think it should rather be bound to the service worker's lifetime instead of the realm and global object's lifetime. It seems most of the settings object's fields can be (or need to be) maintained across multiple global objects' creations and terminations. Moreover, _fetching a * worker script_ requires a settings object, which isn't what service worker's [Update](https://w3c.github.io/ServiceWorker/#update-algorithm) algorithm strictly follows at the moment. (Update step 7 now passes a to-be-created environment settings object to _fetch a * worker script_.)

That considered, I propose we create an environment settings object for a service worker in [Update](https://w3c.github.io/ServiceWorker/#update-algorithm) algorithm before invoking _fetch a * worker script_. This settings object is associated with the service worker's script during _fetch a * worker script_ execution. And [Run Service Worker](https://w3c.github.io/ServiceWorker/#run-service-worker-algorithm) algorithm later uses this service worker's script's settings object to run the script with additional set up steps that associate the settings object with the created realm and the global object.

One tricky part is [fetch a module worker script graph](https://html.spec.whatwg.org/#fetch-a-module-worker-script-tree) > .. > [create a module script](https://html.spec.whatwg.org/#creating-a-module-script) > [ParseModule()](https://tc39.github.io/ecma262/#sec-parsemodule) requires a realm while service worker's Update doesn't create a realm. In my understanding, the only reason that the realm is passed is to set the created script's module record's [[Realm]] internal slot, right? If so, can I set/pass the environment settings object's Realm to undefined at this stage? And later override the value to the actual realm created in Run Service Worker?

[...]

PR #1023, opened 2016-12-12
Include imported scripts to byte-check

### LINKED ISSUE #839: Consider relying on eTags (or other headers) for service worker dependencies to check for updates
For the sake of modularization and isolation. Could it be possible to improve the update algorithm to rely on the eTag / content-size / other headers sent by the server to decide when a service worker changed? Now we need to include a mark of change in the sw and this forces a lot of developments to postprocess service worker files.

jungkees:
This changes the behavior of the service worker script resource
comparison. Before this, only the main service worker script was
compared to a new version. With this change, all the imported scripts
stored in the imported scripts map as well as the main script are
inspected against the corresponding network resources (based on the
urls.)

Note:
 - Service worker's script resource map has been renamed and moved to
 service worker's script resource's imported scritps map.
 - registration's last update check time's always updated whenever the
 response is fetched from the network (regardless it's a main script or
 an imported script.)

Fixes #839.


docs/index.bs
```diff
@@ -3223,6 +3223,37 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+      : Input
+      :: |sourceScript|, a [=script=]
+      :: |targetScript|, a [=script=]
+      :: |registration|, a [=/service worker registration=]
+      :: |job|, a [=job=]
+      : Output
+      :: True or false, a boolean
+
+      1. If |sourceScript| and |targetScript| are not both [=classic scripts=] or not both [=module scripts=], return false.
```
jakearchibald: Oh wow good catch

docs/index.bs
```diff
@@ -3223,6 +3223,37 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+              1. Let |targetResponse| be the result of [=fetch|fetching=] |request|.
+              1. If |targetResponse|'s <a for="response" href="https://github.com/whatwg/fetch/issues/376">cache state</a> is not "<code>local</code>", set |registration|’s [=last update check time=] to the current time.
+              1. Let |targetResponse| be |targetResponse|'s [=unsafe response=].
+              1. If |targetResponse|'s [=response/type=] is "<code>error</code>", or |targetResponse|'s [=response/status=] is not an [=ok status=], return false.
+              1. If the result of [=UTF-8 decoding=] |response|'s [=response/body=] is not a byte-for-byte match with the result of [=UTF-8 decoding=] |targetResponse|'s [=response/body=], return false.
+      1. If |sourceScript| is a [=module script=], then:
+          1. If |sourceScript|’s [=module script/module record=]'s \[[ECMAScriptCode]] is not a byte-for-byte match with |targetScript|’s [=module script/module record=]'s \[[ECMAScriptCode]], return false.
```
jakearchibald: Shouldn't we check imported scripts too?

docs/index.bs
```diff
@@ -3223,6 +3223,37 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      : Output
+      :: True or false, a boolean
+
+      1. If |sourceScript| and |targetScript| are not both [=classic scripts=] or not both [=module scripts=], return false.
+      1. If |sourceScript| is a [=classic script=], then:
+          1. If |sourceScript|'s [=source text=] is not a byte-for-byte match with |targetScript|'s [=source text=], return false.
+          1. Let |sourceMap| be |sourceScript|'s [=script resource/imported scripts map=].
+          1. [=map/For each=] |url| → |response| of |sourceMap|:
+              1. Let |request| be a new [=/request=] whose [=request/url=] is |url|, [=request/client=] is |job|'s [=job/client=], [=request/type=] is "<code>script</code>", [=request/destination=] is "<code>script</code>", [=request/parser metadata=] is "<code>not parser-inserted</code>", [=request/synchronous flag=] is set, and whose [=request/use-URL-credentials flag=] is set.
+              1. Set |request|'s [=request/cache mode=] to "<code>no-cache</code>" if any of the following are true:
```
jakearchibald: This is going to cause a double-download. I'll explain this in another comment…

jakearchibald:
When the imported scripts are checked, we're doing it with `no-cache`, but then later when we run the script it'll execute `importScripts` and once again request the resources `no-cache`.

I hadn't thought about this until reading this PR, but I think it's something we need to solve.

"Check If Service Worker Resources Are Identical" could return a boolean as it does now, but also a map of all the scripts it fetched. Then, `importScripts` could check this cache before going to the network. This cache can be deleted once install completes - or whenever we set the "imported scripts updated flag" flag, which discards imports that were used in the previous version, but not this one.

docs/index.bs
jungkees: `importScripts()` is not supported in module scripts: https://html.spec.whatwg.org/#import-scripts-into-worker-global-scope.

jungkees:
> When the imported scripts are checked, we're doing it with no-cache, but then later when we run the script it'll execute importScripts and once again request the resources no-cache.

Right. I concerned about this too, but it was still better than invoking Run Service Worker for every navigation.

> "Check If Service Worker Resources Are Identical" could return a boolean as it does now, but also a map of all the scripts it fetched.

It seems a bit irony that we should introduce this kind of temporary cache to support bypassing HTTP cache. But yeah I think this is definitely a way to avoid the double-fetch of the seemingly identical resource (unless the content is dynamically changing in the server.) I'd like to also hear opinions from @wanderview and @mattto.

docs/index.bs
jungkees: I thought those urls are fetched, parsed and stored as part of the module record, but maybe I'm wrong. I noticed an environment settings object has a module map which is referenced throughout traversing the module script graph. I'll check the algorithms again.

jakearchibald: Oh does it flatten it somehow? I couldn't figure it out 😞

jungkees: I believe so with https://github.com/whatwg/html/pull/2157. But I'll double-check.

jungkees: @jakearchibald, I think you're right. I can't find any step that flattens the entire imported assets. It seems each module script has module record that has the parsed source text and info to the required modules (record's [[RequestedModules]] in ECMAScript language.) Getting to [ModuleEvaluation()](https://tc39.github.io/ecma262/#sec-moduleevaluation) when running a module script, it evaluates all the related module scripts recursively finding the dependencies.

And I think we should compare all the module scripts stored in an environment settings object's module map. I'd want to confirm this with @domenic before updating this part.

@domenic, SW intends to compare all the script assets including the main script and the imported scripts for both classic and module scripts. For classic scripts, SW has it's own imported scripts map that's used for this byte-by-byte check. But for module scripts, I think I misunderstood that the main script's module record has a flattened asset that include all the imported scripts in the graph. I'd like to check with you whether this is wrong, and if so, iterating an environment setting's object's module map and compare the source texts of all the module scripts in the map would be a right way for our purpose?

jakearchibald: @jungkees whoa, I replied before seeing your message. Weird how similar they are!

jakearchibald:
> It seems a bit irony that we should introduce this kind of temporary cache to support bypassing HTTP cache

Hah, yeah. But right now we fetch the main script, do the comparison, then reuse that script we fetched. It seems weird to reuse for the main script but refetch imported scripts. Also, this double-fetch means the scripts we byte-check could be different to the ones we end up running, which seems weird.

wanderview:
> When the imported scripts are checked, we're doing it with no-cache, but then later when we run the script it'll execute importScripts and once again request the resources no-cache.

How about:

1. The byte-for-byte check could populate a "script resource map" like `importScripts()` does now.
2. Then the that "script resource map" can be set as the SW's map.
3. The `importScripts()` algorithm could then check the map before going to network/http cache.
4. After script eval or install would need something to remove resources put in the map in step (1), but no longer used by the new service worker script.  This is annoying, but not too with some book keeping.

Note, step (3) would also fix any issues where we go to network/http cache multiple times if script eval calls `importScripts()` multiple times.

jungkees:
I think @jakearchibald and @wanderview largely suggested the same thing. Basically byte-check will populate a temporary map if the fetched resource is new. `importScripts()` will consult this map (to populate SW's script resource's imported scripts map*) before fetching through HTTP cache/network until the imported scripts updated flag is set. Once the flag is set, `importScripts()` always returns the results from the imported scripts map.

I think if we define and maintain the temporary map on a service worker registration, we should nothing but add and update the entry throughout the registration's lifetime.

I'll add this change to the PR.

(*) This is the SW's script resource map in the current spec. I suggested changing the name and the location of the concept as such through this PR for clarity and a better interface to other algorithms.

docs/index.bs
domenic: Right, there is no flattened representation of the graph. I think checking everything in the module map is more correct.

I guess [[ECMAScriptCode]] is indeed not correct :-/. However, what we can do is update HTML to move "source text" from classic scripts only to the base script type. We'd probably want to add a note saying something like "source text is only used for classic scripts in this specification, but the Service Workers specification needs to store it for reasons". I don't think you all should create a separate parallel registry :).

jungkees: That sounds good. Moving the source text to the base script type will help us use the source text of scripts for both classic and module script cases. I'll spec it to checking everything in the module map with that change.

Another thing that needs a solution is the lifetime of the scripts and the map for SW. With other web workers, the lifetime of the scripts is tied to the `WorkerGlobalScope` object, so it doesn't store the main script and only stores the module scripts in the module map on the global. But with SW, the lifetime of the scripts outlives those globals, so it stores the main script and the imported scripts on a service worker concept rather than on the SW global. Because of this difference, I think we need to discuss around how to make the fetch a classic worker script and the fetch a module worker script graph work more correctly with SW. There's an opened issue for this: https://github.com/w3c/ServiceWorker/issues/1013.

jungkees:
https://github.com/w3c/ServiceWorker/pull/1283 obsoleted this PR.

Issue #1039, opened 2016-12-21
async waitUntil() does not allow extension on final promise resolution?


wanderview:
I feel like we discussed this case at one point, but can't find where:

```
addEventListener('fetch', evt => {
  // keep alive the SW based on some promise
  let resolveOuterPromise;
  let p = new Promise(resolve => resolveOuterPromise = resolve);
  evt.waitUntil(p);

  // Code can be defined to run on the final keep alive promise resolution that
  // then keeps the SW alive again.
  p.then(_ => {
    // I think the spec says this should throw, but I thought we agreed this should work.
    evt.waitUntil(fetch(evt.request));
  });

  // promise resolves at some point...
  setTimeout(_ => resolveOuterPromise(), 100);
});
```

I don't think the spec allows this as currently written.  Step 4 here:

https://w3c.github.io/ServiceWorker/#extend-service-worker-lifetime-algorithm

Unsets the flag allowing extension immediately after the current set of extension promises settles in step 2.  I think this blocks the addition of new extension promises?

Or is there something which guarantees other .then() handlers will run between step 2 and 3 here?

I think we should either explicitly have a microtask to run step 3/4 or run add a comment about some hidden invariant which results in this outcome.

wanderview:
FWIW, we have implemented this behavior in FF and written a WPT test verifying it.  See:

https://dxr.mozilla.org/mozilla-central/source/testing/web-platform/tests/service-workers/service-worker/resources/extendable-event-async-waituntil.js#27

And:

https://dxr.mozilla.org/mozilla-central/source/testing/web-platform/tests/service-workers/service-worker/resources/extendable-event-async-waituntil.js#33

jungkees:
> I don't think the spec allows this as currently written.

Yes, that's right. I think we should add all the derived promises called by `Promise.prototype.then` of the promises in the extend lifetime promises list. I'll try to spec it in an implementable way.

stefhak:
Related(?):
I've been trying to use ```.waitUntil()``` with the ```push``` event to make sure things get done before the SW stops, basically:
```
self.addEventListener('push', evt => {
    evt.waitUntil(handlePush(evt.data));
    //or alternatively just: handlePush(evt.data);
}
```
where ```handlePush``` returns a promise that is resolved when done. I could not determine any prolonged run time with ```evt.waitUntil```. 
[...]

Issue #935, opened 2016-07-27
should async waitUntil() work in .then() handler of promise passed to previous waitUntil()?


wanderview:
I'm currently implementing async `waitUntil()` in gecko.  The main intention here is to allow things like:

```
addEventListener('fetch', e => {
  e.respondWith(fetch(e.request)).then(response => {
    e.waitUntil(caches.open('db').then(cache => {
      return cache.put(e.request, response);
    });
    return response.clone();
  });
});
```

So you can call `waitUntil()` as long as there is an extension promise still unsettled.

But what about this non-idiomatic code:

```
addEventListener('fetch', e => {
  let respondWithPromise = fetch(e.request);
  respondWithPromise.then(response => {
    let clone = response.clone();
    e.waitUntil(caches.open('db').then(cache => {
      return cache.put(e.request, clone);
    });
  });
  e.respondWith(respondWithPromise);
});
```

Here pass the `fetch()` promise directly to `respondWith()`.  Or additional caching code is also hung off this same `fetch()` promise.

Technically when the caching code runs the promise passed to `respondWith()` is settled, but if order of promise handlers is maintained then the `respondWith()` code has not been notified yet.

I _think_ this should probably work, but the spec text is a bit vague.  It says:

> Wait until the result of waiting for all of extendLifetimePromises settles. 

Does that respect ordering of promise handlers?

Conversely, if we call `respondWith()` before the `respondWithPromise.then()` above, then I think the `waitUntil` should throw.  The respondWith code would clear the extensions allowed flag before the caching code had a chance to run.

All this is corner casey, but I thought it would be good to clarify.

wanderview:
I guess the "waiting for all" here:

> Wait until the result of waiting for all of extendLifetimePromises settles. 

Explicitly links to the Promise.all() algorithm.  So I guess this is defined and this should work:

```
addEventListener('fetch', e => {
  let respondWithPromise = fetch(e.request);
  respondWithPromise.then(response => {
    let clone = response.clone();
    e.waitUntil(caches.open('db').then(cache => {
      return cache.put(e.request, clone);
    });
  });
  e.respondWith(respondWithPromise);
});
```

But this would trigger a throwing waitUntil():

```
addEventListener('fetch', e => {
  let respondWithPromise = fetch(e.request);
  e.respondWith(respondWithPromise);
  respondWithPromise.then(response => {
    let clone = response.clone();
    e.waitUntil(caches.open('db').then(cache => {
[...]

PR #1049, opened 2017-01-05
Improve handling of extend lifetime promises

jungkees:
(1) This changes the approach to unsetting the extendable events'
extensions allowed flag. This change introduced a reference count based
approach instead of the promise-copying-and-checking approach.

(2) This also extends the opportunities of the lifetime extension by
allowing calling waitUntil() within microtasks queued by the given
promise's Promise.prototype.then callbacks.

Related issues:
 - https://github.com/w3c/ServiceWorker/issues/931 (1)
 - https://github.com/w3c/ServiceWorker/issues/935 (2)
 - https://github.com/w3c/ServiceWorker/issues/1039 (2)

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2855,8 +2855,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       : Output
       :: None
 
-      1. Wait, [=in parallel=], until |promise| is settled and |promise|'s <code>then</code> methods, if any, in the task where |promise| has been settled have executed.
-      2. [=Queue a microtask=], on |promise|'s [=relevant settings object=]'s [=responsible event loop=], to decrease |event|'s [=ExtendableEvent/pending promises count=] by one.
+      1. Wait until |promise| is settled [=in parallel=].
+      2. [=Queue a task=], on |promise|'s [=relevant settings object=]'s [=responsible event loop=], to decrease |event|'s [=ExtendableEvent/pending promises count=] by one.
   </section>
 
   <section algorithm>
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1289,15 +1289,11 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     An {{ExtendableEvent}} object has an associated <dfn for="ExtendableEvent">extend lifetime promises</dfn> (an array of <a>promises</a>). It is initially an empty array.
 
-    An {{ExtendableEvent}} object has an associated <dfn for="ExtendableEvent" id="extensions-allowed-flag">extensions allowed flag</dfn>. It is initially set.
-
     An {{ExtendableEvent}} object has an associated <dfn for="ExtendableEvent">pending promises count</dfn> (the number of pending promises in the [=ExtendableEvent/extend lifetime promises=]). It is initially set to zero.
 
     [=/Service workers=] have two <a>lifecycle events</a>, {{install!!event}} and {{activate!!event}}. [=/Service workers=] use the {{ExtendableEvent}} interface for {{activate!!event}} event and {{install!!event}} event.
 
     <a href="#extensibility">Service worker extensions</a> that <a href="#extension-to-service-worker-global-scope">define event handlers</a> *may* also use or extend the {{ExtendableEvent}} interface.
-
-    Note: To extend the lifetime of a [=/service worker=], algorithms that <a>dispatch</a> events using the {{ExtendableEvent}} interface run [=Extend Service Worker Lifetime=] algorithm after <a>dispatching</a> the event. See <a>Handle Fetch</a>, <a>Handle Foreign Fetch</a>, and <a>Handle Functional Event</a>.
 
     <section algorithm="wait-until-method">
       <h4 id="wait-until-method">{{ExtendableEvent/waitUntil()|event.waitUntil(f)}}</h4>
@@ -1306,9 +1302,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       <dfn method for="ExtendableEvent"><code>waitUntil(|f|)</code></dfn> method *must* run these steps:
 
-        1. If the <a>extensions allowed flag</a> is unset, then:
+        1. If the [=ExtendableEvent/pending promises count=] is zero and the [=dispatch flag=] is unset, then:
             1. <a>Throw</a> an "{{InvalidStateError}}" exception.
             1. Abort these steps.
+
+            Note: If no lifetime extension promise has been added in the task that called the event handlers), calling {{ExtendableEvent/waitUntil()}} in subsequent asynchronous tasks will throw.
+
         1. Add |f| to the [=ExtendableEvent/extend lifetime promises=].
         1. Increase the [=ExtendableEvent/pending promises count=] by one.
 
@@ -1316,6 +1315,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
         1. Run [=Handle Extend Lifetime Promise=] with the [=context object=] and |f|.
         1. Return.
+
+        The user agent *should not* <a lt="terminate service worker">terminate</a> the [=/service worker=] associated with |event|'s <a>relevant settings object</a>'s [=environment settings object/global object=] when |event|'s [=dispatch flag=] is set or |event|'s [=ExtendableEvent/pending promises count=] is not zero. However, the user agent *may* impose a time limit to this lifetime extension.
     </section>
 
     [=/Service workers=] and <a href="#extensibility">extensions</a> that <a href="#extension-to-service-worker-global-scope">define event handlers</a> *may* define their own behaviors, allowing the [=ExtendableEvent/extend lifetime promises=] to suggest operation length, and the rejected state of any of the <a>promise</a> in [=ExtendableEvent/extend lifetime promises=] to suggest operation failure.
@@ -2696,9 +2697,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Let |e| be the result of <a>creating an event</a> with {{InstallEvent}}.
           1. Initialize |e|’s {{Event/type}} attribute to {{install!!event}}.
           1. <a>Dispatch</a> |e| at |installingWorker|'s [=service worker/global object=].
-          1. Invoke [=Extend Service Worker Lifetime=] with |e|.
           1. *WaitForAsynchronousExtensions*: Run the following substeps <a>in parallel</a>:
-              1. Wait until |e|'s <a>extensions allowed flag</a> is unset.
+              1. Wait until |e|'s [=ExtendableEvent/pending promises count=] is zero.
               1. If the result of <a>waiting for all</a> of |e|'s <a>extend lifetime promises</a> rejected, set |installFailed| to true.
 
           If |task| is discarded or the script has been aborted by the <a lt="Terminate Service Worker">termination</a> of |installingWorker|, set |installFailed| to true.
@@ -2762,8 +2762,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Let |e| be the result of <a>creating an event</a> with {{ExtendableEvent}}.
           1. Initialize |e|’s {{Event/type}} attribute to {{activate!!event}}.
           1. <a>Dispatch</a> |e| at |activeWorker|'s [=service worker/global object=].
-          1. Invoke [=Extend Service Worker Lifetime=] with |e|.
-          1. *WaitForAsynchronousExtensions*: Wait, <a>in parallel</a>, until |e|'s <a>extensions allowed flag</a> is unset.
+          1. *WaitForAsynchronousExtensions*: Wait, <a>in parallel</a>, until |e|'s [=ExtendableEvent/pending promises count=] is zero.
       1. Wait for |task| to have executed or been discarded, or the script to have been aborted by the <a lt="terminate service worker">termination</a> of |activeWorker|.
       1. Wait for the step labeled *WaitForAsynchronousExtensions* to complete.
       1. Run the <a>Update Worker State</a> algorithm passing |registration|'s <a>active worker</a> and *activated* as the arguments.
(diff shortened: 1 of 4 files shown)


jungkees:
@jakearchibald, @annevk, this is about (1) introducing reference count for checking extend lifetime promises and (2) allowing `e.waitUntil()` in the given promise's promise.then callbacks. Your review would be appreciated.

@domenic, it'd be great if you had time to review these particular steps where I referenced ECMAScript internal slots:
 - https://github.com/w3c/ServiceWorker/pull/1049/commits/3210c4d08619341accee4d4a84eb161273297afd#diff-27b79860afe28f01aed4f1f6228367faR2878
 - https://github.com/w3c/ServiceWorker/pull/1049/commits/3210c4d08619341accee4d4a84eb161273297afd#diff-27b79860afe28f01aed4f1f6228367faR2880

/cc @wanderview

docs/index.bs
```diff
@@ -1289,7 +1289,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     An {{ExtendableEvent}} object has an associated <dfn for="ExtendableEvent">extend lifetime promises</dfn> (an array of <a>promises</a>). It is initially an empty array.
 
-    An {{ExtendableEvent}} object has an associated <dfn id="extensions-allowed-flag">extensions allowed flag</dfn>. It is initially set.
+    An {{ExtendableEvent}} object has an associated <dfn for="ExtendableEvent" id="extensions-allowed-flag">extensions allowed flag</dfn>. It is initially set.
```
annevk: It would be good to invert this somehow. Flags and booleans that default to false are much more intuitive.

docs/index.bs
```diff
@@ -2837,18 +2855,33 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-      1. If |event|'s <a>extend lifetime promises</a> is empty, unset |event|'s <a>extensions allowed flag</a> and abort these steps.
-      1. Let |extendLifetimePromises| be an empty array.
-      1. Run the following substeps <a>in parallel</a>:
-          1. *SetupPromiseArray*: Set |extendLifetimePromises| to a copy of |event|'s <a>extend lifetime promises</a>.
-          1. Wait until all the <a>promises</a> in |extendLifetimePromises| settle.
-          1. If the length of |extendLifetimePromises| does not equal the length of |event|'s <a>extend lifetime promises</a>, jump to the step labeled *SetupPromiseArray*.
-          1. Unset |event|'s <a>extensions allowed flag</a>.
+      1. If |event|'s [=ExtendableEvent/pending promises count=] is zero, unset |event|'s [=ExtendableEvent/extensions allowed flag=] and abort these steps.
 
+          Note: If no lifetime extension promise has been added up to this point (i.e., at the end of the task that called the event handlers), the [=ExtendableEvent/extensions allowed flag=] is immediately unset. Calling {{ExtendableEvent/waitUntil()}} in subsequent asynchronous tasks will throw.
```
annevk: Why do you need both the flag and the count then?

docs/index.bs
```diff
@@ -2837,18 +2855,33 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+    <h3 id="handle-extend-lifetime-promise-algorithm"><dfn>Handle Extend Lifetime Promise</dfn></h3>
+
+      : Input
+      :: |event|, an {{ExtendableEvent}} object
+      :: |promise|, a [=promise=]
+      : Output
+      :: None
+
+      1. Wait until |promise| is settled [=in parallel=].
+      2. [=Queue a microtask=] to run the following substeps:
```
annevk: You cannot queue microtasks from something that runs in parallel. Needs to be a regular task.

docs/index.bs
```diff
@@ -2837,18 +2855,33 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+          1. Decrease |event|'s [=ExtendableEvent/pending promises count=] by one.
+          1. For each |reaction| in |promise|.\[[PromiseFulfillReactions]]:
```
annevk: This seems extremely hacky and probably wrong. I don't think we should be poking at the guts of promises.

domenic: I haven't reviewed in detail yet, but I agree poking at [[PromiseFulfillReactions]] is a bad idea. Poking at [[PromiseState]] and [[PromiseResult]] is OK though IMO.

docs/index.bs
domenic: Wait, why is that? Microtasks are just a type of task, in general, so I thought this would work.

annevk: Microtasks belong to a task. But there's no task here they could belong too.

domenic:
This doesn't appear to be a good idea. I'm not sure what is trying to be accomplished, but as specced this means that given

```js
const p = getSomePromise();
const q = p.then(doSomething);
e.waitUntil(p);
const r = p.then(doSomethingElse);
```

will only wait on `p` and `q` but not `r`, which seems surprising. I would suggest trying to find some approach that does not look at [[PromiseFulfillReactions]] at all. Maybe just looking at `p` is sufficient, in this example, since `p` is what you actually are passing to `waitUntil`.

docs/index.bs
jungkees: Agreed. Will try to change it to something like _extensions opportunity closed_ flag instead.

slightlyoff:
@domenic: that's exactly the designed behavior. You can extend the lifetime again by calling `e.waitUntil(r)` if needed.

domenic:
My point is that it's a bizarre design that the lifetime waits on q when you didn't pass q to waitUntil.

docs/index.bs
jungkees: I think the flag is still needed. Otherwise, we can't determine whether the pending count is in the initial state zero or zero as the extension possibility is over.

jungkees:
@domenic, let me try to clarify the logic first (and then let's continue about the design):

> > will only wait on p and q but not r, which seems surprising.

I think it also will wait on r too. My intention was that when Handle Extend Lifetime Promise called in waitUntil goes async, `const r = p.then(doSomethingElse);` is run and queue the microtask for the doSomethingElse before the Handle Extend Lifetime Promise queues a microtask to chase p.

Referencing Promise.[[PromiseFulfillReactions]] from the internal spec algorithm seems hacky but I think the implementation also has to attach some kind of native handlers to catch and handle the then callbacks. Still looking for a better way to spec it but haven't found one yet.

domenic:
No implementations provide such hooks today, and they're not meant to: doing so is invasive and constrains the promise implementation from performing optimizations that we're actively working on. Internal state is meant to stay internal. If you can't accomplish this with the public .then API, I'd strongly suggest rethinking the design.

Besides, it's just strange that it's waiting on promises you didn't pass to waitUntil.

jungkees:
I'll take this design discussion back to https://github.com/w3c/ServiceWorker/issues/1039. I thought @wanderview commented there that Firefox already implemented this behavior. Would like to look into that and also discuss with Chromium implementers around this too.

docs/index.bs
annevk: Can't you just check the event dispatch flag or some such for that?

docs/index.bs
jungkees: The flag is replaced by the use of the pending promises count and the dispatch flag.

docs/index.bs
jungkees: I was between leaving the flag as a variable helping the understanding of the control flows and removing it as the cleanup of unnecessary variables. I think the latter wins, so removed it.

docs/index.bs
jungkees: This should be queued from the task where the promise is settled. I explicitly used promise’s relevant settings object's responsible event loop to queue this microtask.

docs/index.bs
```diff
@@ -2830,22 +2847,16 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-      1. Let |extendLifetimePromises| be an empty array.
-      1. Run the following substeps <a>in parallel</a>:
-          1. *SetupPromiseArray*: Set |extendLifetimePromises| to a copy of |event|'s <a>extend lifetime promises</a>.
-          1. Wait until all the <a>promises</a> in |extendLifetimePromises| settle.
-          1. If the length of |extendLifetimePromises| does not equal the length of |event|'s <a>extend lifetime promises</a>, jump to the step labeled *SetupPromiseArray*.
-          1. Unset |event|'s <a>extensions allowed flag</a>.
-
-      The user agent *should not* <a lt="terminate service worker">terminate</a> the [=/service worker=] associated with |event|'s <a>relevant settings object</a>'s [=environment settings object/global object=] until |event|'s <a>extensions allowed flag</a> is unset. However, the user agent *may* impose a time limit to this lifetime extension.
+      1. Wait, [=in parallel=], until |promise| is settled and |promise|'s <code>then</code> methods, if any, in the task where |promise| has been settled have executed.
```
annevk: How would you test the "and" requirement here? That should be removed I think.

docs/index.bs
```diff
@@ -2830,22 +2847,16 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      2. [=Queue a microtask=], on |promise|'s [=relevant settings object=]'s [=responsible event loop=], to decrease |event|'s [=ExtendableEvent/pending promises count=] by one.
```
annevk: As I said earlier, you cannot queue microtasks from parallel algorithms.

docs/index.bs
jungkees: Addressed the comments in https://github.com/w3c/ServiceWorker/issues/1039#issuecomment-270826618.

docs/index.bs
jungkees: Without the condition after "and", I'm concerned about the race between this microtask (decrease the count) and the promise.then methods that can be called in the main thread (event handler tasks.) Is it an unnecessary concern?

annevk: As I said, this cannot be a microtask. It can only be a task. And therefore it should be fine.

docs/index.bs
jungkees: Well.. I couldn't find any constraints a microtask should belong to a particular task. Doesn't it work even if the step designate the target event loop? If this doesn't make sense, should I queue a task to queue this microtask?

docs/index.bs
jungkees: Yeah, if I queue a normal task here. It'd be all fine. But I thought it'd be ideal if we could use a microtask in a way. Firefox actually queues a microtask to check this: https://dxr.mozilla.org/mozilla-central/source/dom/workers/ServiceWorkerPrivate.cpp#379.

docs/index.bs
jungkees: Now ECMAScript's "PromiseJobs" are queued through https://html.spec.whatwg.org/#enqueuejob(queuename,-job,-arguments). I think those steps are not run in the main thread but in a thread in the JS engine internals.

>  you cannot queue microtasks from parallel algorithms.

The reason it doesn't make sense is because it is run in a parallel thread? Or it's running off the main thread?

annevk: Just queue a task to decrease the thing. It doesn't have to be a microtask.

Queue a microtask from a parallel thread is racy as you say, since you don't know what task it ends up in. User agents only have the ability to queue tasks to get things into the main thread.

docs/index.bs
annevk: That code doesn't seem to communicate across threads though.

docs/index.bs
jungkees: Addressed it to using a normal task.

jungkees:
@annevk, do you have any further comments? I'd like to merge it if it's ok.

annevk:
I guess a remaining problem is that you haven't defined the task source.

domenic:
Hmm, do you even need to go in parallel/queue a task? It seems like for promises you should be able to just do "upon fulfillment"/"upon rejection".

annevk:
That's a good point. If you don't queue a task though, it does start to matter when the fulfillment observer is attached, right? I guess as long as that's well-defined it's not really an issue, but would be worth testing.

Do we have tests for this feature?

wanderview:
> Do we have tests for this feature?

Not sure if its exactly what you mean, but we wrote this wpt test:

https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/extendable-event-async-waituntil.https.html#L70

jungkees:
> Hmm, do you even need to go in parallel/queue a task? It seems like for promises you should be able to just do "upon fulfillment"/"upon rejection".

I don't think we can guarantee the expected sequence by using "upon fulfillment"/"upon rejection" here. This `promise.then` for decreasing the count should be called after all the `promise.then()` functions called during the event dispatch. Either queuing a normal task or explicitly specifying the order of the microtask seems to be needed.

domenic:
Why does it need to happen after the then callbacks? That doesn't show up in the problem description. If you want it to happen after p.then(f) executes f, pass p.then(f) to waitUntil.

jungkees:
In the following snippet for example:
```js
addEventListener('push', e => {
  let outResolve;
  let p = new Promise(resolve => { outResolve = resolve; });
  e.waitUntil(p); // Upon fulfillment/rejection of p inside
  p.then(() => {
    e.waitUntil(q);
  });
  setTimeout(() => outResolve(), 1);
});
```
I think the "Upon fulfillment/rejection of p" triggered inside of `waitUntil()` queues (precisely, it appends the reaction to p.[[PromiseFulfillReactions]] as p is pending) the dcrease job before the promise job for `e.waitUntil(q);` is queued, which makes `e.waitUntil(q);` throws.

domenic:
Hmm I guess that does appear in https://github.com/w3c/ServiceWorker/issues/1039#issue-197049082. I think that's a bad design, but it's not a huge deal. I guess if you want that you can just do "upon fulfillment/rejection, queue a task".

domenic:
@jungkees yeah, I see. I think that is a bad design and if you want that code to work it should be coded as

```js
addEventListener('push', e => {
  let outResolve;
  let p = new Promise(resolve => { outResolve = resolve; });
  e.waitUntil(p.then(() => {
    e.waitUntil(q);
  });
  setTimeout(() => outResolve(), 1);
});
```

since then you actually "wait until" the argument (instead of "wait until-but-also-wait-an-extra-turn" the argument). But it's not a huge deal.

jungkees:
Yeah, "wait until-but-also-wait-an-extra-turn" the argument is the original problem statement and the behavior people agreed on. With your comment, I sort of come to think of again whether this was a right decision but would like to settle on what we decided.

For the spec wording, which of the following would be better?
```
Upon fulfillment of promise, queue a task to decrease event’s pending promises count by one.
Upon rejection of promise, queue a task to decrease event’s pending promises count by one.
```
or
```
Upon fulfillment or rejection of promise, queue a task to decrease event’s pending promises count by one.
```
or
```
Upon settlement of promise, queue a task to decrease event’s pending promises count by one.
```

domenic:
> Yeah, "wait until-but-also-wait-an-extra-turn" the argument is the original problem statement and the behavior people agreed on. With your comment, I sort of come to think of again whether this was a right decision but would like to settle on what we decided.

Yep, that's fair; I totally understand.

> For the spec wording, which of the following would be better?

IMO `Upon fulfillment or rejection of promise, queue a task to decrease event’s pending promises count by one.` is the best combination of brevity and using well-defined terms. We should maybe define "upon fulfillment" but currently we don't.

jungkees:
OK. That sounds good. Thanks!

jungkees:
@annevk, I merged it having addressed your task source comment and @domenic's promise shorthand comment. Thanks for reviewing it!

slightlyoff:
@domenic: thanks for the clarification. I agree it's a bit bizarre to have a derived promise extend the lifetime and that only the argument to `waitUntil()` should be gauranteed to extend things.

PR #1054, opened 2017-01-12
Use right value for Resolve Job Promise

### LINKED ISSUE #829: Resolve Job Promise can resolve jobs with an incorrect value
Currently the Resolve Job Promise algorithm resolves all equivalent jobs with the same value, I don't think this is correct. For example multiple equivalent register jobs should possibly be resolved with distinct `ServiceWorkerRegistration` instances (if they were called from different javascript contexts). But maybe more importantly, two equivalent jobs can have different `client` properties; this means that for example with a register job each equivalent promise needs a `ServiceWorkerRegistration` instance with a different associated service worker client.

jungkees:
This changes Resolve Job Promise algorithm to use a right value for
equivalent job promises by specifying the value should be the one for
its relevant Realm.

Fixes #829.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2544,14 +2544,14 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       : Input
       :: |job|, a [=job=]
-      :: |reason|, an [=exception=]
+      :: |errorName|, a string
       : Output
       :: none
 
-      1. If |job|'s [=job/client=] is not null, [=queue a task=], on |job|'s [=job/client=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to reject |job|'s [=job/job promise=] with an exception that is a copy of |reason|, in |job|'s [=job/client=]'s [=environment settings object/Realm=].
+      1. If |job|'s [=job/client=] is not null, [=queue a task=], on |job|'s [=job/client=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to reject |job|'s [=job/job promise=] with a [=exception/create|new=] [=exception=] with |errorName| and a user agent-defined [=exception/message=].
       1. For each |equivalentJob| in |job|'s [=list of equivalent jobs=]:
-          1. If |equivalentJob|'s [=job/client=] is null, continue to the next iteration of the loop:
-          1. [=Queue a task=], on |equivalentJob|'s [=job/client=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to reject |equivalentJob|'s [=job/job promise=] with an exception that is a copy of |reason|, in |equivalentJob|'s [=job/client=]'s [=environment settings object/Realm=].
+          1. If |equivalentJob|'s [=job/client=] is null, [=iteration/continue=].
+          1. [=Queue a task=], on |equivalentJob|'s [=job/client=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to reject |equivalentJob|'s [=job/job promise=] with a [=exception/create|new=] [=exception=] with |errorName| and a user agent-defined [=exception/message=].
   </section>
 
   <section algorithm>
@@ -2594,13 +2594,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       :: none
 
       1. If the result of running <a>potentially trustworthy origin</a> with the [=environment settings object/origin=] of |job|'s [=job/script url=] as the argument is <code>Not Trusted</code>, then:
-          1. Invoke <a>Reject Job Promise</a> with |job| and a "{{SecurityError}}" exception.
+          1. Invoke [=Reject Job Promise=] with |job| and "`SecurityError`".
           1. Invoke <a>Finish Job</a> with |job| and abort these steps.
       1. If the [=environment settings object/origin=] of |job|'s [=job/script url=] is not |job|'s [=job/referrer=]'s [=environment settings object/origin=], then:
-          1. Invoke <a>Reject Job Promise</a> with |job| and a "{{SecurityError}}" exception.
+          1. Invoke [=Reject Job Promise=] with |job| and "`SecurityError`".
           1. Invoke <a>Finish Job</a> with |job| and abort these steps.
       1. If the [=environment settings object/origin=] of |job|'s [=job/scope url=] is not |job|'s [=job/referrer=]'s [=environment settings object/origin=], then:
-          1. Invoke <a>Reject Job Promise</a> with |job| and a "{{SecurityError}}" exception.
+          1. Invoke [=Reject Job Promise=] with |job| and "`SecurityError`".
           1. Invoke <a>Finish Job</a> with |job| and abort these steps.
       1. Let |registration| be the result of running the <a>Get Registration</a> algorithm passing |job|'s [=job/scope url=] as the argument.
       1. If |registration| is not null, then:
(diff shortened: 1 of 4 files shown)


jungkees:
@mkruisselbrink, from https://github.com/w3c/ServiceWorker/issues/829#issuecomment-180536147,
>  So yes, removing the associated client from ServiceWorker and ServiceWorkerRegistration is probably the sanest solution
This was done in other patches already. This PR addresses using the right value for resolving the equivalent jobs' promises. Please review.

mkruisselbrink:
It kind of feels extremely pedantic (and specs in general seem to do a terribly job at specifying what Realm objects are created in), but this doesn't quite seem correct yet (other than "the value that corresponds to value" not seeming very well defined). In particular at the places where _Resolve Job Promise_ is called we currently call it with a `ServiceWorkerRegistration` instance, without specifying in which Realm that instance is created (and all of that generally is done from code that is run _in parallel_, so there isn't really an implied task/real/context either). 

Maybe the right fix would be to pass the _service worker registration_ itself, rather than a javasript object associated with it to _Resolve Job Promise_, and then have the tasks that are queued by _Resolve Job Promise_ create (or lookup? I'm not sure from reading the current spec if we're supposed to always return the same instance for the same registration in the same realm, or if a different instance can be returned sometimes) the correct instance to resolve the promise with.

Not sure how we'd then deal with the one job type that doesn't resolve with a registration, but instead with a boolean. And again all of this feels very pedantic and doesn't seem to be something that most specs get right either (probably because actually getting it right is so tricky/complicated).

Also to be truly pedantic, _Reject Job Promise_ should probably make it explicit that all the different promises get rejected with different instances of the same type of exception. As currently spec'ed the following test would pass, which I don't think is desired:

```js
let p1 = navigator.serviceWorker.register("http://example.com/foo.js");
let p2 = navigator.serviceWorker.register("http://example.com/foo.js");

p1.catch(e => e.custom = "bla");
p2.catch(e => assert_equals(e.custom, "bla"));
```

Maybe the easiest fix is to completely get rid of the _Resolve Job Promise_ and _Reject Job Promise_ algorithms, but that of course would make all the algorithms currently using them more complicated, which would also suck.

jungkees:
Thanks to https://github.com/whatwg/html/commit/58879aaf921cdde656d5cf83ebae831f382f4919, I tried to address it again. @mkruisselbrink, @domenic, please take a look.

jungkees:
Sorry for losing the commit history for @mkruisselbrink's previous review.

docs/index.bs
```diff
@@ -2521,28 +2521,37 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-      :: |reason|, an <a>exception</a>
+      :: |job|, a [=job=]
+      :: |reason|, an [=exception=]
       : Output
       :: none
 
-      1. If |job|'s [=job/client=] is not null, <a>queue a task</a> to reject |job|'s [=job/job promise=] with |reason| on |job|'s [=job/client=]'s <a>responsible event loop</a> using the <a>DOM manipulation task source</a> as the <a>task source</a>.
-      1. For each |equivalentJob| in |job|'s <a>list of equivalent jobs</a>:
-          1. If |equivalentJob|'s [=job/client=] is not null, <a>queue a task</a> to reject |equivalentJob|'s [=job/job promise=] with |reason| on |equivalentJob|'s [=job/client=]'s <a>responsible event loop</a> using the <a>DOM manipulation task source</a> as the <a>task source</a>.
+      1. If |job|'s [=job/client=] is not null, [=queue a task=], on |job|'s [=job/client=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to reject |job|'s [=job/job promise=] with an exception that is a copy of |reason|, in |job|'s [=job/client=]'s [=environment settings object/Realm=].
```
domenic: "A copy of" is a bit imprecise, but it's probably OK. I guess the truly rigorous way to do this would be to just have the reason be the type of the exception (plus maybe any data necessary to construct a good message string, but that's kind of out of scope for the spec). Then you actually create the exception object during the queued task.

This assumes _reason_ is only ever UA-created. If it is user-code-created, then we have a bigger problem.

docs/index.bs
```diff
@@ -2521,28 +2521,37 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      1. Let |convertedValue| be null.
+      1. If |job|'s [=job/client=] is not null, [=queue a task=], on |job|'s [=job/client=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to run the following substeps:
+          1. If |job|'s [=job/job type=] is either *register* or *update*, set |convertedValue| to the {{ServiceWorkerRegistration}} object that represents |value|, in |job|'s [=job/client=]'s [=environment settings object/Realm=].
+          1. Else, set |convertedValue| to |value|, in |job|'s [=job/client=]'s [=environment settings object/Realm=].
+          1. Resolve |job|'s [=job/job promise=] with |convertedValue|.
       1. For each |equivalentJob| in |job|'s <a>list of equivalent jobs</a>:
-          1. If |equivalentJob|'s [=job/client=] is not null, <a>queue a task</a> to resolve |equivalentJob|'s [=job/job promise=] with |value| on |equivalentJob|'s [=job/client=]'s <a>responsible event loop</a> using the <a>DOM manipulation task source</a> as the <a>task source</a>.
+          1. If |equivalentJob|'s [=job/client=] is null, continue to the next iteration of the loop.
```
domenic: Feel free to use `[=iteration/continue=]` instead of "continue to the next iteration of the loop"

jungkees: Addressed. I'll change the other places in a separate patch.

docs/index.bs
jungkees: Good point. I looked up and tried to use the WebIDL definitions in the follow-up commit. I referenced https://heycam.github.io/webidl/#es-creating-throwing-exceptions to [create](https://heycam.github.io/webidl/#dfn-create-exception) an exception. In that algorithm, it uses the current Realm to construct an object. So, I didn't put the target realm in the step within the queued task. Please take a look.

docs/index.bs
```diff
@@ -2544,14 +2544,14 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      1. If |job|'s [=job/client=] is not null, [=queue a task=], on |job|'s [=job/client=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to reject |job|'s [=job/job promise=] with a [=exception/create|new=] [=exception=] with |errorName| and a user agent-defined [=exception/message=].
       1. For each |equivalentJob| in |job|'s [=list of equivalent jobs=]:
-          1. If |equivalentJob|'s [=job/client=] is null, continue to the next iteration of the loop:
-          1. [=Queue a task=], on |equivalentJob|'s [=job/client=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to reject |equivalentJob|'s [=job/job promise=] with an exception that is a copy of |reason|, in |equivalentJob|'s [=job/client=]'s [=environment settings object/Realm=].
+          1. If |equivalentJob|'s [=job/client=] is null, [=iteration/continue=].
+          1. [=Queue a task=], on |equivalentJob|'s [=job/client=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to reject |equivalentJob|'s [=job/job promise=] with a [=exception/create|new=] [=exception=] with |errorName| and a user agent-defined [=exception/message=].
```
domenic: This structure doesn't quite work, because you need to say "DOMException" for the "SecurityError" cases, and not say DOMException for the TypeError case.

I'd maybe make the argument "_errorData_, the information necessary to [=exception/create|create an exception=]". Then you could say "with _job_ and `"SecurityError"` `DOMException`", or "with _job_ and `TypeError`".

Also, you should state the realm. Inside a queued task, you've "lost" the current realm, since current realm refers to that of the executing function object.

jungkees: Right. That structure didn't cover both cases. I tried it again on your comments. I left the "user agent-defined message" bit as-is. Would it make sense?

domenic: Looks great!

jungkees: Thanks for your review!

Issue #916, opened 2016-06-17
More clarity around waiting for an active worker to finish


mattto:
I'm working on a Chrome [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=616331) where we're immediately terminating the incumbent worker during activation instead of waiting for it to finish its "in-progress requests" as the spec says. Since update occurs soon after page load, with skipWaiting() we run into this issue in the wild.

I'm wondering though, what exactly is a "request". Is this referring to fetch event only or is it also intended to mean functional events?

Elsewhere the spec also mentions "in-flight requests" which are "triggered by" a worker, which seems a different thing that "in-progress requests" being handled by a worker. 

The spec also doesn't really mention what to do with new incoming requests: do they go to the incumbent worker or the soon-to-be activating worker?

My current plan is to:
- Wait for all events already dispatched to the incumbent worker to finish (subject to our usual timeout)
- All new events are put on hold until the soon-to-be activating worker is activated

How does Firefox deal with this issue?

wanderview:
It looks like if someone is holding on to the ServiceWorker object in js we will mark it redundant, but not immediately kill it.

NekR:
> All new events are put on hold until the soon-to-be activating worker is activated

By "soon-to-be activating" you mean while `e.waitUntil(...)` is waiting, right? What if it fails?

wanderview:
No, this is referring to service workers in the `waiting` state that are being promoted to the `active` state.  So things like FetchEvent have to be held in a queue until the old worker exits per the current spec.  I'm still not sure this is a good idea.

mattto:
> Is this a real problem in practice? It sounds like a lot of add complexity for what is probably fairly rare problem.

I think we're seeing this issue on Inbox and it's causing broken page loads. The page starts loading, there's sometimes one second of network quiescence that triggers our update on navigation, then skipWaiting() results in termination of the worker with unfinished fetch events.

I see Firefox uses DOMContentLoaded rather than the network quiescence heuristic to trigger update, maybe that would be safer.

wanderview:
It seems like waiting for events to clear is essentially creating a new state.

Can we instead just restructure things such that:

1. Normal criteria for going active is no controlled documents and no events being processed.
2. Calling skipWaiting bypasses the controlled document requirement, but we still defer going active until all outstanding events are done.

This lets us use the existing wait mechanism without inventing a new one.

[...]

PR #1065, opened 2017-02-03
Improve Activate with Try Activate

jungkees:
This adds concepts and algorithms to track events (all service worker
events: lifecycle events, functional events, extendable message events)
that have pending extension promises so they can be used to check if the
registration's waiting worker can be promoted.

This also changes Install algorithm and related steps to match to the
implementations behavior. That is, instead of waiting for the promotion
condition in Install algorithm, newly introduced Try Activate algorithm
just tries to activate depending on the condition and return. Try
Activate is called when:
 - A service worker is installed.
 - The last client controlled by the existing active worker is unloaded.
 - skipWaiting() is called.
 - The extend lifetime promises for the existing active worker settle.

Related issue: #916.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -168,7 +168,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker=] has an associated <dfn export id="dfn-set-of-event-types-to-handle">set of event types to handle</dfn> (a [=ordered set|set=]) whose [=item=] is an <a>event listener</a>'s event type. It is initially an empty set.
 
-    A [=/service worker=] has an associated <dfn export id="dfn-set-of-extended-events">set of extended events</dfn> (a [=ordered set|set=]) whose [=item=] is an [=event=]. It is initially an empty set.
+    A [=/service worker=] has an associated <dfn export id="dfn-set-of-extended-events">set of extended events</dfn> (a [=ordered set|set=]) whose [=item=] is an {{ExtendableEvent}}. It is initially an empty set.
 
     A [=/service worker=] has an associated <dfn export id="dfn-foreign-fetch-scopes">list of foreign fetch scopes</dfn> whose element type is a [=/URL=]. It is initially empty.
 
@@ -2701,7 +2701,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       1. Let |installFailed| be false.
       1. Let |newestWorker| be the result of running <a>Get Newest Worker</a> algorithm passing |registration| as its argument.
-      1. Let |redundantWorker| be null.
       1. Run the <a>Update Registration State</a> algorithm passing |registration|, "<code>installing</code>" and |worker| as the arguments.
       1. Run the <a>Update Worker State</a> algorithm passing |registration|'s <a>installing worker</a> and *installing* as the arguments.
       1. Assert: |job|'s [=job/job promise=] is not null.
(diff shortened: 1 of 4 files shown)


jungkees:
Chromium and Gecko implemented and shipped the resolution in #916 already. The spec follows with this PR. Please review.

/cc @mattto @wanderview @annevk @slightlyoff

docs/index.bs
```diff
@@ -364,6 +366,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
             1. Else, let it be initialized to a new {{Client}} object that represents the worker associated with |incumbentGlobal|.
             1. Let the {{ExtendableMessageEvent/ports}} attribute of |e| be initialized to |newPorts|.
             1. <a>Dispatch</a> |e| at |destination|.
+            1. Invoke [=Update Service Worker Extended Events Set=] with |serviceWorker| and |e|.
```
jakearchibald: So is this something every service worker event needs to do? Can we do this automatically when extendable events are created? Worried about every spec remembering to do this.

docs/index.bs
```diff
@@ -2775,6 +2782,18 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
 
   <section algorithm>
+    <h3 id="try-activate-algorithm"><dfn>Try Activate</dfn></h3>
+
+      : Input
+      :: |registration|, a [=/service worker registration=]
+      : Output
+      :: None
+
+      1. Assert: |registration|'s [=waiting worker=] is not null.
```
jakearchibald: This is called when promises passed to `event.waitUntil` settle, which can happen when there isn't a waiting worker, so isn't it common for the assertion to fail?

docs/index.bs
```diff
@@ -2742,10 +2752,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       1. If |registration|'s <a>waiting worker</a> is null, abort these steps.
       1. Let |redundantWorker| be null.
-      1. If |registration|'s <a>active worker</a> is not null, then:
-          1. Set |redundantWorker| to |registration|'s <a>active worker</a>.
-          1. Wait for |redundantWorker| to finish handling any in-progress requests.
-          1. <a lt="Terminate Service Worker">Terminate</a> |redundantWorker|.
+      1. If |registration|'s [=active worker=] is not null, |redundantWorker| be |registration|'s [=active worker=].
```
jakearchibald: nit: Missing let/set before `|redundantWorker|`?

docs/index.bs
```diff
@@ -2742,10 +2752,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       1. If |registration|'s <a>waiting worker</a> is null, abort these steps.
       1. Let |redundantWorker| be null.
-      1. If |registration|'s <a>active worker</a> is not null, then:
-          1. Set |redundantWorker| to |registration|'s <a>active worker</a>.
-          1. Wait for |redundantWorker| to finish handling any in-progress requests.
-          1. <a lt="Terminate Service Worker">Terminate</a> |redundantWorker|.
```
jakearchibald: When are these workers terminated now?

docs/index.bs
```diff
@@ -3100,6 +3122,21 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   </section>
 
   <section algorithm>
+    <h3 id="update-service-worker-extended-events-set-algorithm"><dfn>Update Service Worker Extended Events Set</dfn></h3>
```
jakearchibald: I'm a little worried about the lifecycle of this, but I might be worrying over nothing.

If a service worker abruptly terminates (due to long-running tasks or whatever), what happens to these events? Should we clear the set of extended events on service worker terminate?

I guess we'll need to handle timeouts to `waitUntil` extension promises at some point.

docs/index.bs
jungkees: We gate all the service worker events through Handle Fetch, Handle Foreign Fetch, Handle Functional Events, and ServiceWorker.postMessage(). So, maintenance would be okay. Also, I intended it to be called after dispatching an event returns.

docs/index.bs
jungkees: I thought it would be not needed when Try Activate successfully called Activate. But you're right. It'd be safer to call Terminate Service Worker before get it redundant. In the follow-up commit, I addressed this point and also moved steps for invoking Update Worker State for the redundant works up, which both Chromium and Gecko do.

docs/index.bs
jungkees: Addressed.

docs/index.bs
```diff
@@ -168,6 +168,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker=] has an associated <dfn export id="dfn-set-of-event-types-to-handle">set of event types to handle</dfn> (a [=ordered set|set=]) whose [=item=] is an <a>event listener</a>'s event type. It is initially an empty set.
 
+    A [=/service worker=] has an associated <dfn export id="dfn-set-of-extended-events">set of extended events</dfn> (a [=ordered set|set=]) whose [=item=] is an [=event=]. It is initially an empty set.
```
jungkees: Yes, that narrow typing would be better. Done.

docs/index.bs
jungkees: Right. I missed that point. Made it return null when waiting worker is null there.

docs/index.bs
jungkees: Good point! I added a step to Terminate Service Worker to remove all the items in the service worker's set of extended events.

For extended events, I thought we decided to leave them run to completion: https://github.com/w3c/ServiceWorker/issues/916#issuecomment-229332311. So, I removed the steps that abort the in-progress requests.

To clarify what happens when a service worker abruptly terminates, Terminate Service Worker resets the set of extended events to an empty set (done in this PR), retains the tasks (for functional events) in the queue so the next worker run will execute them.

jungkees:
@jakearchibald, thanks for review! I uploaded a follow-up commit. PTAL.

docs/index.bs
jakearchibald: I think I'm missing something. How would this work for `push` events?

docs/index.bs
jakearchibald: Ah, I missed that. Sounds good!

docs/index.bs
jungkees: Handle Functional Event runs https://github.com/w3c/ServiceWorker/pull/1065/files#diff-27b79860afe28f01aed4f1f6228367faR3082 after the callback steps given by other specs that are supposed to dispatch the functional events.

Issue #880, opened 2016-04-12
CORS pre-flight for foreign fetch


mkruisselbrink:
At the F2F the point was brought up if foreign fetch should involve some kind of pre-flight like regular cross origin CORS requests has. This could take the form of extending the `ForeignFetchOptions` dictionary to have additional fields to limit which requests can get intercepted by the foreign fetch service worker, such as allowed methods and headers.

mkruisselbrink:
@wanderview does this sum up your concerns?

wanderview:
Yea, I think some kind of safety belt seems reasonable since I expect many writers of ForeignFetchEvent service workers won't understand all the different types of requests that could be thrown at it.  Having to explicitly opt in to handling POSTs would be good to prevent naive pass-through code accidentally sending posts to the server without CORS protections.

Thanks.

annevk:
I thought the idea was that registering for foreign fetch was sufficient for this. Note that with on the web you can POST without CORS preflight.

wanderview:
In theory yes you are opting in, but I don't think devs will realize all the kinds of things which might get maliciously sent.  I think we should safety belts similar to what cors provides to protect the server.

This could just be implementing cors at this layer which was suggested by Alex.

annevk:
I really don't see why. We already have all those safety belts with the protocol we designed around respondWith(). The preflight only exists because servers might not know about CORS at all. That's not the case for entirely new features.

delapuente:
@wanderview can you explain an example of malicious request where we should use these belts and `respondWith()` protocols are not enough? @annevk , what happen if one of this request reaches the network?

annevk:
@delapuente then it would go through the normal non-foreign-fetch code path, right?

delapuente:
And the usual CORS mechanisms actuated and all is fine. Isn't it?

annevk:
Yeah.

wanderview:
If a site only tests with GET requests and just blanket passes requests through with respondWith(evt.request).  Once they go through the foreign fetch they are considered from the new origin, past the CORS boundary.  If some sends a POST with a dangerous header this naive foreign fetch SW script will send it to server without any pre-flight.

delapuente:
Do you mean this?
```js
self.onforeignfetch = evt => {
  var evilRequest = evt.request;
  evt.respondWith(fetch(evilRequest));
};
```

wanderview:
Essentially yes.

annevk:
[...]

PR #1067, opened 2017-02-03
Add check for CORS-safelisted methods and headers to foreign fetch.

mkruisselbrink:
We probably ultimately want to add the extra options to
ForeignFetchOptions as discussed in #880, but for now just limiting
headers and methods to CORS safelisted values seems reasonable, and
matches chrome's implementation.


mkruisselbrink:
Tests for this behavior are in chrome's foreign fetch tests, which I'm in the process of upstreaming to WPT (and if we decide not to merge this, I'll of course take out those tests from what I upstream).

mkruisselbrink:
> Extending ForeignFetchOptions and sorting out the behavior for extra headers seem to be still needed, right?

Yes, to lift the restrictions this PR introduces we'd still want to add options to ForeignFetchOptions to support intercepting requests with extra headers or non-standard methods.

PR #1070, opened 2017-02-13
Correcting usecache attribute. Fixes #1069

### LINKED ISSUE #1069: useCache attribute design seems very very wrong
> If useCache is not a valid boolean value, queue a task to fire an event named error at the link element, and abort these steps.

This is *not* how HTML boolean attributes work. You should never do `usecache="true"` or `usecache="false"`.

Instead this step and the previous should just check for the presence of the usecache attribute. There is no such thing as an invalid value. The attribute is either present or absent; it's value does not matter. That's how boolean attributes work.

This is especially important since the `useCache` IDL attribute is supposed to reflect the content attribute. That will work very very poorly with the above processing model.

jakearchibald:
I also did the same for the header. I assume we handle booleans the same way there?

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1745,7 +1745,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. If the "<code>scope</code>" <a>target attribute</a> of the <code>Link</code> header is present, set |scopeURL| to the result of <a lt="URL parser">parsing</a> the "<code>scope</code>" <a>target attribute</a> with |scriptURL|.
       1. Let |workerType| be the "<code>workertype</code>" <a>target attribute</a> of the <code>Link</code> header, or "<code>classic</code>" if no such attribute is present.
       1. If |workerType| is not a valid {{WorkerType}} value, abort these steps.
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -1745,8 +1745,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. If the "<code>scope</code>" <a>target attribute</a> of the <code>Link</code> header is present, set |scopeURL| to the result of <a lt="URL parser">parsing</a> the "<code>scope</code>" <a>target attribute</a> with |scriptURL|.
       1. Let |workerType| be the "<code>workertype</code>" <a>target attribute</a> of the <code>Link</code> header, or "<code>classic</code>" if no such attribute is present.
       1. If |workerType| is not a valid {{WorkerType}} value, abort these steps.
-      1. Let |useCache| be the "<code>usecache</code>" <a>target attribute</a> of the <code>Link</code> header, or false if no such attribute is present.
-      1. If |useCache| is not a valid boolean value, abort these steps.
+      1. Let |useCache| be be true if the the <code>Link</code> header has a <a>target attribute</a> named "<code>usecache</code>", otherwise false.
```
jungkees: Nits: s/be be/be/, s/the the/the/

jungkees:
LGTM with the above nit. Please include the bikesheded HTML document when merging. Thanks!

PR #1071, opened 2017-02-14
Add specific Processing headings + minor editorial

sideshowbarker:
This splits the “Link type "serviceworker"” ≫ “Processing” section into
two sections with separate headings:

* Processing the Link header
* Processing the link element

It also includes a minor editorial change to add the word “or” in a
place where lack of it makes the text ambiguous; this sentence:

> When a serviceworker link’s link element is inserted into a document,
> a serviceworker link is created on a link element that is already in a
> document tree, or the href or scope attributes of the link element of a
> serviceworker link is changed, the user agent should run these steps

...could be read as saying:

> When a serviceworker link’s link element is inserted into a document,
> **then** a serviceworker link is created on a link element...

When what it actually seems to mean is:

> When a serviceworker link’s link element is inserted into a document,
> **or** a serviceworker link is created on a link element...

That “or” is somewhat implied from the rest of the sentence, but this
this change just inserts an explicit “or” to make it more clear.


sideshowbarker:

jungkees:
LGTM.

> a minor editorial change to add the word “or”

I think the original sentence was still okay to understand within the context but agree to this change with no strong opinion. /cc @mkruisselbrink

jungkees:
Thanks @sideshowbarker!

mkruisselbrink:
Thanks!

Issue #1036, opened 2016-12-19
Support returning ServiceWorker Client environments from `Clients.matchAll()`


wanderview:
Now that we support matching uncontrolled Client environments it seems we should reasonably be able to get a Client object for a ServiceWorker.  I believe the spec mostly already supports this as it doesn't explicitly exclude ServiceWorker environments.  We just need to add a "serviceworker" enum value to ClientType.

jakearchibald:
I like the idea of eventually making the clients API available to documents, so yeah, it makes sense to surface service worker clients here.

wanderview:
Well, this is possible even without window.clients.  If you have two service workers registered for the same origin and one does `clients.matchAll({ includeUncontrolled: true})` it should in theory include the other service worker as a Client if it happens to be running.

jungkees:
I'll review https://github.com/w3c/ServiceWorker/pull/1074.

> I like the idea of eventually making the clients API available to documents,

That seems like a good direction, but I'm worried the API as-is seems a bit awkward for the case between non-service worker clients. `window.clients.get(idOfAnotherWindow).then(client => client.postMessage(msg));` will land on `navigator.serviceWorker`. This seems to be a separate issue if we would work on that.

wanderview:
Yes, that is issue #955.

> That seems like a good direction, but I'm worried the API as-is seems a bit awkward for the case between non-service worker clients. window.clients.get(idOfAnotherWindow).then(client => client.postMessage(msg)); will land on navigator.serviceWorker. This seems to be a separate issue if we would work on that.

My proposal to fix that is in this comment:

https://github.com/w3c/ServiceWorker/issues/955#issuecomment-261056530

TravisGit:
https://github.com/wanderview/ServiceWorker/commit/e4205076e7fb59356655bcc47c6ce5a8048afb95

monica-ch:
[Bug triage] keeping it open, but think of other cases of unnecessary communication with service workers via post messages

PR #1074, opened 2017-02-15
Add Client type for ServiceWorkerGlobalScope environments

wanderview:
This pull request is broken up into two commits:

1. First we rename the current 'service worker client' concept to just 'client'.  I think this makes sense because the client concept is more general than just service workers and we would like to expose it to other places in the future.  Also, I need the 'service worker client' name to represent my new client type.
2. Add the service worker client type.

This PR addresses issue #1036.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -231,7 +231,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A <dfn export id="dfn-sharedworker-client">shared worker client</dfn> is a [=/client=] whose [=environment settings object/global object=] is a {{SharedWorkerGlobalScope}} object.
 
-    A <dfn export id="dfn-worker-client">worker client</dfn> is either a <a>dedicated worker client</a> or a <a>shared worker client</a>.
+    A <dfn export id="dfn-serviceworker-client">service worker client</dfn> is a [=/client=] whose [=environment settings object/global object=] is a {{ServiceWorkerGlobalScope}} object.
+
+    A <dfn export id="dfn-worker-client">worker client</dfn> is a <a>dedicated worker client</a>, a <a>shared worker client</a>, or a <a>service worker client</a>.
   </section>
 
   <section>
@@ -999,6 +1001,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. If the [=context object=]'s [=Client/client=] is a type of [=environment=] or is a [=window client=], return {{ClientType/"window"}}.
         1. Else if the [=context object=]'s [=Client/client=] is a [=dedicated worker client=], return {{ClientType/"worker"}}.
         1. Else if the [=context object=]'s [=Client/client=] is a [=shared worker client=], return {{ClientType/"sharedworker"}}.
+        1. Else if the [=context object=]'s [=Client/client=] is a [=service worker client=], return {{ClientType/"serviceworker"}}.
     </section>
 
     <section>
@@ -1129,6 +1132,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         "window",
         "worker",
         "sharedworker",
+        "serviceworker",
         "all"
       };
     </pre>
@@ -1189,6 +1193,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                 1. Add |client| to |targetClients|.
             1. Let |matchedClients| be an empty array.
             1. For each [=/client=] |client| in |targetClients|:
+                1. If |client|'s [=environment settings object/global object=] is the [=current global object=], then continue to the next iteration of the loop.
                 1. If |options|.{{ClientQueryOptions/type}} is {{ClientType/"window"}} or {{ClientType/"all"}}, and |client| is a type of <a>environment</a> or is a <a>window client</a>, then:
                     1. Let |browsingContext| be null.
                     1. Let |isClientEnumerable| be true.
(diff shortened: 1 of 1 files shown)


docs/index.bs
```diff
@@ -214,36 +214,36 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
     </section>
   </section>
 
-  <section dfn-for="service worker client">
-    <h3 id="service-worker-client-concept">Service Worker Client</h3>
+  <section dfn-for="client">
+    <h3 id="client-concept">Client</h3>
 
-    A <dfn export id="dfn-service-worker-client" for="">service worker client</dfn> is a type of <a>environment</a> or <a>environment settings object</a>.
+    A <dfn export id="dfn-client" for="">client</dfn> is a type of <a>environment</a> or <a>environment settings object</a>.
```
jungkees: Question to @annevk in spec maintenance perspective: generally when we change definitions, changing dfn's id would be okay? (They may break existing links.) Or should we retain the ids even if they sound wrong after the change?

docs/index.bs
```diff
@@ -1260,15 +1260,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. If the [=ServiceWorkerGlobalScope/service worker=] is not an <a>active worker</a>, return a <a>promise</a> rejected with an "{{InvalidStateError}}" exception.
         1. Let |promise| be a new <a>promise</a>.
         1. Run the following substeps <a>in parallel</a>:
-            1. For each [=/service worker client=] |client| whose [=service worker client/origin=] is the <a lt="same origin">same</a> as the [=ServiceWorkerGlobalScope/service worker=]'s [=environment settings object/origin=]:
+            1. For each [=/client=] |client| whose [=client/origin=] is the <a lt="same origin">same</a> as the [=ServiceWorkerGlobalScope/service worker=]'s [=environment settings object/origin=]:
```
jungkees: It seems the service worker client itself should be excluded from claiming.

docs/index.bs
```diff
@@ -1210,7 +1215,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                     1. If |isClientEnumerable| is true, then:
                         1. Let |windowClient| be the result of running <a>Create Window Client</a> algorithm with |client|, |visibilityState|, |focusState|, and |ancestorOrigins| as the arguments.
                         1. Add |windowClient| to |matchedClients|.
-                1. Else if |options|.{{ClientQueryOptions/type}} is {{ClientType/"worker"}} or {{ClientType/"all"}} and |client| is a <a>dedicated worker client</a>, or |options|.{{ClientQueryOptions/type}} is {{ClientType/"sharedworker"}} or {{ClientType/"all"}} and |client| is a <a>shared worker client</a>, then:
+                1. Else if |options|.{{ClientQueryOptions/type}} is {{ClientType/"worker"}} or {{ClientType/"all"}} and |client| is a <a>dedicated worker client</a>, then:
```
jungkees: The sub steps of the "else if" steps seem to be repeated here. How about we make the above window client "if" block continue to the next iteration after its own substeps. And do the following for worker blocks:

Add the result of running <a>Create Client</a> algorithm with |client| to |matchedClients| if any of the following are true:
 - |options|.{{ClientQueryOptions/type}} is {{ClientType/"worker"}} or {{ClientType/"all"}} and |client| is a <a>dedicated worker client</a>
 - |options|.{{ClientQueryOptions/type}} is {{ClientType/"sharedworker"}} or {{ClientType/"all"}} and |client| is a <a>shared worker client</a>
 - |options|.{{ClientQueryOptions/type}} is {{ClientType/"serviceworker"}} or {{ClientType/"all"}} and |client| is a <a>service worker client</a>

jungkees:
Additionally, I think we need to sort out how `client.reserved` works for service worker clients. `reserved` becomes false when client's [execution ready flag](https://html.spec.whatwg.org/multipage/webappapis.html#concept-environment-execution-ready-flag) is set. IMO, the flag should be set during the first service worker run in Run Service Worker algorithm. If you need my help here, let me know.

ylafon:

docs/index.bs
annevk: I'd generally try to preserve IDs somehow since links that don't work are annoying. bikeshed has an `oldids` feature you could maybe use if you really wanted to change the ID.

jungkees: `oldids` sounds good. @wanderview, please preserve the IDs. I'll at some point sort out the IDs that need changes using `oldids`.

docs/index.bs
```diff
@@ -214,36 +214,38 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      A user agent *must* persistently keep a list of <a>registered</a> [=/service worker registrations=] unless otherwise they are explicitly <a>unregistered</a>. A user agent has a <a>scope to registration map</a> that stores the entries of the tuple of [=/service worker registration=]'s [=service worker registration/scope url=] and the corresponding [=/service worker registration=]. The lifetime of [=/service worker registrations=] is beyond that of the {{ServiceWorkerRegistration}} objects which represent them within the lifetime of their corresponding [=/clients=].
     </section>
   </section>
 
-  <section dfn-for="service worker client">
-    <h3 id="service-worker-client-concept">Service Worker Client</h3>
+  <section dfn-for="client">
+    <h3 id="client-concept">Client</h3>
```
jungkees: Preserved the old id here.

docs/index.bs
```diff
@@ -969,36 +971,37 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       };
     </pre>
 
-    A {{Client}} object has an associated <dfn for="Client">service worker client</dfn> (a [=/service worker client=]).
+    A {{Client}} object has an associated <dfn for="Client">client</dfn> (a [=/client=]).
 
     A {{Client}} object has an associated <dfn for="Client">reserved state</dfn>, which is either true or false.
 
-    A {{WindowClient}} object has an associated <dfn id="dfn-service-worker-client-visibilitystate">visibility state</dfn>, which is one of {{Document/visibilityState}} attribute value.
+    A {{WindowClient}} object has an associated <dfn id="dfn-client-visibilitystate">visibility state</dfn>, which is one of {{Document/visibilityState}} attribute value.
```
jungkees: Preserved the old id here.

docs/index.bs
```diff
@@ -969,36 +971,37 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
 
-    A {{WindowClient}} object has an associated <dfn id="dfn-service-worker-client-focusstate">focus state</dfn>, which is either true or false (initially false).
+    A {{WindowClient}} object has an associated <dfn id="dfn-focusstate">focus state</dfn>, which is either true or false (initially false).
```
jungkees: Preserved the old id here.

PR #1079, opened 2017-02-23
Introduce Try Clear Registration; Delay activation

### LINKED ISSUE #916: More clarity around waiting for an active worker to finish

jungkees:
This is a follow-up change, to [1], which addresses the issues posed in
[2] and the comments afterwards.

This change makes the callers to the Clear Registration algorithm to
check if no client is using the registration and the registration's
associated service workers have no pending extended events before
invoking the algorithm. Unregister, Handle Service Worker Client Unload,
and the microtasks queued by the extension promises' settlement handlers
invoke Try Clear Registration.

Other error handling steps on registrations that have no associated
worker and Handle User Agent Shutdown invoke Clear Registration directly
as they do.

[1] https://github.com/w3c/ServiceWorker/commit/f1a1120ff07120a2e8f311fb61f429c9a74db1ef
[2] https://github.com/w3c/ServiceWorker/issues/916#issuecomment-237272060

Summary of the issues: https://github.com/w3c/ServiceWorker/issues/916#issuecomment-281274267.

Fixes #916.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1323,7 +1323,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
         1. Upon [=upon fulfillment|fulfillment=] or [=upon rejection|rejection=] of |f|, [=queue a microtask=] to run these substeps:
             1. Decrement the [=ExtendableEvent/pending promises count=] by one.
-            1. Invoke [=Try Activate=] with the [=context object=]'s [=relevant global object=]'s associated [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=].
+            1. Let |registration| be the [=context object=]'s [=relevant global object=]'s associated [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=].
+            1. If |registration|'s [=uninstalling flag=] is set, invoke [=Try Clear Registration=] with |registration|.
+            1. If |registration| is not null, invoke [=Try Activate=] with |registration|.
 
         The user agent *should not* <a lt="terminate service worker">terminate</a> the [=/service worker=] associated with |event|'s <a>relevant settings object</a>'s [=environment settings object/global object=] when |event|'s [=dispatch flag=] is set or |event|'s [=ExtendableEvent/pending promises count=] is not zero. However, the user agent *may* impose a time limit to this lifetime extension.
     </section>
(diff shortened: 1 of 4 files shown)


jungkees:
@jakearchibald, @mkruisselbrink, This PR is to address https://github.com/w3c/ServiceWorker/issues/916#issuecomment-281274267. PTAL.

/cc @wanderview @mattto @annevk @slightlyoff

jungkees:
I left Handle User Agent Shutdown unchanged as I think it shouldn't miss the chance to update the registrations (clear and activate operations).

docs/index.bs
```diff
@@ -1323,7 +1323,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
         1. Upon [=upon fulfillment|fulfillment=] or [=upon rejection|rejection=] of |f|, [=queue a microtask=] to run these substeps:
             1. Decrement the [=ExtendableEvent/pending promises count=] by one.
-            1. Invoke [=Try Activate=] with the [=context object=]'s [=relevant global object=]'s associated [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=].
+            1. Let |registration| be the [=context object=]'s [=relevant global object=]'s associated [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=].
+            1. If |registration|'s [=uninstalling flag=] is set, invoke [=Try Clear Registration=] with |registration|.
```
jakearchibald: Should this check to see if the registration is not null before checking the flag?

jungkees: I think the event has been alive up to this point so no Try Clear Registration should have cleared the registration yet. So, it seems we can assert that the registration must not be null here.

jungkees:
@jakearchibald, thanks for the review!

PR #1084, opened 2017-02-28
Fix old text stating Client is a browsing context. Fixes #1083

### LINKED ISSUE #1083: SW spec associates Client objects with browsing context in a couple places
A Client is defined to reference an environment or environment settings object here:

https://w3c.github.io/ServiceWorker/#dfn-service-worker-client

But the spec associates the Client with a browsing context in a couple places still:

* Step 10.5 of https://w3c.github.io/ServiceWorker/#service-worker-postmessage
* `clients` comment in https://w3c.github.io/ServiceWorker/#serviceworkerglobalscope-interface

wanderview:


ylafon:

docs/index.bs
```diff
@@ -362,7 +362,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
             1. Let the {{ExtendableMessageEvent/data}} attribute of |e| be initialized to |clonedMessage|.
             1. Let the {{ExtendableMessageEvent/origin}} attribute of |e| be initialized to the <a lt="Unicode serialization of an origin">Unicode serialization</a> of |incumbentSettings|'s [=environment settings object/origin=].
             1. If |incumbentGlobal| is a {{ServiceWorkerGlobalScope}} object, let the {{ExtendableMessageEvent/source}} attribute of |e| be initialized to a new {{ServiceWorker}} object that represents |incumbentGlobal|'s [=ServiceWorkerGlobalScope/service worker=].
-            1. Else if |incumbentGlobal| is a {{Window}} object, let the {{ExtendableMessageEvent/source}} attribute of |e| be initialized to a new {{WindowClient}} object that represents |incumbentGlobal|'s [=/browsing context=].
+            1. Else if |incumbentGlobal| is a {{Window}} object, let the {{ExtendableMessageEvent/source}} attribute of |e| be initialized to a new {{WindowClient}} object that represents |incumbentGlobal|'s [=/environment settings object=].
```
jungkees: @wanderview, thanks for bringing this discussion first of all. Actually, I intended to describe this as the created windowClient is a thing that represents a browsing context (seems like a Window is a more relevant term?) instead of an environment settings object here because an WindowClient object captures the visual properties as well as the settings object. Reviewing this step now, I agree it's not really accurate in the implementation point of view. Maybe we should invoke [Create Window Client](https://w3c.github.io/ServiceWorker/#create-window-client) algorithm with the settings object and the related states retrieved from the Window here to create a WindowClient? WDYT?

wanderview: > Actually, I intended to describe this as the created windowClient is a thing that represents a browsing context (seems like a Window is a more relevant term?) instead of an environment settings object here because an WindowClient object captures the visual properties as well as the settings object.

I don't think this is correct.

A browsing context can have multiple environment settings objects over time as navigations occur.  Calling `document.open()` creates a new environment settings object within the browsing context.

Per my understanding of our intent is that each navigation should create a new Client object.  This suggests that the Client must be associated with the environment settings object.

The Client does have access to the visual attributes of the browsing context, but that's because the environment settings object must have a browsing context.  We get the browsing context from the environment settings object and then get the visual attributes from the browsing context.

I believe our intent and implicit in the spec is that we do not expose inactive documents or environment settings objects as Client objects.  For example, an environment settings objects in the bfcache would not be exposed.  I believe this requirement for active documents ensures there is a browsing context available.

I feel strongly at this point that the Client should represent the environment settings object and I'm not sure we can achieve our goals if we try to map it to the browsing context.

jungkees: I understand what you mean and agree that the browsing context doesn't fit here. I was concerned more about whether saying "a new WindowClient object that represents incumbentGlobal’s environment settings object" would be enough or explicitly invoking the creation algorithm would be needed. Anyway, I'm happy to merge this change as the browsing context is not correct here.

jungkees:
@wanderview, thanks for your contribution! :)

docs/index.bs
wanderview: Sorry for my confusion.  I probably over-reacted in my response here because I've been having this discussion with people inside mozilla recently.

PR #1097, opened 2017-03-30
Cast `resolveWith` argument to a Promise

jugglinmike:
The non-normative Note that describes the `event.respondWith` algorithm
states:

> Developers can set the argument r with either a promise that resolves
> with a Response object or a Response object (which is automatically
> cast to a promise). [...]

Previously, the algorithm did not explicitly perform such type casting.

Insert an additional step to harmonize the algorithm and its
description.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1457,12 +1457,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. If the [=FetchEvent/respond-with entered flag=] is set, then:
             1. <a>Throw</a> an "{{InvalidStateError}}" exception.
             1. Abort these steps.
-        1. Add |r| to the <a>extend lifetime promises</a>.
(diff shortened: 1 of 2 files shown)


ylafon:

mkruisselbrink:
I don't believe this change should be necessary. WebIDL already takes care of more-or-less wrapping non-promise values in Promise.resolve() for methods that are defined as taking a Promise&lt;T>.

jakearchibald:
That was my understanding too. @domenic can you confirm? `void respondWith(Promise<Response> r);` - webidl will take care of casting `r` to a promise (via `Promise.resolve()`) right?

jakearchibald:
Confirmed f2f with @domenic.

@jugglinmike - thanks for the PR, but I don't think we need these steps.

jugglinmike:
No problem, and my apologies for the noise :)

PR #1106, opened 2017-04-05
Use HTML spec JS mime types. Fixes #1033

### LINKED ISSUE #1033: Why are only a subset of JS mime types allowed?
> Extract a MIME type from the response’s header list. If this MIME type (ignoring parameters) is not one of text/javascript, application/x-javascript, and application/javascript, then:

Why doesn't this use [JavaScript MIME type](https://html.spec.whatwg.org/multipage/scripting.html#javascript-mime-type) definition that the rest of the platform uses?

jakearchibald:


jakearchibald:
Tests: https://github.com/w3c/web-platform-tests/pull/5376

PR #1107, opened 2017-04-07
Changing useCache boolean to updateViaCache enum

jakearchibald:

PR #1110, opened 2017-04-09
Update the “Link type "serviceworker"” section

### LINKED ISSUE #1073: Update to reference HTML for link rel=serviceworker
@sideshowbarker added formal definitions for link rel=serviceworker to HTML, containing e.g. authoring conformance requirements, in https://github.com/whatwg/html/pull/2356. The processing model of course stays in this spec. But there are some small tweaks that can be done for integration purposes.

- Update the introductory paragraph in https://w3c.github.io/ServiceWorker/#link-type-serviceworker. You still need to define the term "serviceworker link" but I'd say something like how it can be created with either a `Link` header or a HTML `<{link}>` element with `<{link/rel}>` equal to `"serviceworker"`. Ideally that last should link to https://html.spec.whatwg.org/multipage/semantics.html#link-type-serviceworker.
- https://w3c.github.io/ServiceWorker/#link-element-interface-extensions can be removed. That should hopefully make all links to the content attributes automatically go to their definitions in HTML.
- Step 7 should change to "Let workerType be the state of the `<{link/workertype}>` attribute". HTML now takes care of the missing and invalid value default.
  - Either step 8 should be removed (normative change!), or we should update HTML to have an invalid value default state (say "invalid"), which you can then throw an error on. Let me know what you think is best.

sideshowbarker:
* Add new section “Declaring a "serviceworker" Link header”
* Drop normative requirement to fire an error if link[workertype] value is
  not a valid worker type; instead just abort without firing an error.
* Drop the partial interface IDL for the HTMLLinkElement interface (the
  definitions for the IDL attributes have already been merged directly
  into the HTMLLinkElement interface IDL definition in the HTML spec).

Fixes #1073

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1777,10 +1777,10 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   <section>
     <h3 id="serviceworker-link-header">Declaring a "serviceworker" <code>Link</code> header</h3>
 
-    A <a>serviceworker link</a> can be declared using a `Link` header</a> [[!RFC5988]] with "`serviceworker`" as the value of the "`rel`" parameter and a [=service worker/script url=] as the <a>target IRI</a>, and the following optional <a>target attributes</a>:
+    A <a>serviceworker link</a> can be declared using a `Link` header</a> [[!RFC5988]] with "`serviceworker`" as the value of the "`rel`" parameter and a [=URL serializer|serialized=] [=service worker/script url=] inside angle brackets ("`<>`") as the <a>target IRI</a>, and the following optional <a>target attributes</a>:
 
     : `scope`
-    :: Value: A [=service worker registration/scope url=].
+    :: Value: A [=URL serializer|serialized=] [=service worker registration/scope url=].
 
     : `workertype`
     :: Value: A [=job/worker type=].
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -86,6 +86,7 @@ spec: rfc5988; urlPrefix: https://tools.ietf.org/html/rfc5988
     type: dfn
         text: context IRI; url: section-5.2
         text: target attribute; url: section-5.4
+        text: target attributes; url: section-5.4
         text: target IRI; url: section-5.1
 
 spec: rfc7230; urlPrefix: https://tools.ietf.org/html/rfc7230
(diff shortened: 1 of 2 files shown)


sideshowbarker:
This needs to wait for #1107 (_Changing useCache boolean to updateViaCache enum_) and needs a further tweak/update if/when https://github.com/whatwg/html/pull/2517 is merged.

sideshowbarker:
I am wondering if there’s an inconsistency in behavior here that we should fix.

Specifically, in this PR we’re dropping the following normative requirement:

> If _workerType_ is not a valid `WorkerType` value, queue a task to fire an event named `error` at the `link` element, and abort these steps.

That’s for the case of the `link` element with an a `workertype` attribute whose value is not a valid `WorkerType` value. But in contrast, in the case of `Link` header with a `workertype` attribute whose value is not a valid `WorkerType` value, the spec still says this:

> Let _workerType_ be the "`workertype`" target attribute of the `Link` header, or "`classic`" if no such attribute is present.
> If _workerType_ is not a valid `WorkerType` value, abort these steps.

So we’re aborting when the `Link` header provides a `workertype` value that’s not valid, but we’re not aborting in when a `link` element provides a `workertype` value that’s not valid.

Is that an inconsistency in behavior that we should fix?

domenic:
> Is that an inconsistency in behavior that we should fix?

I think we could go either way. It's very common for HTML to have an invalid value default for enumerated attributes; I can't think of any that fall back to an error behavior. But I'm not sure about HTTP link parameters. Maybe @mnot can lend us his HTTP knowledge.

Re-stating things: I think an inconsistency is fine if each behavior makes sense within its own domain (HTML vs. HTTP).

domenic:
Wait, no, now I am not so sure, because I started comparing it to HTML on another dimension: script type="" attributes. Those are no-ops for unknown types. (They are not errors though.) Maybe we should be consistent with that...

ylafon:

jungkees:
@mkruisselbrink, do you have bandwidth to review this PR?

docs/index.bs
```diff
@@ -1778,8 +1803,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |scriptURL| be the result of <a lt="URL parser">parsing</a> the <{link/href}> attribute with the <{link}> element's <a>node document</a>'s <a>document base URL</a>.
       1. Let |scopeURL| be null.
       1. If the <{link/scope}> attribute is present, set |scopeURL| to the result of <a lt="URL parser">parsing</a> the <{link/scope}> attribute with the <{link}> element's <a>node document</a>'s <a>document base URL</a>.
-      1. Let |workerType| be the <{link/workertype}> attribute, or "<code>classic</code>" if the <{link/workertype}> attribute is omitted.
-      1. If |workerType| is not a valid {{WorkerType}} value, <a>queue a task</a> to <a>fire an event</a> named <code>error</code> at the <{link}> element, and abort these steps.
```
jungkees: Removing this condition entirely seems not desirable. Keep it as-is or at least abort would be needed I guess.

jungkees:
@sideshowbarker, both https://github.com/w3c/ServiceWorker/pull/1107 and https://github.com/whatwg/html/pull/2517 were merged. So, I think we can continue on this PR. It looks good with needing to have a decision on https://github.com/w3c/ServiceWorker/pull/1110#discussion_r120542687.

/cc @domenic, @mnot

sideshowbarker:
> both https://github.com/w3c/ServiceWorker/pull/1107 and https://github.com/whatwg/html/pull/2517 were merged. So, I think we can continue on this PR. It looks good with needing to have a decision on https://github.com/w3c/ServiceWorker/pull/1110#discussion_r120542687

OK, great—so @domenic, what do we want to do for https://github.com/w3c/ServiceWorker/pull/1110#discussion_r120542687?

domenic:
Hmm, I did seem to forget that.

I have started to think that the best thing to do is to no-op on invalid worker types. This means we need another HTML spec tweak (maybe update https://github.com/whatwg/html/pull/2740) to add a third state, _invalid_, and make that the invalid value default. Then this spec can no-op when it encounters the _invalid_ state.

domenic:
https://github.com/whatwg/html/pull/2740 is updated so you can now count on the attribute being in one of three states: classic, module, or invalid.

sideshowbarker:
@jungkees after you (re)review this I’ll rebase so it’ll be ready to merge

docs/index.bs
```diff
@@ -1744,12 +1745,36 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+  A [=/service worker registration=] and its [=service worker registration/scope url=] are created by a <dfn id="dfn-serviceworker-link">serviceworker link</dfn>, which is declared using a <a href="#serviceworker-link-header">"serviceworker" `Link` header</a> or a <{link}> element whose <{link/rel}> attribute contains the keyword "<{link/rel/serviceworker}>".
 
   <section>
-    <h3 id="link-header-processing">Processing the <code>Link</code> header</h3>
+    <h3 id="serviceworker-link-header">Declaring a "serviceworker" <code>Link</code> header</h3>
+
+    A <a>serviceworker link</a> can be declared using a `Link` header</a> [[!RFC5988]] with "`serviceworker`" as the value of the "`rel`" parameter and a [=service worker/script url=] as the <a>target IRI</a>, and the following optional <a>target attributes</a>:
+
+    : `scope`
+    :: Value: A [=service worker registration/scope url=].
```
jungkees: I'm not sure but would like to check if some serialized url should be used for this? scope url is a URL record.

Can @annevk help?

jungkees:
@sideshowbarker, LGTM with checking https://github.com/w3c/ServiceWorker/pull/1110#discussion_r121320854. Thanks!

docs/index.bs
jungkees: @sideshowbarker, checking https://tools.ietf.org/html/rfc5988#section-5, I think we can just say ":: Value: A [=service worker registration/scope url=] as a URI-Reference" or something like that. WDYT?

And for the script url above, "a [=service worker/script url=] as a URI-Reference inside angle brackets ("<>")" seems clearer I guess. Please correct me if I'm wrong.

mnot:
Sorry, missed that. There aren't any special considerations for `Link` header attributes regarding defaults; if you want to define them, and make them different in different places, that's up to you.

docs/index.bs
domenic: I think it's better not to reference obsoleted concepts like URI. Instead I'd do ", serialized", linking to https://url.spec.whatwg.org/#concept-url-serializer (so I think that's `[=URL serializer|serialized=]`).

jungkees:
@sideshowbarker, could you address https://github.com/w3c/ServiceWorker/pull/1110#discussion_r122037640 to merge?

sideshowbarker:
> @sideshowbarker, could you address #1110 (comment) to merge?

Done in https://github.com/w3c/ServiceWorker/pull/1110/commits/555acd398e3fd3669a772497d0d6775b4a9a6a0d

Sorry for having taken so long to get back to this

jungkees:
Merged. Thanks a lot @sideshowbarker!

PR #1113, opened 2017-04-11
Use [SameObject] for the navigationPreload attribute

foolip:
The spec says "A service worker registration has an associated
NavigationPreloadManager object" and this object never changes. This is
also true of Blink's implementation.


jungkees:
Thanks!

PR #1120, opened 2017-04-17
Typo fix

aliams:
'Let navigateFailed to false' to 'Let navigateFailed be false'


aliams:
cc @jakearchibald and @jungkees

PR #1136, opened 2017-05-05
Editorial: Simplify `ready`

jugglinmike:
The previous structure for the `ready` algorithm involved a recursive
path through an "else" branch. This tended to obscure the fact that the
corresponding branch was *not* recursive and that the following steps
were indeed reachable.

Re-factor the algorithm to include fewer total branches and to contain
the recursion more tightly.

---

I recognize that this is pretty subjective, so I'll be satisfied with the
opinion of the editors, whatever that is.


ylafon:

jungkees:
Thanks @jugglinmike! It reads easier with the change.

(This change doesn't affect any behavior.)

PR #1138, opened 2017-05-12
Assert same-origin for registration matching

### LINKED ISSUE #1118: Match Service Worker Registration should assert same-origin?
It's non-obvious from reading https://w3c.github.io/ServiceWorker/#scope-match-algorithm that `clientURL` can't match a cross-origin service worker registration. AFAICT it relies on URLs serializing with a slash at the end. Can Step 5 assert the `matchingScope` is same-origin with `clientURL`?

jungkees:
This patch adds an assertion in the Match Service Worker Registration
that the given URL and the matched URL are the same origin. This also
adds a note that the URL serializer serializes the URLs with a trailing
slash at the end of the origin part of the URLs.

Fixes #1118.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3345,12 +3345,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |scopeStringSet| be the result of [=map/get the keys|getting the keys=] from <a>scope to registration map</a>.
       1. Set |matchingScopeString| to the longest value in |scopeStringSet| which the value of |clientURLString| starts with, if it exists.
 
-          Note: The URL string matching in this step is prefix-based rather than path-structural. E.g. a client URL string with "https://example.com/prefix-of/resource.html" will match a registration for a scope with "https://example.com/prefix". The URL string comparison is safe for the same-origin security as the URLs are serialized with a trailing slash at the end of the origin part of the URLs.
+          Note: The URL string matching in this step is prefix-based rather than path-structural. E.g. a client URL string with "https://example.com/prefix-of/resource.html" will match a registration for a scope with "https://example.com/prefix". The URL string comparison is safe for the same-origin security as HTTP(S) URLs are always [=URL serializer|serialized=].
 
       1. Let |matchingScope| be null.
       1. If |matchingScopeString| is not the empty string, then:
           1. Set |matchingScope| to the result of <a lt="URL parser">parsing</a> |matchingScopeString|.
-          1. Assert: |matchingScope|'s [=url/origin=] and |clientURL|'s [=url/origin=] are the [=same origin=].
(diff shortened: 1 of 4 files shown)


jungkees:
@annevk, could you review?

/cc @estark37 @noncombatant

docs/index.bs
```diff
@@ -3345,10 +3345,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |scopeStringSet| be the result of [=map/get the keys|getting the keys=] from <a>scope to registration map</a>.
       1. Set |matchingScopeString| to the longest value in |scopeStringSet| which the value of |clientURLString| starts with, if it exists.
 
-          Note: The URL string matching in this step is prefix-based rather than path-structural (e.g. a client URL string with "/prefix-of/resource.html" will match a registration for a scope with "/prefix").
+          Note: The URL string matching in this step is prefix-based rather than path-structural. E.g. a client URL string with "https://example.com/prefix-of/resource.html" will match a registration for a scope with "https://example.com/prefix". The URL string comparison is safe for the same-origin security as the URLs are serialized with a trailing slash at the end of the origin part of the URLs.
```
annevk: as HTTP(S) URLs are always serialized*

docs/index.bs
```diff
@@ -3345,10 +3345,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
 
       1. Let |matchingScope| be null.
-      1. If |matchingScopeString| is not the empty string, set |matchingScope| to the result of <a lt="URL parser">parsing</a> |matchingScopeString|.
+      1. If |matchingScopeString| is not the empty string, then:
+          1. Set |matchingScope| to the result of <a lt="URL parser">parsing</a> |matchingScopeString|.
+          1. Assert: |matchingScope|'s [=url/origin=] and |clientURL|'s [=url/origin=] are the [=same origin=].
```
annevk: s/the//

jungkees: Done.

jungkees:
@annevk, thanks for reviewing!

annevk:
So I didn't mean that you should remove the end of the sentence... The bit about slashes and origins was still useful.

annevk:
I don't think the sentence makes much sense as-is.

jungkees:
Oh.. so you meant something like the following?

> The URL string comparison is safe for the same-origin security as HTTP(S) URLs are always serialized with a trailing slash at the end of the origin part of the URLs.

annevk:
Yeah, that looks better.

annevk:
(For next time, ask for review again. Happy to do it.)

jungkees:
Okay. Will do. Thanks!

jungkees:
Addressed: https://github.com/w3c/ServiceWorker/commit/46b5d67b1cff072eae82383729448c757a6a83ab.

PR #1143, opened 2017-05-17
Remove unused flag

jugglinmike:
This flag is not listed in the definition of the ServiceWorkerGlobalScope's available flags. Because it is not referenced by any other specification text, the value has no effect on the observable semantics.

Remove the algorithm step that references the otherwise-unused "closing" flag.


jugglinmike:
Found it: https://w3c.github.io/workers/#event-loop . Sorry for the noise!

ylafon:

PR #1144, opened 2017-05-17
Extend attribute ref with link to external spec

jugglinmike:
Because the "closing" attribute is not referenced elsewhere within the
specification, this step may cause confusion for readers who are not
familiar with its origins in the Workers specification. Extend the
reference to include a hyperlink to that document.

(The "confusion" mentioned above was my own. See gh-1143.)

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2891,7 +2891,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       1. If |serviceWorker| is not running, abort these steps.
       1. Let |serviceWorkerGlobalScope| be |serviceWorker|'s [=service worker/global object=].
-      1. Set |serviceWorkerGlobalScope|'s closing flag to true.
+      1. Set |serviceWorkerGlobalScope|'s {{closing!!attribute}} flag to true.
       1. [=set/Remove=] all the [=items=] from |serviceWorker|'s [=set of extended events=].
       1. If there are any <a>tasks</a>, whose <a>task source</a> is either the <a>handle fetch task source</a> or the <a>handle functional event task source</a>, queued in |serviceWorkerGlobalScope|'s <a>event loop</a>'s [=/task queues=], <a lt="queue a task">queue</a> them to |serviceWorker|'s <a>containing service worker registration</a>'s corresponding [=service worker registration/task queues=] in the same order using their original <a>task sources</a>, and discard all the <a>tasks</a> (including <a>tasks</a> whose <a>task source</a> is neither the <a>handle fetch task source</a> nor the <a>handle functional event task source</a>) from |serviceWorkerGlobalScope|'s <a>event loop</a>'s [=/task queues=] without processing them.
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -2891,7 +2891,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       1. If |serviceWorker| is not running, abort these steps.
       1. Let |serviceWorkerGlobalScope| be |serviceWorker|'s [=service worker/global object=].
-      1. Set |serviceWorkerGlobalScope|'s closing flag to true.
+      1. Set |serviceWorkerGlobalScope|'s {{closing!!attribute}} flag to true.
```
domenic: It's not an attribute (i.e. there is no `self.closing`); it's a flag. The reason this works is because of some faulty Bikeshed heuristics.

To make this work properly you need to edit the HTML spec to add `data-export="" data-dfn-for="WorkerGlobalScope"` to the `<dfn>`. Then you can do `[=WorkerGlobalScope/closing=]` in this document.

Alternately you can use `<pre class="anchors">` to manually add something that isn't exported, but the best fix is to export it.

ylafon:

docs/index.bs
jugglinmike: Got it. Thanks!

jugglinmike: (I'll try to follow up on this tomorrow.)

jugglinmike:
@domenic It's been a few days since https://github.com/whatwg/html/pull/2703
landed, so I tried running `bikeshed update` and updating this patch as you
requested. Unfortunately, that failed with the following error:

    LINK ERROR: No 'dfn' refs found for 'closing'.
    <a data-link-type="dfn" data-lt="closing" data-link-for="WorkerGlobalScope">closing</a>

I've pushed up a commit to show the input; can you see what's going wrong here?

jugglinmike:
@tabatkins would you mind helping me figure out what's going wrong here?

domenic:
Yeah, unfortunately everything seems like it should be working, so I have no ideas myself...

tabatkins:
Because this definition doesn't have an explicit `data-dfn-type`, Shepherd kicks into heuristic mode, which detects the `id="dom-Interface-something"` pattern and infers that it's an attribute of the WorkerGlobalScope.  Thus its type ends up being `attribute`, rather than `dfn`, so that autolinking syntax won't pick it up.

(You can see it by running `bikeshed refs --text=closing --for=WorkerGlobalScope` - the reference shows up properly, but with a `type: attribute` line.)

jugglinmike:
Thanks, @tabatkins! I've opened https://github.com/whatwg/html/pull/2721 to fix the definition in the HTML specification.

jugglinmike:
@domenic https://github.com/whatwg/html/pull/2721 has landed and made its way to Bikeshed's database. This patch ought to be ready to land, now.

jungkees:
Thanks!

PR #1145, opened 2017-05-18
Editorial: Correct grammar of informative note

jugglinmike:
I'm not sure if the language "x has been the target of a y event" is
conventional, but I haven't been able to find similar text in any other
specification. Happy to re-phrase as the editors please.


ylafon:

docs/index.bs
```diff
@@ -2486,7 +2486,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. If |job|'s <a>job type</a> is *register*, run <a>Register</a> with |job| <a>in parallel</a>.
           1. Else if |job|'s <a>job type</a> is *update*, run <a>Update</a> with |job| <a>in parallel</a>.
 
-              Note: For a register job and an update job, the user agent delays queuing a task for running the job until after the document initiated the job has been dispatched {{Document/DOMContentLoaded}} event.
+              Note: For a register job and an update job, the user agent delays queuing a task for running the job until after the document that initiated the job has been the target of a {{Document/DOMContentLoaded}} event.
```
jungkees: > x has been the target of a y event

I don't think this expression's clearly saying the event has been *dispatched*. How about something like "until after a DOMContentLoaded event has been dispatched to the document that initiated the job."?

jugglinmike: Works for me!

jungkees:
LGTM. Thanks!

PR #1147, opened 2017-05-19
Fix fetchEvent.request initialization

### LINKED ISSUE #951: Handle Fetch creates JavaScript objects too early (and other feedback)
1. JavaScript objects are being created in step 4 (you also need to explicitly create a `Headers` object) without a global object. You should first find the service worker, start it if it hasn't started, then queue a task, and from that task create these objects.
2. Don't grab the headers attribute directly. Just use the associated concept.
3. In step 9 "corresponds" is not defined. E.g., what if request's client is null?
4. Also, if service worker client is just another word for environment settings object, why even mention it in step 9? Why not just set client to request's client?

jungkees:
This patch moves the creation of a Request object and its associated
Headers object to a right place. It moves those steps from outside into
the task where the fetch event's request attribute is initialized.

Fixes #951.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2913,9 +2913,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |handleFetchFailed| be false.
       1. Let |respondWithEntered| be false.
       1. Let |eventCanceled| be false.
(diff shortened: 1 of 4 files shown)


jungkees:
@annevk, this work intends to fix #951. 1 and 2 in the issue would be resolved by this patch. 3 and 4 seem to have gone by previous works already I believe. Please take a look.

PR #1150, opened 2017-05-25
Editorial: Correct value references

jugglinmike:
Set job properties based on the value of the like-named algorithm inputs
(not the job properties themselves).

---

Even after application of this patch, the language is still somewhat confusing
because the job properties and algorithm inputs continue to share the same
name. I think a better fix would be to re-name the algorithm inputs (maybe
simply by removing the word "flag"), but I'd like to hear from an editor about
that.


ylafon:

PR #1152, opened 2017-05-26
Editorial: correct method control flow

jugglinmike:
The previous control flow for the `ServiceWorkerRegistration#update`
algorithm did not return a value under certain error conditions. Update
the algorithm steps to consistently return the Promise defined to track
the status of the operation.


ylafon:

jungkees:
@jugglinmike, sorry I didn't recall my intention on the steps of this method reviewing back after a long time and reverted the patch.

The update() method isn't actually supposed to go in parallel until the update job that it creates is scheduled by Run Job algorithm later. The issue that you posed in this PR should still be addressed. I'll do that by filing an issue and moving the guarding steps to appropriate places - maybe to the Update algorithm. (Need to check.)

Issue #1134, opened 2017-04-29
Register headers with IANA


dret:
the HTTP headers specified in https://www.w3.org/TR/service-workers-1/#extended-http-headers should be registered with IANA, so that they get listed in the IANA HTTP header field registry. if that would help, i can write-up a PR for that.

jungkees:
Thanks for the information and suggestion! I'm not familiar with the process there. It'd be of great if you could help.

/cc @mnot @annevk @slightlyoff @jakearchibald @mkruisselbrink

mnot:
Happy to help, @dret are you going to do a PR?

dret:
On 2017-05-18 03:51, Mark Nottingham wrote:
> Happy to help, @dret <https://github.com/dret> are you going to do a PR?

i'll do that, please stay tuned!

dret:
https://github.com/w3c/ServiceWorker/pull/1154 is a first attempt, please have a look.

yoshisatoyanagisawa:
Triage: https://github.com/w3c/ServiceWorker/pull/1154 is still in progress. keep it open.

PR #1154, opened 2017-05-29
adding IANA considerations section

dret:
starting work on #1134

### DIFF
--- a/docs/v1/index.bs
+++ b/docs/v1/index.bs
@@ -2050,6 +2050,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 </section>
 
 <section>
+  <h2 id="iana-considerations">IANA Considerations</h2>
+  
+  <p>The following registrations are for community review and will be submitted to the <a href="https://www.ietf.org/iesg/">IESG</a> for review, approval, and registration with <a href="https://www.iana.org/">IANA</a>.</p>
+  
+</section>
+
+<section>
   <h2 id="algorithms">Appendix A: Algorithms</h2>
 
   The following definitions are the user agent's internal data structures used throughout the specification.


ylafon:

dret:
first attempt at resolving #1134.

docs/v1/index.bs
```diff
@@ -2050,6 +2050,36 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 </section>
 
 <section>
+  <h2 id="iana-considerations">IANA Considerations</h2>
+  
+  <p>The following registrations are for community review and will be submitted to the <a href="https://www.ietf.org/iesg/">IESG</a> for review, approval, and registration with <a href="https://www.iana.org/">IANA</a>.</p>
```
jungkees: Will this part of the section need to be changed or replaced when the fields are registered to IANA?

docs/v1/index.bs
```diff
@@ -2050,6 +2050,36 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+  
+  <section>
+    <h3 id="iana-registration-service-worker">HTTP Header Field Service-Worker</h3>
+    <ul>
+        <li>Header Field Name: Service-Worker</li>
+        <li>Applicable Protocol: Hypertext Transfer Protocol (HTTP)</li>
+        <li>Status: Standard</li>
+        <li>Author/Change controller: W3C</li>
+        <li>Specification document(s): https://www.w3.org/TR/service-workers-1/</li>
```
jungkees: Not sure which document it should point to. Service Workers 1 is a snapshot of Service Workers which is a living document.

jungkees:
Thanks @dret!

Could you please work on  docs/index.bs instead of  docs/v1/index.bs? Also, it'd be great you could use markdown format as we do for new content. If you prefer HTML, that'd be still okay.

@mnot, @ylafon, your cross review would be appreciated.

dret:
On 2017-05-29 12:51, Jungkee Song wrote:
> *@jungkees* commented on this pull request.
> Will this part of the section need to be changed or replaced when the 
> fields are registered to IANA?

not really. the language is always a bit odd because it is both a 
request for registration and announces the fact that a registration has 
been made (after the spec is stable and published).

dret:
On 2017-05-29 12:54, Jungkee Song wrote:
> *@jungkees* commented on this pull request.
> +        <li>Header Field Name: Service-Worker</li>
> +        <li>Applicable Protocol: Hypertext Transfer Protocol (HTTP)</li>
> +        <li>Status: Standard</li>
> +        <li>Author/Change controller: W3C</li>
> +        <li>Specification document(s): https://www.w3.org/TR/service-workers-1/</li>
> Not sure which document it should point to. Service Workers 1 is a 
> snapshot of Service Workers which is a living document.

from the IETF point of view a stable document would be preferable. i 
guess in principle what's really required is a stable definition, and 
stable documents simply are a good proxy for that. is there any 
precedence in documents that are both W3C and WHATWG? maybe at least 
shoot for consistency...

dret:
On 2017-05-29 12:59, Jungkee Song wrote:
> Thanks @dret <https://github.com/dret>!
> Could you please work on docs/index.bs instead of docs/v1/index.bs? 
> Also, it'd be great you could use markdown format as we do for new 
> content. If you prefer HTML, that'd be still okay.

sorry! will do, i tried to find the right document to work on and 
apparently made the wrong choice.

jungkees:
> not really. the language is always a bit odd because it is both a request for registration and announces the fact that a registration has been made (after the spec is stable and published).

Okay.

> is there any precedence in documents that are both W3C and WHATWG?

No. There are only W3C documents.

> sorry! will do, i tried to find the right document to work on and apparently made the wrong choice.

Sorry if the guide was not clear. Thanks for your understanding.

domenic:
Service Workers Nightly should be referenced; the registration/header-defining portions are stable, and this ensures that any bugfixes are present.

mnot:
To be clear -- the document's contents don't need to be "stable" (i.e., unchanging); only the means of referencing it (the URL). And worst case, that can be updated.

jungkees:
Okay. In that case, https://w3c.github.io/ServiceWorker/ (Nightly) would make sense.

PR #1151, opened 2017-05-26
Use ASCII serialization for origins

jungkees:
This patch makes it to use ASCII serialization for
(Extendable)MessageEvent.origin instead of Unicode serialization which
has been removed from HTML.

HTML issue: https://github.com/whatwg/html/issues/2568.
HTML change: https://github.com/whatwg/html/pull/2689/commits/c3b2511ee0158fa4b5bd4e4689c6be54132f3d90.

Fixes #1142.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1151.html" title="Last updated on Jun 13, 2019, 5:57 PM UTC (7c0a94c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1151/9a74fc0...7c0a94c.html" title="Last updated on Jun 13, 2019, 5:57 PM UTC (7c0a94c)">Diff</a>

PR #1160, opened 2017-06-11
Use “serialization of an origin” w/o “Unicode”

sideshowbarker:
See https://github.com/whatwg/html/commit/59ebd9c094d9d532458a9ee61f307bf41bc70811

In place of “Unicode serialization of an origin” we should now just use
plain “serialization of an origin”.


sideshowbarker:
Duplicate of https://github.com/w3c/ServiceWorker/pull/1151

PR #1166, opened 2017-07-11
Disallow waitUntil() of manually constructed events

### LINKED ISSUE #1040: clarify that manually constructed ExtendableEvent objects may not extend lifetime via waitUntil()
Currently its possible to do this in the spec:

```
let e = new ExtendableEvent('foo');
e.waitUntil(new Promise());
```

By default the extensions allowed flag is set, so this does not throw an InvalidStateError.

We should be clear, though, that this does not actually extend the lifetime of the SW.  Currently the spec is written in a way that suggests that this might imbue some keep-alive budget to the SW.

jungkees:
This change adds a guard to waitUntil() that disallows the extension of
the event that is not dispatched by the user agent.

Fixes #1040.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1333,9 +1333,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       <dfn method for="ExtendableEvent"><code>waitUntil(|f|)</code></dfn> method *must* run these steps:
 
-        1. If the [=ExtendableEvent/pending promises count=] is zero and the [=dispatch flag=] is unset, then:
-            1. <a>Throw</a> an "{{InvalidStateError}}" {{DOMException}}.
-            1. Abort these steps.
+        1. If the {{Event/isTrusted}} attribute is false, [=throw=] an "{{InvalidStateError}}" {{DOMException}} and abort these steps.
+
+        1. If the [=ExtendableEvent/pending promises count=] is zero and the [=dispatch flag=] is unset, [=throw=] an "{{InvalidStateError}}" {{DOMException}} and abort these steps.
 
             Note: If no lifetime extension promise has been added in the task that called the event handlers, calling {{ExtendableEvent/waitUntil()}} in subsequent asynchronous tasks will throw.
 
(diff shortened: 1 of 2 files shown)


jungkees:
@jakearchibald, I added a guard to prevent the script created event from extending the lifetime via waitUntil(). There's a test already: https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/resources/extendable-event-async-waituntil.js#L45. But I tried to update it to make the isTrusted test independant from other conditions. Please take a look.

Chrome implemented this by checking whether the event is implementing the internal observer structure instead of directly using isTrusted state. (@mattto, I can change this to using isTrusted if you think that makes sense.)

/cc @wanderview, @mattto

mattto:
I'm not sure this spec change is needed, because it already disallowed waitUntil() because the dispatch flag would not be set. At least that's why Chromium landed https://codereview.chromium.org/2877543003

Is there a material behavior change here?

jungkees:
waitUntil() can run outside of a dispatch task, but the dispatch flag is set when an event is dispatched anyway. Devs can self.dispatchEvent(syntheticEvent), and event.waitUntil() in the event handler won't throw if we don't check isTrusted there. So, the spec behavior's changing with this PR. I think Chrome already handles this in a similar way though it isn't directly using isTrusted: https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/serviceworkers/ExtendableEvent.cpp?q=extendableevent&dr=CSs&l=55

mattto:
Ah, OK didn't realize that.

jakearchibald:
Sorry for the delay on this, it's still on my TODO list

jakearchibald:
LGTM, merge it!

Sorry for not getting to it sooner – I'd got it into my head that this was a bigger change that would have taken more time 😄

jungkees:
It seems the merge automatically closed the issue in the OP even though I changed the commit message to not use "Fix": https://github.com/w3c/ServiceWorker/commit/c2b59bfd3fd268e8fff95c278d8a9177a51852cc. The test is underway: https://github.com/w3c/web-platform-tests/pull/6517.

PR #1178, opened 2017-08-07
Handing fetch termination

jakearchibald:
Is this a valid way of handling this, in terms of https://github.com/whatwg/fetch/pull/523.

It seems too easy.

(I'll add the same to foreign fetch if this is 👍 )

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/fetch-abort.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/d0df742...b035492.html)

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2996,7 +2996,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Let |requestObject| be a new {{Request}} object associated with |request| and a new associated {{Headers}} object whose [=guard=] is "`immutable`".
           1. Initialize |e|’s {{Event/type}} attribute to {{fetch!!event}}.
           1. Initialize |e|’s {{Event/cancelable}} attribute to true.
-          1. Initialize |e|’s {{FetchEvent/request}} attribute to |requestObject|
+          1. Initialize |e|’s {{FetchEvent/request}} attribute to |requestObject|.
           1. Initialize |e|’s {{FetchEvent/preloadResponse}} to |preloadResponse|.
           1. Initialize |e|'s {{FetchEvent/clientId}} attribute to |client|'s [=environment/id=].
           1. If |request| is a <a>non-subresource request</a> and |request|'s [=request/destination=] is not {{RequestDestination/"report"}}, initialize |e|'s {{FetchEvent/reservedClientId}} attribute to |reservedClient|'s [=environment/id=], and to the empty string otherwise.
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2940,6 +2940,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |client| be |request|'s [=request/client=].
       1. Let |reservedClient| be |request|'s [=request/reserved client=].
       1. Let |preloadResponse| be a new [=promise=].
+      1. Let |fetchInstance| be the instance of the [=/fetch=] algorithm representing the ongoing fetch.
       1. Assert: |request|'s [=request/destination=] is not "<code>serviceworker</code>".
       1. If |request| is a <a>potential-navigation-or-subresource request</a>, then:
           1. Return null.
@@ -2965,13 +2966,14 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
               1. [=header list/Append=] to |preloadRequestHeaders| a new [=header=] whose [=header/name=] is \`<code>Service-Worker-Navigation-Preload</code>\` and [=header/value=] is |registration|'s [=navigation preload header value=].
               1. Set |preloadRequest|'s [=service-workers mode=] to "`none`".
               1. Run the following substeps [=in parallel=]:
-                  1. [=Fetch=] |preloadRequest|.
+                  1. [=Fetch=] |preloadRequest| and let |preloadFetchInstance| be the instance of the [=/fetch=] algorithm.
 
                       To [=process response=] for |navigationPreloadResponse|, run these substeps:
 
                       1. If |navigationPreloadResponse|'s [=response/type=] is "`error`", reject |preloadResponse| with a `TypeError` and terminate these substeps.
                       1. Associate |preloadResponseObject| with |navigationPreloadResponse|.
                       1. Resolve |preloadResponse| with |navigationPreloadResponse|.
+                  1. If |fetchInstance| is [=fetch/terminated=], then [=fetch/terminate=] |preloadFetchInstance| with the aborted flag set.
           1. Else, resolve |preloadResponse| with undefined.
 
           Note: From this point, the [=/service worker client=] starts to <a>use</a> its <a>active service worker</a>'s <a>containing service worker registration</a>.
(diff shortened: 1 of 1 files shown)


docs/index.bs
```diff
@@ -2996,6 +2997,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
               1. If |e|'s [=FetchEvent/respond-with error flag=] is set, set |handleFetchFailed| to true.
               1. Else, set |response| to |e|'s [=FetchEvent/potential response=].
           1. If |e|'s <a>canceled flag</a> is set, set |eventCanceled| to true.
+          1. Run the following steps [=in parallel=]:
+              1. Wait for the fetch to become [=fetch/terminated=].
+              1. [=Queue a task=] to [=AbortSignal/signal abort=] on |requestObject|'s [=Request/signal=].
```
annevk: It seems this algorithm has a chance to never finish.

jakearchibald: Yeah, I figured "GC" would take care of it.

What's the alternative? "Wait for the fetch to become [=fetch/terminated=], or (something that signal successful completion)"

annevk: I tried to think about this a bit and the way XMLHttpRequest seems to approach is that you always get a response from the fetch process, but it might be a network error annotated with a termination reason. And for the request/response bodies we got the stream that errors.

jakearchibald: I could wait until on of the following:

* Fetch is terminated.
* "Process response" for fetch occurs, and the response type is "error".
* "Process response done" for fetch occurs.

Then, only signal abort if fetch is terminated.

annevk: What I was saying is that currently "Fetch is terminated." is not really a signal we expose to the outside (other than the response having a termination reason or the stream erroring).

jakearchibald:
@annevk re:

> "Fetch is terminated." is not really a signal we expose to the outside (other than the response having a termination reason or the stream erroring).

I agree, but this feels like a special case since fetch called me, I didn't call fetch.

In the fetch spec I've tried to keep termination handling as close to the leaf end as I can. [HTTP fetch](https://fetch.spec.whatwg.org/#concept-http-fetch) doesn't have a "If terminates" break clause around the call to service workers, so it feels "handle fetch" should be handling the termination.

Alternatively, I could add a break clause the call to "handle fetch" in "main fetch". Then, in "handle fetch" I can somehow "handle response" for the outer fetch, and see if its abort flag is set. But this will be really messy, as the fetch spec is waiting for a response from the service worker spec, but the service worker spec is also waiting for a response from the fetch spec. I'd also end up looping the service worker response back into the service worker spec, and I'd need to ignore that somehow.

I've updated the PR a little, but it seems to express the behaviour pretty clearly. It's a bit hand-wavey around getting an instance of the fetch, but I think that's part of https://github.com/whatwg/fetch/issues/536#issuecomment-330482027.

docs/index.bs
```diff
@@ -2991,9 +2993,10 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Invoke <a>Run Service Worker</a> algorithm with |activeWorker| as the argument.
       1. <a>Queue a task</a> |task| to run the following substeps:
           1. Let |e| be the result of <a>creating an event</a> with {{FetchEvent}}.
+          1. Let |requestObject| be a new {{Request}} object associated with |request| and a new associated {{Headers}} object whose [=guard=] is "`immutable`".
           1. Initialize |e|’s {{Event/type}} attribute to {{fetch!!event}}.
           1. Initialize |e|’s {{Event/cancelable}} attribute to true.
-          1. Initialize |e|’s {{FetchEvent/request}} attribute to a new {{Request}} object associated with |request| and a new associated {{Headers}} object whose [=guard=] is "`immutable`".
+          1. Initialize |e|’s {{FetchEvent/request}} attribute to |requestObject|
```
annevk: Needs a trailing dot.

annevk:
Yeah, I guess we need to fix that issue at some point. Not looking forward to updating all the callers again.

jakearchibald:
@annevk Are you ok with:

> Let |fetchInstance| be the instance of the [=/fetch=] algorithm representing the ongoing fetch.

And

> If |fetchInstance| is [=fetch/terminated=], then…

Here?

If so I'll write some tests and get this merged.

annevk:
Yeah, as you said there's not really any better way to do it at the moment. Maybe note these cases in the Fetch issue so we make sure they get addressed by whatever we come up with?

jakearchibald:
Tests at https://github.com/w3c/web-platform-tests/pull/7674

jungkees:
@jakearchibald, it seems this is ready to be merged?

jakearchibald:
Done. I need to take another pass at the tests. Just minor stuff though.

ItalyPaleAle:
Hi all, sorry for resurrecting this old thread. Wondering if this behavior that I'm noticing is the one as intended according to the spec, as I'm noticing the same thing on all major browsers (Firefox, Chromium, Safari).

It seems that this doesn't catch situations where the browser itself is terminating a request.

Example:

### index.html

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Test fetch</title>
</head>
<body>

<button onclick="getVideo()">Get video</button>
<button onclick="stop()">Stop</button>
<div id="out"></div>

<script>
;(async function() {
    await navigator.serviceWorker.register('sw.js')
    console.info('Service worker registered')
    await navigator.serviceWorker.ready
    console.log('Service worker installed')
    document.getElementById('out').innerText = 'Ready'
})()

function getVideo() {
    document.getElementById('out').innerHTML = '<video autoplay controls  muted="muted" src="video.mp4" style="width: 400px; height: auto;"></video>'
}

function stop() {
    document.getElementById('out').innerHTML = ''
}
</script>

</body>
</html>
```

### sw.js

```js
self.addEventListener('install', () => {
    self.skipWaiting()
})

self.addEventListener('fetch', (event) => {
    console.log('fetch', event)
    event.request.signal.addEventListener('abort', (event) => {
        console.log('aborted', event)
    })
})

self.addEventListener('activate', () => self.clients.claim())
```

### Description

Place a video file in the same folder called `video.mp4`, ideally a large-r file (say, 100MB+). Start a server (e.g. `docker run --rm --name some-nginx -v $PWD:/usr/share/nginx/html:ro -p 8080:80 nginx`) and navigate to the page (http://localhost:8080).

Start playing the video (hint: it helps to throttle the network), then press the "stop" button before the video is done loading. Alternatively, skip forward in the video to a part that hasn't been buffered yet.

In the dev tools, you'll see the browser interrupted the old request (and made a new one if necessary).

However, the `abort` event is never triggered. And, in the console, even after the request is done, the `signal` objected nested into `event` still reports the request as not aborted.

Is this how it's supposed to behave according to specs? Or is it a bug in the browsers?

Why am I doing this? I have a SW that intercepts a request and makes a new one instead. I need to be able to understand if the browser has canceled the "original" request to cancel the new one too.

wanderview:
I recently raised the same issue on https://github.com/w3c/ServiceWorker/issues/1544.

I think the intent was to abort the signal here, but no browser implements it yet.  Also, as I mention in the issue above there may be a spec problem that also makes it not do what is intended.

ItalyPaleAle:
Thanks, @wanderview ! I'll follow the other thread then. If you need more details on what I'm trying to do, to add to your case on #1544, please let me know.

PR #1190, opened 2017-08-22
Remove incumbent/fetching record from Cache behavior

### LINKED ISSUE #884: addAll should resolve when the cache is fully & successfully written
At the moment it resolves on headers, which doesn't feel like a good fit for install `waitUntil`.

We changed `.put`, and I think it's an oversight that we didn't fix it for `addAll`. What do you think @wanderview?

jungkees:
This change removes the incumbent/fetching record concept that allowed
committing a fetching resource to cache and match/matchAll to return a
fully written existing resource, if any, with priority over a fetching
resource. After this change, add/addAll/put promises resolve when the
resources ared fully fetched and committed to cache. match/matchAll will
not return any on-going fetching resources.

This changes the specification type of underlying cache objects from a
map (request to response map) to a list (request response list). The
change is to make the arguments and the return values of the Cache
methods and algorithms (Batch Cache Operations and Query Cache) conform
to one another. The list type seems fine as the algorithms tend to
iterate through the list to find certain items with search options.
Looking at the details of the acutal implementations, I plan to update
it further if needed.

Fixes #884.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/remove-incumbent-fetching-concept.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/0aefcc4...fbf6f44.html)


jakearchibald:
Whoop! We now have PR previews!

docs/index.bs
```diff
@@ -1965,17 +1961,14 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. Run these substeps <a>in parallel</a>:
             1. Let |responseArray| be an empty array.
             1. If the optional argument |request| is omitted, then:
-                1. For each <a>fetching record</a> |entry| of its <a>request to response map</a>, in key insertion order:
-                    1. Add a  copy of |entry|.\[[value]] to |responseArray|.
+                1. [=list/For each=] |item| of the [=context object=]'s [=request response list=]:
+                    1. Add a copy of |item|'s |response| to |responseArray|.
```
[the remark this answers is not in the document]
jakearchibald: This could be part of an additional PR though, as this PR doesn't introduce this problem.

docs/index.bs
```diff
@@ -3481,93 +3452,79 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
                       1. If the cache write operation in the previous two steps failed due to exceeding the granted quota limit, <a>throw</a> a "{{QuotaExceededError}}" {{DOMException}}.
-                      1. Add an array [|operation|.request, |operation|.response] to |addedRecords|.
-                  1. Add |operation|.response to |resultArray|.
-              1. Return |resultArray|.
+                      1. [=list/Append=] |operation|.{{CacheBatchOperation/request}}/|operation|.{{CacheBatchOperation/response}} to |addedItems|.
+                  1. [=Append=] |operation|.{{CacheBatchOperation/request}}/|operation|.{{CacheBatchOperation/response}} to |resultList|.
+              1. Return |resultList|.
           1. And then, if an exception was <a lt="throw">thrown</a>, then:
-              1. Set the <a>context object</a>'s <a>request to response map</a> to |itemsCopy|.
+              1. Set |cache| to |itemsCopy|.
```
jakearchibald: Isn't this just setting a local variable? As in, it no longer reverts the operation.

Again, this isn't a new problem, but isn't there a bit of a race condition here? Since we're replacing the whole cache with an older copy, there may be concurrent operations of this resulting in data loss.

docs/index.bs
```diff
@@ -3481,93 +3452,79 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+          1. If |matchFailed| is false, [=list/append=] |cachedRequest|/|cachedResponse| to |resultList|.
+      1. Return |resultList|.
   </section>
 
   <section algorithm>
     <h3 id="batch-cache-operations-algorithm"><dfn>Batch Cache Operations</dfn></h3>
 
       : Input
-      :: |operations|, an array of  {{CacheBatchOperation}} dictionary objects
+      :: |operations|, a [=list=] of {{CacheBatchOperation}} dictionary objects
```
jakearchibald: We never expose this, so it doesn't need to be a dictionary right?

docs/index.bs
```diff
@@ -3481,93 +3452,79 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-          1. Let |itemsCopy| be a new <a>request to response map</a> that is a copy of its <a>context object</a>'s <a>request to response map</a>.
-          1. Let |addedRecords| be an empty array.
+          1. Let |itemsCopy| be a new [=request response list=] that is a copy of |cache|.
+          1. Let |addedItems| be an empty [=list=].
           1. Try running the following substeps atomically:
-              1. Let |resultArray| be an empty array.
-              1. For each |operation| in |operations|:
+              1. Let |resultList| be an empty [=list=].
+              1. [=list/For each=] |operation| in |operations|:
                   1. If |operation|.{{CacheBatchOperation/type}} matches neither "delete" nor "put", <a>throw</a> a <code>TypeError</code>.
```
jakearchibald: We should probably make this an enum or list of possible values.

docs/index.bs
```diff
@@ -3481,93 +3452,79 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-                          1. Let |fetchingRecord| be the corresponding <a>fetching record</a> for |requestResponse|[0] and |requestResponse|[1] in <a>request to response map</a>.
-                          1. Set |fetchingRecord|.\[[key]] to |operation|.{{CacheBatchOperation/request}} and |fetchingRecord|.\[[value]] to |operation|.{{CacheBatchOperation/response}}.
-                      1. Else:
-                          1. Set a newly-created <a>fetching record</a> {\[[key]]: |operation|.{{CacheBatchOperation/request}}, \[[value]]: |operation|.{{CacheBatchOperation/response}}} to <a>request to response map</a>.
-
-                          Note: The cache commit is allowed as long as the response's headers are available.
-
+                      1. Set |requestResponseList| to the result of running [=Query Cache=] with |operation|.{{CacheBatchOperation/request}}.
+                      1. If |requestResponseList| [=list/is not empty=], [=list/replace=] the [=list/item=] of |cache| that matches |requestResponseList|[0] with |operation|.{{CacheBatchOperation/request}}/|operation|.{{CacheBatchOperation/response}}.
```
jakearchibald: What if there are multiple matches, shouldn't we be removing those? It might be easiest to remove all matches then just append the new entry.

jakearchibald: Oh this is now done in addAll, but I'm not sure if it's right.

docs/index.bs
```diff
@@ -1861,13 +1861,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   <section>
     <h3 id="cache-constructs">Constructs</h3>
 
-    A <dfn id="dfn-fetching-record">fetching record</dfn> is a <a>Record</a> {\[[key]], \[[value]]} where \[[key]] is a {{Request}} and \[[value]] is a {{Response}}.
+    A <dfn id="dfn-request-response-list">request response list</dfn> is a [=list=] of [=pairs=] consisting of |request| (a {{Request}} object) and |response| (a {{Response}} object).
```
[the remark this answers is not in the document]
wanderview: Yea, what Jake recommends matches implementations.  We explicitly don't always return the same Response js object from match() even if its stored in the same entry.

wanderview:
Since this PR touches the case where the body stream errors during a put(), it would be nice to add WPT tests to cover that.  I think it should be somewhat easy to do now that we have ReadableStream bodies.

jungkees:
@jakearchibald, @wanderview, thanks for reviewing. I couldn't make it before I leave for a vacation. I'm off until 6th of Sep. Will follow up on it when I come back.

docs/index.bs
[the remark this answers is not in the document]
jungkees: Absolutely. I changed it to store the request and response structs instead of the JS objects and to create the JS objects in matchAll() and keys() when requested.

docs/index.bs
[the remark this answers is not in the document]
jungkees: I changed it to using the internal structs instead of the JS objects, so it seems fine. But as Fetch provides the [cloning](https://fetch.spec.whatwg.org/#concept-response-clone) algorithm, I'll look at it further whether using that'd make it more precise.

docs/index.bs
jungkees: Yes. I replaced it with a struct.

docs/index.bs
jungkees: For now, I made it as possible values for the newly defined cache batch operations's type item.

docs/index.bs
jungkees: I addressed it as you suggested. I think we still can do better for the return value of Batch Cache Operations somehow. I'll look into it as a separate work.

docs/index.bs
jungkees: > Isn't this just setting a local variable?

I think |cache| becomes a reference to the request response list. It's confusing though.

> isn't there a bit of a race condition here?

I wanted making the cache write atomic (the step 3.3) would do some magic, but it's absolutely a part that I should audit and improve.

jungkees:
Sorry for coming back to this late. PTAL.

docs/index.bs
```diff
@@ -1986,78 +1983,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
       <dfn method for="Cache"><code>addAll(|requests|)</code></dfn> method *must* run these steps:
 
-        1. Let |responsePromiseArray| be an empty array.
-        1. Let |requestArray| be an empty array.
+        1. Let |r| be null.
+        1. Let |responsePromises| be an empty [=list=].
+        1. Let |requestList| be an empty [=list=].
         1. For each |request| whose type is {{Request}} in |requests|:
-            1. Let |r| be |request|'s [=Request/request=].
+            1. Set |r| to |request|'s [=Request/request=].
```
jakearchibald: "Let" is [block-scoped](https://infra.spec.whatwg.org/#variables), so I think this should have remained "let".

docs/index.bs
```diff
@@ -1986,78 +1983,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
             1. If |r|'s [=request/url=]'s [=url/scheme=] is not one of "<code>http</code>" and "<code>https</code>", or |r|'s [=request/method=] is not \`<code>GET</code>\`, return a <a>promise</a> rejected with a <code>TypeError</code>.
         1. For each |request| in |requests|:
-            1. Let |r| be the associated [=Request/request=] of the result of invoking the initial value of {{Request}} as constructor with |request| as its argument. If this <a>throws</a> an exception, return a <a>promise</a> rejected with that exception.
+            1. Set |r| to the associated [=Request/request=] of the result of invoking the initial value of {{Request}} as constructor with |request| as its argument. If this <a>throws</a> an exception, return [=a promise rejected with=] that exception.
```
jakearchibald: As above, should have remained "let"

docs/index.bs
```diff
@@ -1986,78 +1983,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
     </section>
 
     <section algorithm="cache-addAll">
       <h4 id="cache-addAll">{{Cache/addAll(requests)}}</h4>
 
       <dfn method for="Cache"><code>addAll(|requests|)</code></dfn> method *must* run these steps:
 
-        1. Let |responsePromiseArray| be an empty array.
-        1. Let |requestArray| be an empty array.
+        1. Let |r| be null.
```
jakearchibald: I don't think we need this, see below.

docs/index.bs
```diff
@@ -1986,78 +1983,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+        1. Let |responsePromises| be an empty [=list=].
+        1. Let |requestList| be an empty [=list=].
         1. For each |request| whose type is {{Request}} in |requests|:
```
jakearchibald: Is this so we reject without consuming the body of the request? Might be worth adding a note about that.

docs/index.bs
```diff
@@ -1986,78 +1983,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
             1. If |r|'s [=request/url=]'s [=url/scheme=] is not one of "<code>http</code>" and "<code>https</code>", or |r|'s [=request/method=] is not \`<code>GET</code>\`, return a <a>promise</a> rejected with a <code>TypeError</code>.
         1. For each |request| in |requests|:
-            1. Let |r| be the associated [=Request/request=] of the result of invoking the initial value of {{Request}} as constructor with |request| as its argument. If this <a>throws</a> an exception, return a <a>promise</a> rejected with that exception.
+            1. Set |r| to the associated [=Request/request=] of the result of invoking the initial value of {{Request}} as constructor with |request| as its argument. If this <a>throws</a> an exception, return [=a promise rejected with=] that exception.
             1. If |r|'s [=request/url=]'s [=url/scheme=] is not one of "<code>http</code>" and "<code>https</code>", then:
-                1. [=fetch/Terminate=] all the ongoing <a>fetches</a> initiated by |requests| with reason *fatal*.
+                1. [=fetch/Terminate=] all the ongoing [=fetches=] initiated by |requests| with reason *fatal*.
```
jakearchibald: We've since dropped the "reason". There's an aborted flag which should probably be set in this case. https://fetch.spec.whatwg.org/#concept-fetch-terminate

docs/index.bs
```diff
@@ -1986,78 +1983,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
                 1. Break the loop.
```
jakearchibald: Should we just return at this point?

docs/index.bs
```diff
@@ -1986,78 +1983,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-                            1. For each |invalidRecord| in |invalidRecords|:
-                                1. If |invalidRecord| is not |fetchingRecord|, delete it from <a>request to response map</a>.
-                            1. Resolve |responseBodyPromise| with |response|.
-                    1. Add |responseBodyPromise| to |responseBodyPromiseArray|.
-                1. Let |q| be <a>waiting for all</a> of |responseBodyPromiseArray|.
-                1. Return the result of <a>transforming</a> |q|  with a fulfillment handler that returns undefined.
+                        1. [=list/For each=] |fieldValue| of |fieldValues|:
+                            1. If |fieldValue| matches "`*`", then:
+                                1. Set |matchAsterisk| to true.
+                                1. [=Break=].
```
jakearchibald: Could we just reject and abort steps here? Means we don't need *matchAsterisk*.

docs/index.bs
```diff
@@ -1986,78 +1983,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+                        1. If |matchAsterisk| is true, reject |responsePromise| with a `TypeError`.
+
+                * To [=process response end-of-body=] for |response|, run these substeps:
+                    1. If |response|'s [=response/aborted flag=] is set, reject |responsePromise| with a `TypeError` and abort these steps.
```
jakearchibald: Feels like this should be an AbortError.

docs/index.bs
```diff
@@ -1986,78 +1983,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+                    1. Resolve |responsePromise| with |response|.
+
+                    Note: The cache commit is allowed when the response's body is fully received.
+
+                * To [=process response done=] for |response|, do nothing.
```
jakearchibald: I don't think we need this line.

docs/index.bs
```diff
@@ -2134,23 +2104,29 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-        1. Run these substeps <a>in parallel</a>:
-            1. Let |resultArray| be an empty array.
+        1. Let |r| be null.
+        1. If the optional argument |request| is not omitted, then:
+            1. If |request| is a {{Request}} object, then:
+                1. Set |r| to |request|'s [=Request/request=].
+                1. If |r|'s [=request/method=] is not \`<code>GET</code>\` and |options|.ignoreMethod is false, return [=a promise resolved with=] an empty array.
+            1. Else if |request| is a string, then:
+                1. Set |r| to the associated [=Request/request=] of the result of invoking the initial value of {{Request}} as constructor with |request| as its argument. If this [=throws=] an exception, return [=a promise rejected with=] that exception.
+        1. Let |realm| be the [=current Realm Record=].
```
jakearchibald: As above, we might not need this if we use frozenarray.

docs/index.bs
```diff
@@ -2181,62 +2157,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-                    1. If |v| is not undefined, return |v|.
-                    1. Return the result of running the algorithm specified in {{Cache/match(request, options)}} method of {{Cache}} interface with |request| and |options| as the arguments (providing |cache|] as thisArgument to the <a>\[[Call]]</a> internal method of {{Cache/match(request, options)}}.)
-            1. Return |p|.
+            1. Let |promise| be [=a promise resolved with=] undefined.
+            1. [=map/For each=] <var ignore>cacheName</var> → |cache| of the [=name to cache map=]:
+                1. Set |promise| to the result of [=transforming=] itself with a fulfillment handler that, when called with argument |response|, performs the following substeps [=in parallel=]:
+                    1. If |response| is not undefined, return |response|.
+                    1. Return the result of running the algorithm specified in {{Cache/match(request, options)}} method of {{Cache}} interface with |request| and |options| as the arguments (providing |cache| as thisArgument to the `\[[Call]]` internal method of {{Cache/match(request, options)}}.)
```
jakearchibald: I don't think you need the slash in \[[Call]]

docs/index.bs
```diff
@@ -2181,62 +2157,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-        1. Return a <a>promise</a> resolved with the result of running the following substeps:
-            1. [=map/For each=] |key| → <var ignore>value</var> of <a>name to cache map</a>:
+        1. Let |promise| be [=a new promise=].
+        1. Run the following substeps [=in parallel=]:
+            1. [=map/For each=] |key| → <var ignore>value</var> of the [=name to cache map=]:
                 1. If |cacheName| matches |key|, then:
-                    1. Return true.
-            1. Return false.
+                    1. Resolve |promise| with true.
+                    1. Abort these steps.
```
jakearchibald: Could probably roll this into the line above. "Resolve promise with true and abort these steps".

docs/index.bs
```diff
@@ -2181,62 +2157,64 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
       <dfn method for="CacheStorage"><code>delete(|cacheName|)</code></dfn> method *must* run these steps:
 
-        1. Let |p| be the result of running the algorithm specified in {{CacheStorage/has(cacheName)}} method with |cacheName| as the argument.
-        1. Return the result of <a>transforming</a> |p| with a fulfillment handler that, when called with argument |cacheExists|, performs the following substeps <a>in parallel</a>:
+        1. Let |promise| be the result of running the algorithm specified in {{CacheStorage/has(cacheName)}} method with |cacheName|.
+        1. Return the result of [=transforming=] |promise| with a fulfillment handler that, when called with argument |cacheExists|, performs the following substeps [=in parallel=]:
             1. If |cacheExists| is true, then:
-                1. Delete a <a>Record</a> {\[[key]], \[[value]]} <var ignore>entry</var> from its <a>name to cache map</a> where |cacheName| matches entry.\[[key]].
+                1. [=map/Remove=] the [=name to cache map=][|cacheName|].
                 1. Return true.
```
jakearchibald: I'm not sure we can "return" from in parallel steps.

docs/index.bs
```diff
@@ -1861,15 +1862,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   <section>
     <h3 id="cache-constructs">Constructs</h3>
 
-    A <dfn id="dfn-fetching-record">fetching record</dfn> is a <a>Record</a> {\[[key]], \[[value]]} where \[[key]] is a {{Request}} and \[[value]] is a {{Response}}.
+    A <dfn id="dfn-request-response-list">request response list</dfn> is a [=list=] of [=pairs=] consisting of a request (a [=/request=]) and a response (a [=/response=]).
```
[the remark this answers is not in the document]
jungkees: I think the request and the response definitions then should belong to something like a _request response pair_ concept instead of the list itself. I don't see any other particular needs for the definition of the pair though. It seems okay as-is as the request and the response can be identified as the items of the pairs in the list.

docs/index.bs
```diff
@@ -1861,15 +1862,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
 
-    A <a>fetching record</a> has an associated <dfn id="dfn-incumbent-record">incumbent record</dfn> (a <a>fetching record</a>). It is initially set to null.
+    When a [=request response list=] is referenced from within an algorithm, an attribute getter, an attribute setter, or a method, it designates the instance that the [=context object=] represents, unless specified otherwise.
```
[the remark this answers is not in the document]
jungkees: Good idea. Addressed.

docs/index.bs
```diff
@@ -1861,15 +1862,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
 
-    A <dfn id="dfn-request-to-response-map">request to response map</dfn> is a <a>List</a> of <a>fetching records</a>.
+    A <dfn id="dfn-name-to-cache-map">name to cache map</dfn> is an <a>ordered map</a> whose [=map/entry=] consists of a [=map/key=] (a string that represents the name of a [=request response list=]) and a [=map/value=] (a [=request response list=]).
 
-    A <dfn id="dfn-name-to-cache-map">name to cache map</dfn> is an <a>ordered map</a> where the keys are strings that represents a name of the {{Cache}} object, and the values are {{Cache}} objects.
+    Each [=/origin=] has an associated [=name to cache map=].
 
-    Each [=/origin=] has an associated <a>name to cache map</a>.
+    When a [=name to cache map=] is referenced from within an algorithm, an attribute getter, an attribute setter, or a method, it designates the instance of the [=context object=]'s associated [=CacheStorage/global object=]'s [=environment settings object=]'s [=environment settings object/origin=], unless specified otherwise.
```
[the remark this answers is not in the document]
jungkees: Addressed.

docs/index.bs
```diff
@@ -1958,24 +1955,24 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
                 1. Set |r| to |request|'s [=Request/request=].
-                1. If |r|'s [=request/method=] is not \`<code>GET</code>\` and |options|.ignoreMethod is false, return a <a>promise</a> resolved with an empty array.
+                1. If |r|'s [=request/method=] is not \`<code>GET</code>\` and |options|.ignoreMethod is false, return [=a promise resolved with=] an empty array.
             1. Else if |request| is a string, then:
-                1. Set |r| to the associated [=Request/request=] of the result of invoking the initial value of {{Request}} as constructor with |request| as its argument. If this <a>throws</a> an exception, return a <a>promise</a> rejected with that exception.
-        1. Let |promise| be a new <a>promise</a>.
-        1. Run these substeps <a>in parallel</a>:
-            1. Let |responseArray| be an empty array.
+                1. Set |r| to the associated [=Request/request=] of the result of invoking the initial value of {{Request}} as constructor with |request| as its argument. If this [=throws=] an exception, return [=a promise rejected with=] that exception.
+        1. Let |realm| be the [=current Realm Record=].
```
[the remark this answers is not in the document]
jungkees: I referenced it from the example in https://html.spec.whatwg.org/#event-loop-for-spec-authors. But I think the relevant realm of the context object is correct indeed. Addressed as such.

docs/index.bs
```diff
@@ -1958,24 +1955,24 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-                1. For each |entry| of |entries|:
-                    1. Let |response| be null.
-                    1. If the <a>incumbent record</a> |incumbentRecord| of the corresponding <a>fetching record</a> in <a>request to response map</a> is not null, set |response| to a copy of |incumbentRecord|.\[[value]].
-                    1. Else, set |response| to a copy of |entry|[1].
-                    1. Add |response| to |responseArray|.
+                1. Let |requestResponses| be the result of running [=Query Cache=] with |r| and |options|.
+                1. [=list/For each=] |requestResponse| of |requestResponses|:
+                    1. Add a copy of |requestResponse|'s response to |responses|.
+            1. [=Queue a task=], on |promise|'s [=relevant settings object=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to perform the following steps:
+                1. Let |responseArray| be an empty JavaScript array, in |realm|.
```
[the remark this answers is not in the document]
jungkees: Addressed. But I want to make sure if I used "in realm" correctly in this change. I suppose the created frozen array is actually converted to a JavaScript array by Web IDL. If so, designating the realm when creating a frozen array as I did here makes sense?

docs/index.bs
jungkees: I intentionally factored the definition of |r| out of the nested blocks to reduce redundancy. It seems not a big deal as the local variable will likely to be stored in a register in machine level. But which is a better practice? I thought defining them within for-blocks is analogous to doing the same in implementations.

docs/index.bs
jungkees: I don't think we changed it so for that particular reason. The only apparent reason I recall is to make it return early when requests with a non-GET method come in. I'll inspect how we can improve/simplify this part of the algorithm as a separate work.

docs/index.bs
jungkees: Addressed.

docs/index.bs
jungkees: Currently, the loop breaks there, but the already-created promises of the ongoing fetches will be handled by the rejection handler of the aggregate promise for them. Is it better/okay to just return a rejected promise at that point without handling those ongoing promise tasks?

docs/index.bs
jungkees: The same thought/discussion as above.

docs/index.bs
jungkees: Yes. Addressed.

docs/index.bs
jungkees: Removed.

docs/index.bs
jungkees: In the above case, I changed it to using a frozen array but still left "in realm" when creating the frozen array. If we use frozen array here, don't we need to specify a realm?

docs/index.bs
jungkees: Addressed.

docs/index.bs
jungkees: Addressed.

docs/index.bs
jungkees: I'm not sure we can. For this particular case, I removed "in parallel". I think that's okay as the fulfillment handler's already scheduled async in the microtask queue. But for other similar cases, I didn't change them in this PR as I wasn't sure if they can all run in the main thread. I'll look at them as a separate work.

jungkees:
@jakearchibald, thanks for reviewing. I addressed your comments. PTAL.

docs/index.bs
jakearchibald: Implementations may optimise by reusing variables, but I don't think it's something we should do in the spec as it makes it harder to read. The way it's currently scoped, I expected the second loop to use the current value of |r|. or |r| to be used after both loops.

I think variables should be as shallowly-scoped as possible, so if they're only used in a loop, then they can be scoped to that loop.

In this case, it means |r| will only ever be "let", never "set", so readers can tell it's a constant.

jakearchibald:
@annevk @domenic 

We've got a couple of instances in the service worker spec that follow this pattern:

> 1. Let *realm* be the context object's relevant realm.
> 1. Then later, in parallel:
> 1. Resolve promise with a frozen array created from *someList*, in *realm*.

…where *someList* is an infra list.

Do we need to do this? I thought it would be enough to resolve the promise with *someList*, and IDL takes care of the FrozenArray creation in the correct realm, but I can't find a direct spec reference for that.

docs/index.bs
jakearchibald: Yeah I think it's ok to do that, and an early-exit is clearer if the behaviour is identical.

docs/index.bs
jungkees: Convinced. I'll change them back. Thanks!

docs/index.bs
jakearchibald: This one is slightly different since it's in parallel steps.

Right now we're waiting on fetches to complete even though we know the operation is a failure, so it seems better to reject *responsePromise*, abort the ongoing fetches, then abort the steps.

docs/index.bs
jakearchibald: Although it's on the microtask queue, it's still blocking the event loop. I think we need to create a promise and return it.

domenic:
You need to convert lists into FrozenArrays while specifying the realm; there's no way IDL could automatically figure out what kind of object you want to convert it to, or what global you want to create it in.

jungkees:
Okay. So, my try seems to be okay here.

annevk:
@jakearchibald you don't resolve from "in parallel". You need to queue a task on an event loop. That event loop will have a realm that you can use to create the frozen array and resolve the promise. (You don't get in parallel access to resolve either.)

jungkees:
@annevk, that queue a task step is missing in @jakearchibald's example above.

> Resolve promise with a frozen array created from someList, in realm.

is run in a queued task as you pointed indeed.

jakearchibald:
@annevk "resolve" automatically queues a task on the promise's event loop https://www.w3.org/2001/tag/doc/promises-guide#shorthand-manipulating.

jakearchibald:
@domenic The return type is defined in IDL, so I thought it be capable of some casting. A promise created in one realm could resolve with an object from another, but that feels like the exception.

I guess I could write my own helper to do this.

annevk:
@jakearchibald that hook is broken. It doesn't allow you to specify the task source.

jungkees:
Does it make sense to define some sort of default task source (used when not specified otherwise) for promise jobs? That hook is actually handy and makes it read simple in many cases.

annevk:
Action-at-a-distance leads to harder to understand algorithms, I think. If someone fixed the hook to ensure it takes all the arguments it needs it would be much clearer.

docs/index.bs
jungkees: Addressed.

docs/index.bs
jungkees: Done.

docs/index.bs
jungkees: Done.

docs/index.bs
jungkees: Addressed.

docs/index.bs
jungkees: Addressed.

docs/index.bs
jungkees: Yes, actually the fulfillment handler steps were very much incorrect. I made them run in the event loop and create a promise there and resolve/reject the promise from the parallel steps. Also, while changing them, I changed the interface of Batch Cache Operations such that it runs synchronously without returning a promise and the call sites invoke it from a promise job.

jungkees:
> If someone fixed the hook to ensure it takes all the arguments it needs it would be much clearer.

Yes, I think this would help simplifying the caller side steps. I'll take a look when I find time for it.

jungkees:
@jakearchibald, I uploaded another snapshot addressing your additional comments. PTAL.

jungkees:
@jakearchibald, thanks a lot for your review!

Issue #1196, opened 2017-09-21
Issues with "Handle Functional Event"


jakearchibald:
Sorry for the vague title, I couldn't explain it briefly.

https://w3c.github.io/ServiceWorker/#handle-functional-event.

"Handle Functional Event" wants an `ExtendableEvent`, a registration, and some callback steps. These callback steps are run on the active service worker's event loop. However, I need to be on the service worker's event loop in order to create the `ExtendableEvent`.

Maybe the algorithm should take the event name instead of an `ExtendableEvent`, and we just expect other specs to call "Update Service Worker Extended Events Set" themselves.

Also, it feels like "Update Service Worker Extended Events Set" only needs to take an `ExtendableEvent`, since the active worker could be determined from the global the `ExtendableEvent` was created in.

So other specs would do:

1. [Handle a functional event](https://w3c.github.io/ServiceWorker/#handle-functional-event) for *registration* and `"eventName"` by running the following steps:
  1. Let *event* be a trusted event (…new event object).
  1. Dispatch *event* on the [current global object](https://html.spec.whatwg.org/multipage/webappapis.html#current-global-object).
  1. [Update Service Worker Extended Events Set](https://w3c.github.io/ServiceWorker/#update-service-worker-extended-events-set) with *event*.

jungkees:
> "Handle Functional Event" wants an ExtendableEvent, a registration, and some callback steps. These callback steps are run on the active service worker's event loop. However, I need to be on the service worker's event loop in order to create the ExtendableEvent.

Right. I checked two call sites, Notifications API and Push API. Notifications API was creating an event within the callbackSteps without giving an event argument to Handle Funcational Event. (So, the arguments and the algorithm input params needed to be clarified.) Push API was creating an event from the caller's steps. Yeah, so we need to clarify this. I left review comments in #1199.

monica-ch:
Bug triage: Follow Jake's suggestion https://github.com/w3c/ServiceWorker/issues/1196#issuecomment-331160864

yoshisatoyanagisawa:
I think #1813 is the final piece for the issue. Let me close the issue. Please feel free to reopen if not.

PR #1199, opened 2017-09-25
Making functional events simpler

jakearchibald:
This is a fix for https://github.com/w3c/ServiceWorker/issues/1196.

I took inspiration from [fire an event](https://dom.spec.whatwg.org/#concept-event-fire).

This should clear up the confusion around realms and it make it easier for other specs.

@jungkees if you're happy with this, I'll port this to V1 and submit PRs to the specs that fire functional events.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1199.html" title="Last updated on Jun 21, 2018, 1:05 PM GMT (2c85816)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1199/657540a...2c85816.html" title="Last updated on Jun 21, 2018, 1:05 PM GMT (2c85816)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2146,7 +2146,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   <section>
     <h3 id="firing-functional-events">Firing Functional Events</h3>
 
-    To request a <a>functional event</a> dispatch to the [=service worker registration/active worker=] of a [=/service worker registration=], specifications *may* invoke <a>fire a functional event</a>.
+    To request a [=functional event=] dispatch to the [=service worker registration/active worker=] of a [=/service worker registration=], specifications *may* invoke [=Fire Functional Event=].
   </section>
 </section>
 
@@ -2809,13 +2809,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   </section>
 
   <section algorithm>
-    <h3 id="fire-functional-event-algorithm"><dfn export>Fire a Functional Event</dfn></h3>
+    <h3 id="fire-functional-event-algorithm"><dfn export lt="Fire Functional Event|Fire a functional event">Fire Functional Event</dfn></h3>
 
       : Input
-      :: |eventName|, a DOMString
+      :: |eventName|, a string
       :: |eventConstructor|, an event constructor that extends {{ExtendableEvent}}
       :: |registration|, a [=/service worker registration=]
-      :: |initializationSteps|, optional steps to initialise |event|, constructed from |eventConstructor|
+      :: |initialization|, optional property initialization for |event|, constructed from |eventConstructor|
       :: |postDispatchSteps|, optional steps to run on the [=service worker registration/active worker=]'s event loop, with |dispatchedEvent| set to the instance of |eventConstructor| that was [=dispatched=].
       : Output
       :: None
@@ -2833,7 +2833,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Invoke [=Run Service Worker=] algorithm with |activeWorker| as the argument.
       1. [=Queue a task=] |task| to run these substeps:
           1. Let |event| be the result of [=creating an event=] with |eventConstructor| and the [=relevant realm=] of |activeWorker|'s [=service worker/global object=].
-          1. If |initializationSteps| is not null, then initialize |event| with |initializationSteps|.
+          1. If |initialization| is not null, then initialize |event| with |initialization|.
           1. [=Dispatch=] |event| on |activeWorker|'s [=service worker/global object=].
           1. Invoke [=Update Service Worker Extended Events Set=] with |activeWorker| and |event|.
           1. If |postDispatchSteps| is not null, then run |postDispatchSteps| passing |event| as |dispatchedEvent|.
@@ -2846,7 +2846,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       <div class="example">
         To fire an "`amazingthing`" event (which is of type `AmazingThingEvent`) on a particular |serviceWorkerRegistration|, and initialize the event object's properties, the prose would be:
 
-        1. [=Fire a functional event=] named "`amazingthing`" using `AmazingThingEvent` on |serviceWorkerRegistration| with the following properties:
+        1. [=Fire Functional Event=] "`amazingthing`" using `AmazingThingEvent` on |serviceWorkerRegistration| with the following properties:
             : propertyName
             :: value
             : anotherPropertyName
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -2388,13 +2388,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+    <h3 id="firing-functional-events">Firing Functional Events</h3>
 
-    To request a <a>functional event</a> dispatch to a [=/service worker=], specifications *may* invoke <a>Handle Functional Event</a> algorithm with its [=/service worker registration=] <var ignore>registration</var> and the algorithm |callbackSteps| as the arguments.
-
-    Specifications *may* define an algorithm |callbackSteps| where the corresponding <a>functional event</a> can be created and fired with specification specific objects. The algorithm is passed <var ignore>globalObject</var> (a {{ServiceWorkerGlobalScope}} object) at which it *may* fire its <a>functional events</a>. This algorithm is called on a <a>task</a> <a lt="queue a task">queued</a> by <a>Handle Functional Event</a> algorithm.
-
-    Note: See an <a href="https://notifications.spec.whatwg.org/#activating-a-notification">example</a> hook defined in <a biblio data-biblio-type="informative" lt="notifications">Notifications API</a>.
+    To request a <a>functional event</a> dispatch to the [=service worker registration/active worker=] of a [=/service worker registration=], specifications *may* invoke <a>fire a functional event</a>.
```
jungkees: > Nit: SW uses some PascalCase with intervening space for algorithm names. HTML, Fetch, and many others use lower case letters. Streams use PascalCase.. I always wanted to discuss if we better change our naming practice for algorithms' name?

docs/index.bs
```diff
@@ -3105,35 +3101,57 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   </section>
 
   <section algorithm>
-    <h3 id="handle-functional-event-algorithm"><dfn export>Handle Functional Event</dfn></h3>
+    <h3 id="fire-functional-event-algorithm"><dfn export>Fire a Functional Event</dfn></h3>
```
jungkees: Nit: Fire Functional Event for consistency, but opened to any suggestion for naming practice.

docs/index.bs
```diff
@@ -3105,35 +3101,57 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
 
       : Input
-      :: |event|, an {{ExtendableEvent}} object
+      :: |eventName|, a DOMString
+      :: |eventConstructor|, an event constructor that extends {{ExtendableEvent}}
       :: |registration|, a [=/service worker registration=]
-      :: |callbackSteps|, an algorithm
+      :: |initializationSteps|, optional steps to initialise |event|, constructed from |eventConstructor|
```
jungkees: It seems we should explicitly provide |createdEvent| as an argument to |initializationSteps| so the callers will set up the properties to it when needed. Just as you did with |dispatchedEvent| to |postDispatchSteps|.

docs/index.bs
```diff
@@ -3105,35 +3101,57 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      1. If the time difference in seconds calculated by the current time minus |registration|'s [=last update check time=] is greater than 86400, invoke [=Soft Update=] algorithm with |registration|.
+
+      <div class="example">
+        To fire an "`amazingthing`" event (which is of type `AmazingThingEvent`) on a particular |serviceWorkerRegistration|, and initialize the event object's properties, the prose would be:
+
+        1. [=Fire a functional event=] named "`amazingthing`" using `AmazingThingEvent` on |serviceWorkerRegistration| with the following properties:
+            : propertyName
+            :: value
+            : anotherPropertyName
+            :: anotherValue
```
jungkees: It seems callers should give us an |initializationSteps| where they set up the properties themselves (rather than providing the properties to Fire Functional Event as an argument)?

jungkees:
@jakearchibald, I'm happy with the changes this PR's addressing. When we get to the final snapshot, backporting to V1 seems nice.

docs/index.bs
jakearchibald: I did it this way to match invocations of https://dom.spec.whatwg.org/#concept-event-fire, but I'm not against making it sub-steps.

jungkees: I thought the |initializationSteps| would be a place where the properties can be initialized. But checking out the DOM spec link that you shared, I'm fine with that too.

docs/index.bs
```diff
@@ -3105,35 +3101,57 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   </section>
 
   <section algorithm>
-    <h3 id="handle-functional-event-algorithm"><dfn export>Handle Functional Event</dfn></h3>
+    <h3 id="fire-functional-event-algorithm"><dfn export>Fire a Functional Event</dfn></h3>
 
       : Input
-      :: |event|, an {{ExtendableEvent}} object
+      :: |eventName|, a DOMString
```
jungkees: I think it needs to be just a string for internal algorithms.

jakearchibald:
@jungkees I've rebased this and also updated the v1 spec. I think it's best to stick with property initialisation rather than a complete set of steps for now, just to be similar to "fire an event". Happy to change it later if that turns out to be bad.

docs/index.bs
```diff
@@ -2144,13 +2144,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+    <h3 id="firing-functional-events">Firing Functional Events</h3>
 
-    To request a <a>functional event</a> dispatch to a [=/service worker=], specifications *may* invoke <a>Handle Functional Event</a> algorithm with its [=/service worker registration=] <var ignore>registration</var> and the algorithm |callbackSteps| as the arguments.
-
-    Specifications *may* define an algorithm |callbackSteps| where the corresponding <a>functional event</a> can be created and fired with specification specific objects. The algorithm is passed <var ignore>globalObject</var> (a {{ServiceWorkerGlobalScope}} object) at which it *may* fire its <a>functional events</a>. This algorithm is called on a <a>task</a> <a lt="queue a task">queued</a> by <a>Handle Functional Event</a> algorithm.
-
-    Note: See an <a href="https://notifications.spec.whatwg.org/#activating-a-notification">example</a> hook defined in <a biblio data-biblio-type="informative" lt="notifications">Notifications API</a>.
+    To request a [=functional event=] dispatch to the [=service worker registration/active worker=] of a [=/service worker registration=], specifications *may* invoke [=Fire Functional Event=].
```
jungkees: This isn't an issue this PR brought. Would it be better to change the conformance requirement language to *should* or even *must*? I think there's a chance that some prospective specs might want to define their own steps intentionally. So, *should* seems to be a good requirement for this?

jungkees:
LGTM with a comment on the conformance requirement language. It'd be great if you could make PRs to the call sites :).

docs/index.bs
jakearchibald: I went for 'should', as there may be reasons why another spec needs to do it manually

jakearchibald:
Filed: 

https://github.com/whatwg/notifications/pull/135
https://github.com/w3c/push-api/pull/298
https://github.com/WICG/BackgroundSync/pull/146
https://github.com/w3c/payment-handler/pull/305

PR #1205, opened 2017-10-11
Warn readers that V1 is missing features compared to the nightly

jakearchibald:
We're seeing a lot of confusion around V1 and the nightly. As in, developers are linking to the V1 spec and wondering why they can't find particular features.

I've added a warning to the V1 spec. What do you think?

screen shot of spec

### DIFF
--- a/docs/v1/index.bs
+++ b/docs/v1/index.bs
@@ -12,6 +12,9 @@ Editor: Marijn Kruisselbrink, Google, mek@chromium.org
 Repository: w3c/ServiceWorker
 Group: serviceworkers
 !Tests: <a href=https://github.com/w3c/web-platform-tests/tree/master/service-workers>web-platform-tests service-workers/</a> (<a href=https://github.com/w3c/web-platform-tests/labels/service-workers>ongoing work</a>)
+Warning: Custom
+Custom Warning Title: V1 Branch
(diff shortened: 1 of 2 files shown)


ylafon:
I'd rather have a message indicating that V1 (Vn in general) contains only stable and implemented features and that all the experimental and new features (aka, the enjoyable parts) are in the nightly snapshot.

Having a v1 is not really required by W3C process, we use this to signal stable releases where having implemented and tested features is a requirement, to exit CR and go to REC.

jungkees:
I think we can use wordings from https://github.com/w3c/ServiceWorker#spec-and-api-development:

> Service Workers Nightly is a living document. Service Workers 1 is a subset of the nightly version that is advancing toward a W3C Recommendation. For implementers and developers who seek all the latest features, Service Workers Nightly is a right document that is constantly reflecting new requirements. For version 1, contributors plan to focus on fixing bugs and resolving compatibility issues without including new features from the nightly version.

jakearchibald:
I've updated the text.

domenic:
I liked the original text.

jungkees:
Yeah, but this is at least the text that we agreed on in the discussion back then.

jakearchibald:
@domenic yeah, I'd like to revisit this message once we switch to a snapshot model (which we're planning to do after v1)

PR #1207, opened 2017-10-13
Remove foreign fetch

jakearchibald:
Removing the foreign fetch parts. I've left the <link rel=serviceworker> stuff in there for now, we may discuss that at TPAC.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/remove-foreign-fetch.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/0bbf95c...781bef8.html)


jakearchibald:
Creating a PR to remove the tests…

jakearchibald:
https://github.com/w3c/web-platform-tests/pull/7762

docs/index.bs
```diff
@@ -1725,11 +1557,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       </thead>
       <tbody>
         <tr>
-          <td><dfn event id="service-worker-global-scope-install-event"><code>install</code></dfn></td>
```
mkruisselbrink: This block should still be there, right? Just with ExtendableEvent rather than InstallEvent

jakearchibald:
Hah, I forgot to escape the html, I meant the `<link rel=serviceworker>` stuff

docs/index.bs
jakearchibald: Ah yes, correct

jakearchibald:
Before I land this… should we keep `InstallEvent` now we have it? Or just reintroduce it later when we need it?

wanderview:
> Before I land this… should we keep InstallEvent now we have it? Or just reintroduce it later when we need it?

Currently we have the weird thing in `waitUntil()` steps that does special logic for install:

https://w3c.github.io/ServiceWorker/#wait-until-method

If we kept `InstallEvent` we could move that to an `InstallEvent.waitUntil()` that implements that behavior instead of special casing it ExtendableEvent.  Not sure if thats worth it or not.

jakearchibald:
F2F: Keeping the header (for navigation requests) is interesting to CDNs. We'll talk to them this week and find out how important it is.

Going back to `ExtendableEvent` for install seems fine. Mozilla still uses `ExtendableEvent`.

jungkees:
> If we kept InstallEvent we could move that to an InstallEvent.waitUntil() that implements that behavior instead of special casing it ExtendableEvent. Not sure if thats worth it or not.

That behavior's already spec'ed in Install algorithm and Activate algorithm. I think we can safely change the prose in the waitUntil section to notes. Filed https://github.com/w3c/ServiceWorker/issues/1227.

So, we can just use `ExtendableEvent` as discussed in the F2F.

jakearchibald:
I've removed the link header & tag, but we should talk more to CDNs and add it again if needed.

Still TODO: remove from tests and HTML spec.

jakearchibald:
HTML PR: https://github.com/whatwg/html/pull/3233
Tests PR: https://github.com/w3c/web-platform-tests/pull/7762

PR #1210, opened 2017-10-17
Block partial responses entering the cache

### LINKED ISSUE #937: Block 206s entering the cache (via put & addAll)
Right now it seems like this would only happen by accident (passthrough caching), and it's something we can open up later if we need to.

jungkees:
This change adds a guard to put() and addAll() that blocks 206 responses
entering the cache.

Fixes #937.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/not-allow-206-cache.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/1d8fc5a...e2f8f08.html)


jungkees:
The corresponding patches for this change already landed on Chromium, Gecko and WPT. PTAL.

PR #1211, opened 2017-10-19
Meta: mention type:untestable and type:missing-coverage wpt labels

foolip:


foolip:
This is to match the wording in many other places, but mostly because I wrote a heuristic to find testing policies in repos, and it was to look for the strings "web-platform-tests" and "type:untestable" in the same file.

jakearchibald:
Cheers!

PR #1218, opened 2017-11-02
add w3cid for editors to help publication

ylafon:


jakearchibald:
I take your word for it. Certainly doesn't break anything. LGTM.

docs/index.bs
```diff
@@ -42,9 +42,9 @@ spec: infra;
 </pre>
 
 <pre class="anchors">
-spec: push; urlPrefix: https://w3c.github.io/push-api/
+spec: push; urlPrefix: https://www.w3.org/TR/push-api/
```
jakearchibald: Why can't we link to the living spec?

ylafon: Ah! I didn't mean to add this commit to this PR...
In this case, as the spec is auto-published, the TR latest WD and the edcopy are always the same.

PR #1220, opened 2017-11-08
Remove <link rel=serviceworker>

sideshowbarker:
The is for the case where we end up deciding to keep the “serviceworker” Link header but remove the `<link rel=serviceworker>` element.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/sideshowbarker/remove-link-rel-serviceworker.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/ServiceWorker/4b59263...1779dd9.html)

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1760,7 +1760,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 <section>
   <h2 id="link-type-serviceworker">Link type "`serviceworker`"</h2>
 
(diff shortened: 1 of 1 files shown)


jungkees:
https://github.com/w3c/ServiceWorker/pull/1207#issuecomment-344964049 removed the Link header part too. If we come to a conclusion to add the Link header, that part should be added back.

jungkees:
@sideshowbarker, thank you for your follow-up. Let's see what happens to the Link header part.

PR #1229, opened 2017-11-17
Introduce a parallel queue for running Jobs

### LINKED ISSUE #1224: Run Job does not specify which event loop to queue its tasks on
I might be missing something here, but the Run Job algorithm currently just says "Queue a task to run [this job]", and I don't understand what it means to "queue a task" on a unspecified event loop. The definition of "queue a task" just mentions "the relevant event loop", but I'm not sure what that means in the context of the "Run Job" algorithm. When "Run Job" is invoked by "Schedule Job" I suppose it could arguably be the event loop of the document that called register or update (but does that mean that if the document goes away before the task runs, registration should be aborted?), and similarly when invoked by "Finish Job" it could arguably be the event loop the task for the just finshed job was running on, but that still has the same problem with what happens if that event loop goes away (or maybe it is still supposed to be the task queue/event loop of whoever originally scheduled the job?).

Is there supposed to be some separate SW processing event loop or something these tasks are supposed to run on? It doesn't seem like posting tasks to event loops associated with arbitrary (or even non-arbitrary) documents and/or workers could ever work, as that would mean the registration wouldn't complete if the document goes away before registration finishes (well, if registration didn't start before the event loop went away, which I guess might be okay?)

Either way it seems we should clarify what the relevant event loop is for "Run Job"

jungkees:
This change defines a parallel queue called the service worker manager
where the instances of Run Job steps are queued and run in order.

Fixes #1224.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -258,6 +258,14 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   </section>
 
   <section>
+    <h3 id="user-agent-bootup">User Agent Boot up</h3>
+
+    A user agent has an associated <dfn export id="dfn-service-worker-manager">service worker manager</dfn>.
+
+    A user agent *must* [=start a new parallel queue=] when it boots up and set the [=service worker manager=] to the result value.
+  </section>
+
+  <section>
     <h3 id="user-agent-shutdown">User Agent Shutdown</h3>
 
     A user agent *must* maintain the state of its stored [=/service worker registrations=] across restarts with the following rules:
@@ -2223,15 +2231,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       : Output
       :: none
 
-      1. Assert: |jobQueue| [=queue/is not empty=].
-      1. [=Queue a task=] to run these steps:
+      1. [=Enqueue the following steps=] to the [=service worker manager=]:
+          1. Assert: |jobQueue| [=queue/is not empty=].
           1. Let |job| be the first [=queue/item=] in |jobQueue|.
-          1. If |job|'s [=job type=] is *register*, run [=Register=] with |job| [=in parallel=].
-          1. Else if |job|'s [=job type=] is *update*, run [=Update=] with |job| [=in parallel=].
+          1. If |job|'s [=job type=] is *register*, run [=Register=] with |job|.
+          1. Else if |job|'s [=job type=] is *update*, run [=Update=] with |job|.
 
               Note: For a register job and an update job, the user agent delays queuing a task for running the job until after a {{Document/DOMContentLoaded}} event has been dispatched to the document that initiated the job.
 
-          1. Else if |job|'s [=job type=] is *unregister*, run [=Unregister=] with |job| [=in parallel=].
+          1. Else if |job|'s [=job type=] is *unregister*, run [=Unregister=] with |job|.
   </section>
 
   <section algorithm>
--- a/docs/v1/index.bs
+++ b/docs/v1/index.bs
@@ -247,6 +247,14 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   </section>
 
   <section>
+    <h3 id="user-agent-bootup">User Agent Boot up</h3>
+
+    A user agent has an associated <dfn export id="dfn-service-worker-manager">service worker manager</dfn>.
+
+    A user agent *must* [=start a new parallel queue=] when it boots up and set the [=service worker manager=] to the result value.
+  </section>
+
+  <section>
     <h3 id="user-agent-shutdown">User Agent Shutdown</h3>
 
     A user agent *must* maintain the state of its stored [=/service worker registrations=] across restarts with the following rules:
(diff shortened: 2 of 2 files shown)


docs/index.bs
```diff
@@ -257,6 +257,14 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         Note: A user agent may use a separate task source for each functional event type in order to avoid a head-of-line blocking phenomenon for certain functional events.
   </section>
 
+  <section>
+    <h3 id="user-agent-bootup">User Agent Boot up</h3>
+
+    A user agent has an associated <dfn export id="dfn-service-worker-manager">service worker manager</dfn>.
+
+    A user agent *must* [=start a new parallel queue=] when it boots up and set the [=service worker manager=] to the result value.
```
jakearchibald: Does this need to be scoped to the whole user agent, or is per-origin enough?

docs/index.bs
```diff
@@ -2223,15 +2231,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
           1. Let |job| be the first [=queue/item=] in |jobQueue|.
-          1. If |job|'s [=job type=] is *register*, run [=Register=] with |job| [=in parallel=].
-          1. Else if |job|'s [=job type=] is *update*, run [=Update=] with |job| [=in parallel=].
+          1. If |job|'s [=job type=] is *register*, run [=Register=] with |job|.
+          1. Else if |job|'s [=job type=] is *update*, run [=Update=] with |job|.
 
               Note: For a register job and an update job, the user agent delays queuing a task for running the job until after a {{Document/DOMContentLoaded}} event has been dispatched to the document that initiated the job.
 
-          1. Else if |job|'s [=job type=] is *unregister*, run [=Unregister=] with |job| [=in parallel=].
+          1. Else if |job|'s [=job type=] is *unregister*, run [=Unregister=] with |job|.
```
jakearchibald: This is a good step. Eventually I'd like to get rid of the "job" concept entirely, and replace it with appending steps to the appropriate parallel queue.

docs/index.bs
jakearchibald: Btw totally happy for this to merge if there's a reason the queue needs to be across the whole browser.

jungkees: This intentionally suggests to be a parallel execution context across the user agent. This matches pretty much what the implementations do (at least Chromium). We can think of the parallel execution context of the parallel queue as a thread in a browser process in Chromium for instance.

docs/index.bs
jungkees: That'd mean the implementations should create and maintain a separate thread for each scope. That may be an ideal design which allows parallel register/update/unregister even, but a single (off-the-main) thread executing the jobs from all the job queues is what we currently have. I also would like to hear what implementers think on this.

jungkees:
One thing I'd thought of and didn't put in the change was run the Schedule Job's steps in the parallel queue as well. That'd make it more congruent to what Chromium does (with IPC). But I thought it would add somewhat unnecessary complexity to the spec.

For now, I think using a separate (off-the-main) thread running jobs without breaking the order in each job queue would be good enough.

docs/index.bs
jakearchibald: Is there any reason the registration of a service worker on origin A should block the registration of a service worker on origin B?

jungkees: No. They are independent. Per-origin parallel queues would work and provide better concurrency conceptually. In this change, I just tried to match the current implementation, especially Chromium.

jakearchibald: I think the spec should allow for maximum concurrency while maintaining the behaviour we want. Implementations are welcome to be less concurrent, but the spec shouldn't reflect this unless developers come to rely on a particular lack of concurrency.

@annevk is that fair?

annevk: For shared workers I went with a note after requiring a single one for the user agent:

> Each user agent has a single shared worker manager for simplicity. Implementations could use one per origin; that would not be observably different and enables more concurrency.

jungkees: Implementers' feedback would also be great.

/cc @mattto, @wanderview, @mkruisselbrink, @aliams, @hober

talea11:
عمال الخدمة
مسودة توصية مرشح W3C ،12 يوليو 2022

مزيد من التفاصيل حول هذا المستند
هذه النسخة:
https://www.w3.org/TR/2022/CRD-service-workers-20220712/
أحدث نسخة منشورة:
https://www.w3.org/TR/service-workers/
مسودة المحرر:
https://w3c.github.io/ServiceWorker/
تاريخ:
https://www.w3.org/standards/history/service-workers
تقرير التنفيذ:
https://wpt.fyi/service-workers
تعليق:
جيثب
مضمنة في المواصفات
المحررين:
جيك أرشيبالد ( Google )
مارين كرويسلبرينك ( Google )
المحررين السابقين:
أليكس راسل ( Google )
جونغكي سونغ ( مايكروسوفت ‚مثلت سامسونج حتى أبريل 2018 )
الاختبارات:
العاملون في خدمة اختبارات منصة الويب / ( العمل الجاري )
حقوق النشر © 2022 W3C ® ( MIT ، ERCIM ، Keio ، Beihang ). تنطبق قواعد مسؤولية W3C والعلامة التجارية وترخيص المستندات المسموح بها .

talea11:
عمال الخدمة
مسودة توصية مرشح W3C ،12 يوليو 2022

مزيد من التفاصيل حول هذا المستند
هذه النسخة:
https://www.w3.org/TR/2022/CRD-service-workers-20220712/
أحدث نسخة منشورة:
https://www.w3.org/TR/service-workers/
مسودة المحرر:
https://w3c.github.io/ServiceWorker/
تاريخ:
https://www.w3.org/standards/history/service-workers
تقرير التنفيذ:
https://wpt.fyi/service-workers
تعليق:
جيثب
مضمنة في المواصفات
المحررين:
جيك أرشيبالد ( Google )
مارين كرويسلبرينك ( Google )
المحررين السابقين:
أليكس راسل ( Google )
جونغكي سونغ ( مايكروسوفت ‚مثلت سامسونج حتى أبريل 2018 )
الاختبارات:
العاملون في خدمة اختبارات منصة الويب / ( العمل الجاري )
حقوق النشر © 2022 W3C ® ( MIT ، ERCIM ، Keio ، Beihang ). تنطبق قواعد مسؤولية W3C والعلامة التجارية وترخيص المستندات المسموح بها .

Issue #732, opened 2015-08-17
Remove frameType, maybe add ancestorOrigins


annevk:
Frame type does not address the use case of e.g., an embedded youtube.com showing fallback content. (Where youtube.com embedded in youtube.com should not show such a thing.)

jungkees:
F2F: Spec ancestorOrigins and remove frameType for now. We can get it back if it's really required in the future.

jungkees:
Just to check if the decision still holds. Is it still okay to remove `Client.frameType`?

jakearchibald:
F2F: Adding `ancestorOrigins` if fine as long as you can already get this information.

`frameType` is still useful for "auxiliary", which you can't determine from `ancestorOrigins`.

annevk:
When do you want to check `auxiliary`? Should it be `auxiliary` when `noopener` is set?

jakearchibald:
Pre F2F notes: come up with an answer to @annevk's question

jakearchibald:
F2F:

* We don't have solid use-cases for `frameType` right now, we should remove it and add it back if/when we get better data
* Keep `ancestorOrigins`, but it's only needed on `WindowClient`

wanderview:
Gecko bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1290936

jungkees:
@annevk, I'm planning to factor the steps of [`location.ancestorOrigins`](https://html.spec.whatwg.org/multipage/browsers.html#dom-location-ancestororigins) out as an algorithm that both `location.ancestorOrigins` and `WindowClient.ancestorOrigins` can invoke (taking an environment settings object as an argument.)

Currently, `location.ancestorOrigins`'s step 1 throws "SecurityError" if the relevant document's origin is not the same as entry settings object's origin. I don't think this would be meaningful for `WindowClient.ancestorOrigins` where the entry settings object's origin will always be the same as the client's origin. I'm not sure what the purpose of this security check in the first place.

In case we should retain this entry settings object checking step, should I make the algorithm take both relevant settings object and the entry settings object as arguments? Or any good idea how to define this algorithm?

/cc @domenic

annevk:
Yes, that security check is important for `Location` objects. Is a `WindowClient` always same-origin?

[...]

PR #1242, opened 2017-12-07
Add CR Plan for V1

jungkees:
This is a CR plan with a WIP table that maps WPT tests to each spec feature.

@ylafon @jakearchibald @jatindersmann - I'd like to discuss if this format and the information (that will be filled in the table) will meet the requirement to proceed and exit CR.

FYI, I asked @tabatkins if Bikeshed will support some feature to annotate the WPT status directly in the spec, and the answer was yes with no concrete schedule yet - at the end of the year at the earliest.

Until then, I plan to continue with this work. WDYT?

### DIFF
--- a/docs/v1/cr_plan.md
+++ b/docs/v1/cr_plan.md
@@ -1,0 +1,241 @@
+CR Plan for V1
+---
+
+At TPAC 2017, Service Workers WG [decided](https://www.w3.org/2017/11/07-serviceworker-minutes.html#item19) to move [Service Workers V1](https://w3c.github.io/ServiceWorker/v1) to CR with the reasons that we have three implementations (+ an ongoing work) and good WPT coverage for the interop.
+
+We use this document to:
+- Check WPT coverage.
+- Check if more than two browsers implemented the features.
+- Track the issues that should be resolved for this version.
+- Remove the features that are not ready for this version.
+
+Test environments
+---
+- Chromium: Chrome 61
+- Gecko: Firefox 58
+- EdgeHTML: Edge 40
+
+Feature and Test|Chromium|Gecko|EdgeHTML|WebKit
+-- | ------ | --- | ------ | -----
+**ServiceWorker.scriptURL**
+/service-workers/service-worker/serviceworkerobject-scripturl.https.html | | | | 
+/service-workers/service-worker/registration-script-url.https.html | | | | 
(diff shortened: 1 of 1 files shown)


ylafon:
Good work, 
Note that annotating the spec with WPT status would be useful for v2 and beyond, not for v1 as we reach CR now.
Note also that per charter, the exit criteria for CR is to have two implementation for each feature.

The format you propose for the table is good, needed for exiting CR, but not needed at all for entering it. For that we just need ot state the exit criteria and link to a report page that can be this CR Plan (but it can even be empty, at publication time, as it just needs to be complete when we exit).
Also, it would be good to identify the features at risk (we can even use a blanket statement that they are all at risks if implemented, but we can also list the ones that don't qualify as of today).

We basically just need to prepare the v1 CR doc and add in the SoTD that information (I can take care of that once the doc is ready)
Thanks,

jungkees:
@jakearchibald @ylafon, let's publish a CR. I think we can use this PR document to track CR progress (test coverage and compat) after we publish the first CR version.

@jakearchibald, please take a look at the following PRs. I'd like to merge them to publish a CR.

https://github.com/w3c/ServiceWorker/pull/1151 (Approved, I’m checking if we need a separate test for SW.)
https://github.com/w3c/ServiceWorker/pull/1293
https://github.com/w3c/ServiceWorker/pull/1329
https://github.com/w3c/ServiceWorker/pull/1332
https://github.com/w3c/ServiceWorker/pull/1354

From the [V1 issue list](https://github.com/w3c/ServiceWorker/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Version+1%22), https://github.com/w3c/ServiceWorker/issues/732 will be closed by merging https://github.com/w3c/ServiceWorker/pull/1329. I'd like to move other issues to V2. They are dependent to works in HTML and DOM.

ylafon:
As discussed on irc, we also need to select the test relevant to V1 only.
and strong +1 to move remaining issues to V2 (and add an editors' note if needed)

mattto:
@jungkees @jakearchibald I think this PR is probably obsolete now?

jungkees:
@mattto, the content of the PR seems outdated. The purpose of the PR is to track the process of entering CR. So, either using this PR by updating the content or putting together a new document would be fine.

/cc @aliams

PR #1243, opened 2017-12-07
Avoid array manipulation in parallel

jakearchibald:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1243.html" title="Last updated on Dec 8, 2017, 6:52 PM GMT (d5f5306)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1243/979ead7...d5f5306.html" title="Last updated on Dec 8, 2017, 6:52 PM GMT (d5f5306)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -662,7 +662,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
             1. [=Queue a task=] on |promise|'s [=relevant settings object=]'s [=responsible event loop=], using the [=DOM manipulation task source=], to run the following steps:
                 1. Let |registrationObjects| be a new [=list=].
                 1. [=list/For each=] |registration| of |registrations|, [=append=] the {{ServiceWorkerRegistration}} object associated with |registration| to |registrationObjects|.
-                1. [=Resolve=] |promise| with [=create a frozen array|a new frozen array of=] |registrationObjects|.
+                1. [=Resolve=] |promise| with [=create a frozen array|a new frozen array of=] |registrationObjects| in |promise|'s [=relevant Realm=].
         1. Return |promise|.
     </section>
 
@@ -1239,7 +1239,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                     * {{WindowClient}} objects are always placed before {{Client}} objects whose associated [=Client/service worker clients=] are [=worker clients=].
                     * {{WindowClient}} objects that have been [=focusing steps|focused=] are placed first sorted in the most recently [=focusing steps|focused=] order, and {{WindowClient}} objects that have never been [=focusing steps|focused=] are placed next sorted in their [=Client/service worker clients=]' creation order.
                     * {{Client}} objects whose associated [=Client/service worker clients=] are [=worker clients=] are placed next sorted in their [=Client/service worker clients=]' creation order.
-                1. [=Resolve=] |promise| with [=create a frozen array|a new frozen array of=] |clientObjects|.
+                1. [=Resolve=] |promise| with [=create a frozen array|a new frozen array of=] |clientObjects| in |promise|'s [=relevant Realm=].
         1. Return |promise|.
     </section>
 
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -1185,55 +1188,58 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+                    1. [=Append=] |windowClient| to |clientObjects|.
+                1. [=list/For each=] |client| in |matchedClients|:
+                    1. Let |clientObject| be the result of running [=Create Client=] algorithm with |client| as the argument.
+                    1. [=Append=] |clientObject| to |clientObjects|.
+                1. Sort |matchedClients| such that:
+                    * {{WindowClient}} objects are always placed before {{Client}} objects whose associated [=Client/service worker clients=] are [=worker clients=].
+                    * {{WindowClient}} objects that have been [=focusing steps|focused=] are placed first sorted in the most recently [=focusing steps|focused=] order, and {{WindowClient}} objects that have never been [=focusing steps|focused=] are placed next sorted in their [=Client/service worker clients=]' creation order.
+                    * {{Client}} objects whose associated [=Client/service worker clients=] are [=worker clients=] are placed next sorted in their [=Client/service worker clients=]' creation order.
+                1. [=Resolve=] |promise| with [=create a frozen array|a new frozen array of=] |clientObjects|.
```
bzbarsky: Which global?

docs/index.bs
```diff
@@ -653,13 +653,16 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+        1. Let |client| be the [=context object=]'s [=ServiceWorkerContainer/service worker client=].
+        1. Let |promise| be [=a new promise=].
+        1. Run the following steps [=in parallel=]:
+            1. Let |registrations| be a new [=list=].
+            1. [=map/For each=] |scope| → |registration| of [=scope to registration map=]:
+                1. If the [=url/origin=] of the result of [=URL parser|parsing=] |scope| is the [=same origin|same=] as |client|'s [=environment settings object/origin=], and |registration|'s [=uninstalling flag=] is unset, then [=append=] |registration| to |registrations|.
+            1. [=Queue a task=] on |promise|'s [=relevant settings object=]'s [=responsible event loop=], using the [=DOM manipulation task source=], to run the following steps:
+                1. Let |registrationObjects| be a new [=list=].
+                1. [=list/For each=] |registration| of |registrations|, [=append=] the {{ServiceWorkerRegistration}} object associated with |registration| to |registrationObjects|.
+                1. [=Resolve=] |promise| with [=create a frozen array|a new frozen array of=] |registrationObjects|.
```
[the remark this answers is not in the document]
jakearchibald: Ah, I thought if it wasn't otherwise stated, it'd get a default from the event loop, but yeah, one loop can have many realms.

Cheers, will fix that.

FWIW I'm not happy with the association of registration object to registration here, but that's for another PR I think.

jakearchibald:
@bzbarsky for my own learning, is it fine to create objects without specifying a realm before going in parallel? As in, is the realm assumed to be the context object's?

bzbarsky:
> is it fine to create objects without specifying a realm before going in parallel?

"yes".  At that point there is a well-defined current realm.  It's not the context object's; it's that of the function that was called.  Since JS object creation uses the current realm, if there is one, you're good for creating JS things.

For platform objects, you generally need to specify the realm, though there has been some talk of defaulting to the context object's realm somehow.

docs/index.bs
```diff
@@ -1188,55 +1188,58 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
                 1. If |client| is an [=environment settings object=], then:
-                    1. If |client| is not a <a>secure context</a>, continue to the next iteration of the loop.
-                1. Else:
-                    1. If |client|’s <a>creation URL</a> is not a <a>potentially trustworthy URL</a>, continue to the next iteration of the loop.
-                1. If |options|.{{ClientQueryOptions/includeUncontrolled}} is false, then:
-                    1. If |client|'s <a>active service worker</a> is not the associated [=ServiceWorkerGlobalScope/service worker=], continue to the next iteration of the loop.
-                1. If |options|.{{ClientQueryOptions/includeReserved}} is false, then:
-                    1. If |client|'s [=environment/execution ready flag=] is unset, continue to the next iteration of the loop.
+                    1. If |client| is not a [=secure context=], [=continue=].
+                1. Otherwise:
```
jungkees: Nice cleanup! I think "Else" is better here for consistency though.

jakearchibald: Will change it for consistency, but what's the correct form here @annevk? https://infra.spec.whatwg.org/#variables uses "otherwise", but infra doesn't specifically cover if/else/otherwise.

annevk: We should define conditionals more clearly I suppose, raise it? I suspect we'd go with "Otherwise" given the overwhelming precedent, but who knows.

Issue #595, opened 2014-12-26
Default scope to register()


mattto:
Currently the default scope is "/" which will likely result in failure due to the path restriction. There was some suggestions to revise it around https://github.com/slightlyoff/ServiceWorker/issues/468#issuecomment-60584294 but it seems not decided yet.

Chrome is happy making it something like "./". I'm not actually sure the difference between './' and new URL('./', scriptURL).href

mattto:
I think it's ok for the default url and base url to be unrelated. ie., the following seems acceptable to me:

// example.com/foo/page.html
sw.register("bar/sw.js"); // scope == "/foo/bar/"
sw.register("bar/sw.js", { scope: "./" }); // scope == "/foo"

Anyway, this should be decided as soon as possible, as changing it later will likely break sites.

jungkees:
> I think it's ok for the default url and base url to be unrelated.

Agreed. The default url for a scope seems more like just a constraint than a general rule for parsing scope urls.

/cc @jakearchibald @slightlyoff @annevk

jungkees:
The default scope text has been sorted out, and the security condition for checking a scope against the location of the script url has been added: f23c1d4e566847feebcfee8814132b0f8a8d56f1.

jakearchibald:
Agreed with @mattto's example, spec look good too!

KenjiBaheux:
Shall this bug be closed?

Spec changes tracking note:
 - fixed in Blink (M41 via https://code.google.com/p/chromium/issues/detail?id=445154#c4).

jungkees:
Yes, it's done! Closing.

PR #1251, opened 2017-12-19
Remove fragment identifiers from script url and scope url

### LINKED ISSUE #1249: Should the fragment identifier be stripped from the scriptURL?
Should the fragment identifier be stripped from the scriptURL? I cannot find any indication that we should in the specification. However, Firefox and Chrome seem to do so and it is tested via:
- service-workers/service-worker/serviceworkerobject-scripturl.https.html

jungkees:
The spec and the implementations did not agree on the behavior: the spec
didn't remove fragment identifiers and had ServiceWorker.scriptURL and
ServiceWorkerRegistration.scope return urls without stripping them.
Chrome, Firefox, and Edge don't return fragments.

This changes the spec to remove the fragment indentifiers from the
script url and scope url given to register() method before storing them,
in order to match the implementations.

Fixes #1249.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1251.html" title="Last updated on Dec 20, 2017, 11:11 AM GMT (2cf9fc0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1251/90c04cd...2cf9fc0.html" title="Last updated on Dec 20, 2017, 11:11 AM GMT (2cf9fc0)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -637,6 +637,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. Let |client| be the <a>context object</a>'s [=ServiceWorkerContainer/service worker client=].
         1. Let |clientURL| be the result of <a lt="URL parser">parsing</a> |clientURL| with the <a>context object</a>'s <a>relevant settings object</a>'s <a>API base URL</a>.
         1. If |clientURL| is failure, return a <a>promise</a> rejected with a <code>TypeError</code>.
+        1. Set |clientURL|'s [=url/fragment=] to null.
         1. If the [=environment settings object/origin=] of |clientURL| is not |client|'s [=environment settings object/origin=], return a |promise| rejected with a "{{SecurityError}}" {{DOMException}}.
         1. Let |promise| be a new <a>promise</a>.
         1. Run the following substeps <a>in parallel</a>:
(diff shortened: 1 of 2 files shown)


cdumez:
FYI, I have just landed a patch in WebKit to strip the fragment identifier in WebKit, to align with Firefox and Chrome.

wanderview:
FWIW, it seems we added the fragment stripping here:

https://bugzilla.mozilla.org/show_bug.cgi?id=1130688

But I don't see anything in the related spec change that calls that out:

https://github.com/w3c/ServiceWorker/issues/595

Its possible we were trying to match a WPT test or something.

wanderview:
Or maybe we removed it because we also remove the fragment from the scope and that spec change was about trying to compare scriptURL against the scope for validity.

jungkees:
> Its possible we were trying to match a WPT test or something.

Likely!

> https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1251/90c04cd...c3455b0.html#clear-registration-algorithm step 5 – we don't need the exclude fragment flag anymore right?

Yes, I changed it to not set the flag.

> https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1251/90c04cd...c3455b0.html#get-registration-algorithm step 3 – we don't need the exclude fragment flag, but maybe this one is less obvious. Perhaps move it to a note?

Having thought about it, I think Get Registration and Set Registration should set the flag as they take a url as an argument. Currently they are invoked only from SW internal algorithms, but if they were invoked from outside, that flag would be needed.

I intentionally left Match Service Worker Registration step 2 to not set the flag as the given url's fragment doesn't have an effect on matching the scope.

PR #1277, opened 2018-02-09
Improve.ready and Activate

### LINKED ISSUE #1011: .ready promise should be resolved after updating registration.state properties
The steps for resolving the .ready promise contain the following (section 3.4.2):
> If registration’s active worker is null, wait until registration’s active worker changes.
>> Implementers should consider this condition is met when the corresponding registration request >> gets to the step 6 of Activate algorithm.

However, both step 5 and step 6 from Activate algorithm use queues to make changes to the ServiceWorkerRegistration objects and dispatching statechange events.

This may lead to situations where the ready promise is resolved before registration.active is updated.

jungkees:
Before this change, .ready method was waiting for the change of the
state of the regisration's active worker without being synchronized with
the steps that set the corresponding ServiceWorkerRegistration object's
active ServiceWorker object's state.

This changes both .ready and Activate such that .ready tries to resolve
the ready promise, if pending, and just returns. And Activate resolves
pending ready promises after updating active ServiceWorker object's
state.

Fixes #1011

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1277.html" title="Last updated on Feb 12, 2018, 6:12 AM GMT (77eb154)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1277/49f954f...77eb154.html" title="Last updated on Feb 12, 2018, 6:12 AM GMT (77eb154)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -597,9 +597,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       <dfn attribute for="ServiceWorkerContainer"><code>ready</code></dfn> attribute *must* run these steps:
 
         1. Let |readyPromise| be the [=context object=]'s [=ServiceWorkerContainer/ready promise=].
-        1. If |readyPromise| is pending, then:
+        1. If |readyPromise| is pending, run the following substeps [=in parallel=]:
             1. Let |registration| be the result of running [=Match Service Worker Registration=] with the [=context object=]'s [=ServiceWorkerContainer/service worker client=]'s [=creation URL=].
-            1. If |registration| is not null, and |registration|'s [=active worker=] is not null, resolve |readyPromise| with the {{ServiceWorkerRegistration}} object that represents |registration|.
+            1. If |registration| is not null, and |registration|'s [=active worker=] is not null, [=queue a task=] on |readyPromise|'s [=relevant settings object=]'s [=responsible event loop=], using the [=DOM manipulation task source=], to resolve |readyPromise| with the {{ServiceWorkerRegistration}} object that represents |registration| in |readyPromise|'s [=relevant Realm=].
         1. Return |readyPromise|.
 
       Note: The returned [=ServiceWorkerContainer/ready promise=] will never reject. If it does not resolve in this algorithm, it will eventually resolve when a matching [=/service worker registration=] is registered and its [=active worker=] is set. (See the relevant [Activate algorithm step](#activate-resolve-ready-step).)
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -596,22 +596,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-                1. Jump to the step labeled *CheckRegistration*.
-            1. If |registration|'s <a>active worker</a> is null, wait until |registration|'s <a>active worker</a> changes.
-
-                Note: Implementers should consider this condition is met when the corresponding registration request gets to the step 6 of <a>Activate</a> algorithm.
-
-            1. Resolve <a>context object</a>'s [=ServiceWorkerContainer/ready promise=] with the {{ServiceWorkerRegistration}} object which represents |registration|.
-        1. Return <a>context object</a>'s [=ServiceWorkerContainer/ready promise=].
+        1. Let |readyPromise| be the [=context object=]'s [=ServiceWorkerContainer/ready promise=].
+        1. If |readyPromise| is pending, then:
+            1. Let |registration| be the result of running [=Match Service Worker Registration=] with the [=context object=]'s [=ServiceWorkerContainer/service worker client=]'s [=creation URL=].
```
jakearchibald: I don't think we should be running this on the main thread since it's a disk read.

docs/index.bs
```diff
@@ -2541,7 +2532,11 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
           Note: Once an active worker is activating, neither a runtime script error nor a force termination of the active worker prevents the active worker from getting activated.
 
-      1. For each [=/service worker client=] |client| whose <a>creation URL</a> <a lt="Match Service Worker Registration">matches</a> |registration|'s [=service worker registration/scope url=]:
+      1. Let |matchedClients| be a [=list=] of [=/service worker clients=] whose <a>creation URL</a> <a lt="Match Service Worker Registration">matches</a> |registration|'s [=service worker registration/scope url=].
+      1. [=list/For each=] |client| of |matchedClients|:
+          1. Let |readyPromise| be |client|'s [=environment settings object/global object=]'s {{ServiceWorkerContainer}} object's [=ServiceWorkerContainer/ready promise=].
```
jakearchibald: I think we need to queue a task on each client to do this.

docs/index.bs
jungkees: Right. I made it run in parallel and queue a task to check out the ready promise.

docs/index.bs
jungkees: In the subsequent step, I queued a task to resolve the ready promise. But I think your point is we should queue a task to get the ready promise itself for each client in this step, right? I made a change to address that point. Please take a look.

Issue #1212, opened 2017-10-20
The classic script's source text isn't defined in HTML any more


jungkees:
[Update](https://w3c.github.io/ServiceWorker/#update-algorithm) algorithm step 8 compares the updating script's source text against the installed script's source text. This check is to return early when the two scripts are byte identical.

I noticed HTML removed the definition of the source text from the classic script struct and added a record (Script Record) to the script struct instead.

I wonder if comparing the record.[[ECMAScriptCode]] of the two scripts would make sense or we should add the source text to the script struct again. From https://github.com/w3c/ServiceWorker/pull/1023#discussion_r92696321, the latter seems to be more appropriate I suppose.

/cc @domenic

domenic:
Oh, sorry about this. It does sound like we need to add the source text back to the script record. I will try to add it back soon, per the comment you cited.

jungkees:
That sounds good. I presume you meant to add it back to the script struct, right? We need non-parsed raw script texts for both classic scripts and module scripts.

domenic:
Yep.

jungkees:
https://github.com/w3c/ServiceWorker/commit/8e25c26efc932960a1449cf65de36dada371c867 closes this issue by comparing response bodies instead of source texts.

PR #1283, opened 2018-03-05
Improve service worker script caching and update

### LINKED ISSUE #1041: Handling duplicate importScripts()
What should happen when registering a service worker that does:
`importScripts('script.js'); importScripts('script.js');`

In Chrome's implementation, the response for the first importScripts() is stored (so that if SW completes installing, that is the version that is persisted), and the second one goes to network ignoring the first one and is not persisted.

That seems reasonable but as part of another bug fix, I'm thinking of changing this so that the second importScripts() returns the first stored version.

Also what should happen when a service worker recursively importScripts() itself? In Chrome's implementation, this causes a JS stack overflow, and each call goes to network, but only the first one is attempted to be stored. I think I'll simply respond with a network error in this case since handling it is wasteful and futile.

Thoughts?

jungkees:
This change includes/considers the following:
 - Include imported scripts to byte-check (for classic scripts).
 - Compare responses' body instead of source text as per https://github.com/whatwg/html/issues/3316.
 - Handle duplicate importScripts() as per https://github.com/w3c/ServiceWorker/issues/1041.
 - Replace *imported scripts updated flag* referenced in importScripts()
   by using service worker's state item.
 - Have Update's perform the fetch steps cover module scripts.
 - Avoid dobule-download of imported scripts pointed out in https://github.com/w3c/ServiceWorker/pull/1023#discussion_r92201798.

This change basically makes it check out if the main script resource is
identical to the existing resource. If so, it returns; otherwise, it
creates a new service worker and evalute it to check out if any imported
scripts are changed. It continues with Install only when any of the
resources has been changed. With the change, importScripts() returns
resources from the cache for any duplicated requests including the
request for the main script.

Fixes #1041, #1212, #1023.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1283.html" title="Last updated on Mar 13, 2018, 6:51 AM GMT (c34985f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1283/b695958...c34985f.html" title="Last updated on Mar 13, 2018, 6:51 AM GMT (c34985f)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -165,7 +165,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker=] has an associated <dfn export id="dfn-skip-waiting-flag">skip waiting flag</dfn>. Unless stated otherwise it is unset.
 
-    A [=/service worker=] has an associated <dfn export id="dfn-imported-scripts-updated-flag">imported scripts updated flag</dfn>. It is initially unset.
+    A [=/service worker=] has an associated <dfn export id="dfn-classic-scripts-imported-flag">classic scripts imported flag</dfn>. It is initially unset.
+
+    A [=/service worker=] has an associated <dfn export id="dfn-include-updated-resources-flag">include updated resources flag</dfn>. It is initially unset.
 
     A [=/service worker=] has an associated <dfn export id="dfn-set-of-event-types-to-handle">set of event types to handle</dfn> (a [=ordered set|set=]) whose [=list/item=] is an <a>event listener</a>'s event type. It is initially an empty set.
 
@@ -2027,22 +2029,28 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       When the <dfn method for="ServiceWorkerGlobalScope" id="importscripts-method"><code>importScripts(|urls|)</code></dfn> method is called on a {{ServiceWorkerGlobalScope}} object, the user agent *must* <a>import scripts into worker global scope</a>, given this {{ServiceWorkerGlobalScope}} object and |urls|, and with the following steps to [=fetching scripts/perform the fetch=] given the [=/request=] |request|:
 
         1. Let |serviceWorker| be |request|'s [=request/client=]'s [=environment settings object/global object=]'s [=ServiceWorkerGlobalScope/service worker=].
-        1. If |serviceWorker|'s <a>imported scripts updated flag</a> is unset, then:
-            1. Let |registration| be |serviceWorker|'s [=containing service worker registration=].
-            1. Set |request|'s [=service-workers mode=] to "`none`".
-            1. Set |request|'s [=request/cache mode=] to "<code>no-cache</code>" if any of the following are true:
-                * |registration|'s [=service worker registration/update via cache mode=] is "`none`".
-                * The [=current global object=]'s [=force bypass cache for importscripts flag=] is set.
-                * |registration|'s [=last update check time=] is not null and the time difference in seconds calculated by the current time minus |registration|’s [=last update check time=] is greater than 86400.
-            1. Let |response| be the result of <a lt="fetch">fetching</a> |request|.
-            1. If |response|’s <a for="response" href="https://github.com/whatwg/fetch/issues/376">cache state</a> is not "<code>local</code>", set |registration|’s [=service worker registration/last update check time=] to the current time.
-            1. [=Extract a MIME type=] from the |response|'s [=unsafe response=]'s [=response/header list=]. If this MIME type (ignoring parameters) is not a [=JavaScript MIME type=], return a [=network error=].
-            1. If |response|'s <a>unsafe response</a>'s [=response/type=] is not "<code>error</code>", and |response|'s [=response/status=] is an <a>ok status</a>, then:
-                1. [=map/Set=] <a>script resource map</a>[|request|'s [=request/url=]] to |response|.
-            1. Return |response|.
-        1. Else:
-            1. If <a>script resource map</a>[|url|] [=map/exists=], return <a>script resource map</a>[|url|].
-            1. Else, return a <a>network error</a>.
+        1. If |serviceWorker|'s [=service worker/script resource map=][|request|'s [=request/url=]] [=map/exists=], return [=service worker/script resource map=][|request|'s [=request/url=]].
+        1. If |serviceWorker|'s [=state=] is *installed*, *activating*, *activated*, or *redundant*, return a [=network error=].
+        1. Let |registration| be |serviceWorker|'s [=containing service worker registration=].
+        1. Set |request|'s [=service-workers mode=] to "`none`".
+        1. Set |request|'s [=request/cache mode=] to "<code>no-cache</code>" if any of the following are true:
+            * |registration|'s [=service worker registration/update via cache mode=] is "`none`".
+            * The [=current global object=]'s [=force bypass cache for importscripts flag=] is set.
+            * |registration|'s [=last update check time=] is not null and the time difference in seconds calculated by the current time minus |registration|’s [=last update check time=] is greater than 86400.
+        1. Let |response| be the result of <a lt="fetch">fetching</a> |request|.
+        1. If |response|’s <a for="response" href="https://github.com/whatwg/fetch/issues/376">cache state</a> is not "<code>local</code>", set |registration|’s [=service worker registration/last update check time=] to the current time.
+        1. [=Extract a MIME type=] from the |response|'s [=unsafe response=]'s [=response/header list=]. If this MIME type (ignoring parameters) is not a [=JavaScript MIME type=], return a [=network error=].
+        1. If |response|'s <a>unsafe response</a>'s [=response/type=] is not "<code>error</code>", and |response|'s [=response/status=] is an <a>ok status</a>, then:
+            1. Let |newestWorker| be the result of running [=Get Newest Worker=] with |registration|.
+            1. Let |resource| be null.
+            1. If |newestWorker| is not null, set |resource| to |newestWorker|'s [=service worker/script resource map=][|request|'s [=request/url=]], or null if it does not [=map/exist=].
+            1. Set |serviceWorker|'s [=service worker/include updated resources flag=] if any of the following are true:
+                * |newestWorker| is null.
+                * |resource| is null.
+                * |resource|'s [=response/body=] is not byte-for-byte identical with |response|'s [=response/body=]. 
+            1. [=map/Set=] [=service worker/script resource map=][|request|'s [=request/url=]] to |response|.
+            1. Set |serviceWorker|'s [=classic scripts imported flag=].
+        1. Return |response|.
     </section>
   </section>
 
@@ -2069,7 +2077,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   <section>
     <h3 id="privacy">Privacy</h3>
 
-    [=/Service workers=] introduce new persistent storage features including <a>scope to registration map</a> (for [=/service worker registrations=] and their [=/service workers=]), [=request response list=] and <a>name to cache map</a> (for caches), and <a>script resource map</a> (for script resources). In order to protect users from any potential <a biblio data-biblio-type="informative" lt="unsanctioned-tracking">unsanctioned tracking</a> threat, these persistent storages *should* be cleared when users intend to clear them and *should* maintain and interoperate with existing user controls e.g. purging all existing persistent storages.
+    [=/Service workers=] introduce new persistent storage features including <a>scope to registration map</a> (for [=/service worker registrations=] and their [=/service workers=]), [=request response list=] and <a>name to cache map</a> (for caches), and [=service worker/script resource map=] (for script resources). In order to protect users from any potential <a biblio data-biblio-type="informative" lt="unsanctioned-tracking">unsanctioned tracking</a> threat, these persistent storages *should* be cleared when users intend to clear them and *should* maintain and interoperate with existing user controls e.g. purging all existing persistent storages.
   </section>
 </section>
 
@@ -2147,7 +2155,11 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A <a>job</a> has a <dfn id="dfn-job-worker-type">worker type</dfn> ("<code>classic</code>" or "<code>module</code>").
 
+    A [=job=] has a <dfn export id="dfn-job-script-resource-map">script resource map</dfn> which is an <a>ordered map</a> where the keys are [=/URLs=] and the values are [=/responses=].
+
     A <a>job</a> has an <dfn id="dfn-job-update-via-cache-mode">update via cache mode</dfn>, which is "`imports`", "`all`", or "`none`".
+
+    A [=job=] has a <dfn id="dfn-job-potentially-include-updated-resources-flag">potentially include updated resources flag</dfn>. It is initially unset.
 
     A <a>job</a> has a <dfn id="dfn-job-client">client</dfn> (a [=/service worker client=]). It is initially null.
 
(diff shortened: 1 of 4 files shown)


jungkees:
@jakearchibald, this obsoletes #1023 and covers #1041 and #1212. Please take a look.

@domenic, please review if the [perform the fetch](https://github.com/w3c/ServiceWorker/pull/1283/files#diff-27b79860afe28f01aed4f1f6228367faR2433) steps correctly covers module scripts, and also if comparing reponse's body would be correct.

/cc @matto @wanderview @aliams @cdumez

domenic:
The comparison of response body looks perfect. That is much simpler than the work I was planning in https://github.com/whatwg/html/issues/3316, and I think in a good way. And I believe since you updated the "perform the fetch" in both places (importScripts() and fetch a module worker script graph), it should all work great for module scripts too.

Thanks for doing this despite my continual slowness in helping support. I think your version is better than mine anyway, so it all worked out :)

jungkees:
@domenic, thanks for reviewing and having helped out to solve this. I'll work on the rest with other SW reviewers.

docs/index.bs
```diff
@@ -2027,22 +2029,28 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+        1. Let |response| be the result of <a lt="fetch">fetching</a> |request|.
+        1. If |response|’s <a for="response" href="https://github.com/whatwg/fetch/issues/376">cache state</a> is not "<code>local</code>", set |registration|’s [=service worker registration/last update check time=] to the current time.
+        1. [=Extract a MIME type=] from the |response|'s [=unsafe response=]'s [=response/header list=]. If this MIME type (ignoring parameters) is not a [=JavaScript MIME type=], return a [=network error=].
+        1. If |response|'s <a>unsafe response</a>'s [=response/type=] is not "<code>error</code>", and |response|'s [=response/status=] is an <a>ok status</a>, then:
+            1. Let |newestWorker| be the result of running [=Get Newest Worker=] with |registration|.
+            1. Let |resource| be null.
+            1. If |newestWorker| is not null, set |resource| to |newestWorker|'s [=service worker/script resource map=][|request|'s [=request/url=]], or null if it does not [=map/exist=].
```
mattto: I'm a bit confused by this, it makes it look like importScripts() in one service worker (`serviceWorker`) is affecting the state of another service worker (`newestWorker`), which I thought #1041 decided was wrong.

docs/index.bs
```diff
@@ -2027,22 +2029,28 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+            1. Set |serviceWorker|'s [=service worker/include updated resources flag=] if any of the following are true:
+                * |newestWorker| is null.
+                * |resource| is null.
+                * |resource|'s [=response/body=] is not byte-for-byte identical with |response|'s [=response/body=]. 
+            1. [=map/Set=] [=service worker/script resource map=][|request|'s [=request/url=]] to |response|.
+            1. Set |serviceWorker|'s [=classic scripts imported flag=].
```
mattto: Generally, I'm a bit confused what about the meaning of `include updated resources flag` and `classic scripts imported flag`. Could you help me understand what they are doing?

docs/index.bs
jungkees: This is a case where `serviceWorker` doesn't have any stored response and hasn't gone to the installed state. (Cases other than that return early in the step 2 and 3.)

My intention here is to compare the response retrieved from the network (or HTTP cache depending on the cache mode) with the registration's newest worker's stored response, if any. I think we'd want to make it continue with Install only when importScripts() for `serviceWorker` has byte-differnce against `newestWorker`'s stored resources. The service worker's *include updated resources flag* is used to mark this condition that is later checked in Update.

mattto: I'm understanding more.

I thought when we discussed this a while back (https://github.com/w3c/ServiceWorker/issues/839#issuecomment-266373359), we decided the "update" service worker shouldn't run at all if the main script and imported scripts of the newest installed worker haven't changed. I.e., the update check should first fetch the main script and importScripts of the newest worker, if those haven't changed you halt the update. If there is a change you launch the new service worker. It'd be wasteful to spawn a worker just to do the importScripts then discover there was no change, and then terminate the worker before installation. But maybe we should talk about this one of the GitHub issues.

docs/index.bs
```diff
@@ -2027,22 +2029,28 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+        1. [=Extract a MIME type=] from the |response|'s [=unsafe response=]'s [=response/header list=]. If this MIME type (ignoring parameters) is not a [=JavaScript MIME type=], return a [=network error=].
+        1. If |response|'s <a>unsafe response</a>'s [=response/type=] is not "<code>error</code>", and |response|'s [=response/status=] is an <a>ok status</a>, then:
+            1. Let |newestWorker| be the result of running [=Get Newest Worker=] with |registration|.
+            1. Let |resource| be null.
+            1. If |newestWorker| is not null, set |resource| to |newestWorker|'s [=service worker/script resource map=][|request|'s [=request/url=]], or null if it does not [=map/exist=].
+            1. Set |serviceWorker|'s [=service worker/include updated resources flag=] if any of the following are true:
+                * |newestWorker| is null.
+                * |resource| is null.
+                * |resource|'s [=response/body=] is not byte-for-byte identical with |response|'s [=response/body=]. 
+            1. [=map/Set=] [=service worker/script resource map=][|request|'s [=request/url=]] to |response|.
```
mattto: This should be "Set *serviceWorker's* script resource map", right?

docs/index.bs
```diff
@@ -2027,22 +2029,28 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+            * The [=current global object=]'s [=force bypass cache for importscripts flag=] is set.
+            * |registration|'s [=last update check time=] is not null and the time difference in seconds calculated by the current time minus |registration|’s [=last update check time=] is greater than 86400.
+        1. Let |response| be the result of <a lt="fetch">fetching</a> |request|.
+        1. If |response|’s <a for="response" href="https://github.com/whatwg/fetch/issues/376">cache state</a> is not "<code>local</code>", set |registration|’s [=service worker registration/last update check time=] to the current time.
+        1. [=Extract a MIME type=] from the |response|'s [=unsafe response=]'s [=response/header list=]. If this MIME type (ignoring parameters) is not a [=JavaScript MIME type=], return a [=network error=].
+        1. If |response|'s <a>unsafe response</a>'s [=response/type=] is not "<code>error</code>", and |response|'s [=response/status=] is an <a>ok status</a>, then:
+            1. Let |newestWorker| be the result of running [=Get Newest Worker=] with |registration|.
```
mattto: Wouldn't newestWorker just be serviceWorker, it's the new installing worker?

docs/index.bs
mattto: I'm understanding more.

I thought when we discussed this a while back (https://github.com/w3c/ServiceWorker/issues/839#issuecomment-266373359), we decided the "update" service worker shouldn't run at all if the main script and imported scripts of the newest installed worker haven't changed. I.e., the update check should first fetch the main script and importScripts of the newest worker, if those haven't changed you halt the update. If there is a change you launch the new service worker. It'd be wasteful to spawn a worker just to do the importScripts then discover there was no change, and then terminate the worker before installation. But maybe we should talk about this one of the GitHub issues.

docs/index.bs
jungkees: When the `importScripts()` is called during the evaluation in Update, *serviceWorker* ins't in the installing state yet. *newestWorker* here would be one of the workers of the registration. So, we're comparing the new pre-installing worker's resources with the current newest worker's resources here.

docs/index.bs
jungkees: Oh I see.. We had already discussed this a while back ;). But I was worried about the comment in https://github.com/w3c/ServiceWorker/pull/1023/files#r92201798, too. And what should we do if any of the imported scripts were updated, later start a worker with them in the cache, and they contain errors. We couldn't avoid starting a work in this case anyway.

But I'm open to this option or any other smarter ways to improve performance.

docs/index.bs
jungkees: Yes. Fixed.

docs/index.bs
mattto: Ah I see, I think I saw the Run Service Worker algorithm being called in Install and thought that was the first time we are running the worker.

docs/index.bs
mattto: We should be able to spec it in a way that doesn't do a double download...

Currently in the common case a navigation most likely will not cause a separate service worker to run for the update check. This spec change will change it so that most likely one will, so the browser will often start two service workers for a navigation. That would really impact memory/performance for us.

I wouldn't worry too much about the uncommon error case causing an extra service worker to be started.

jungkees: That's a fair point. I think we can also avoid a double download by populating the cache before the first execution of the main script.

docs/index.bs
```diff
@@ -2027,22 +2029,28 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-            1. [=Extract a MIME type=] from the |response|'s [=unsafe response=]'s [=response/header list=]. If this MIME type (ignoring parameters) is not a [=JavaScript MIME type=], return a [=network error=].
-            1. If |response|'s <a>unsafe response</a>'s [=response/type=] is not "<code>error</code>", and |response|'s [=response/status=] is an <a>ok status</a>, then:
-                1. [=map/Set=] <a>script resource map</a>[|request|'s [=request/url=]] to |response|.
-            1. Return |response|.
-        1. Else:
-            1. If <a>script resource map</a>[|url|] [=map/exists=], return <a>script resource map</a>[|url|].
-            1. Else, return a <a>network error</a>.
+        1. If |serviceWorker|'s [=service worker/script resource map=][|request|'s [=request/url=]] [=map/exists=], return [=service worker/script resource map=][|request|'s [=request/url=]].
+        1. If |serviceWorker|'s [=state=] is *installed*, *activating*, *activated*, or *redundant*, return a [=network error=].
```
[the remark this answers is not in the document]
jungkees: Yes. Addressed.

docs/index.bs
```diff
@@ -165,7 +165,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker=] has an associated <dfn export id="dfn-skip-waiting-flag">skip waiting flag</dfn>. Unless stated otherwise it is unset.
 
-    A [=/service worker=] has an associated <dfn export id="dfn-imported-scripts-updated-flag">imported scripts updated flag</dfn>. It is initially unset.
+    A [=/service worker=] has an associated <dfn export id="dfn-classic-scripts-imported-flag">classic scripts imported flag</dfn>. It is initially unset.
+
+    A [=/service worker=] has an associated <dfn export id="dfn-include-updated-resources-flag">include updated resources flag</dfn>. It is initially unset.
```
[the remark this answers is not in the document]
jungkees: I removed this flag as part of the new snapshot.

docs/index.bs
```diff
@@ -2147,8 +2155,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A <a>job</a> has a <dfn id="dfn-job-worker-type">worker type</dfn> ("<code>classic</code>" or "<code>module</code>").
 
+    A [=job=] has a <dfn export id="dfn-job-script-resource-map">script resource map</dfn> which is an <a>ordered map</a> where the keys are [=/URLs=] and the values are [=/responses=].
```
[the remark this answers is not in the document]
jungkees: Good point. I made it a variable.

docs/index.bs
```diff
@@ -2147,8 +2155,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
     A <a>job</a> has an <dfn id="dfn-job-update-via-cache-mode">update via cache mode</dfn>, which is "`imports`", "`all`", or "`none`".
 
+    A [=job=] has a <dfn id="dfn-job-potentially-include-updated-resources-flag">potentially include updated resources flag</dfn>. It is initially unset.
```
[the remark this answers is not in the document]
jungkees: Right. I made it a variable.

jungkees:
@mattto, @jakearchibald, I uploaded a new snapshot. Please take a look.

jakearchibald:
👍Thanks for doing this.

The job stuff is making things increasingly difficult to read IMO. I wonder if [parallel queues](https://html.spec.whatwg.org/multipage/infrastructure.html#starting-a-new-parallel-queue) would let us achieve the same without a state object being passed around.

docs/index.bs
```diff
@@ -2027,22 +2027,22 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+            * |registration|'s [=last update check time=] is not null and the time difference in seconds calculated by the current time minus |registration|’s [=last update check time=] is greater than 86400.
+        1. Let |response| be the result of [=fetch|fetching=] |request|.
+        1. Set |response| to |response|'s [=unsafe response=].
+        1. If |response|’s <a for="response" href="https://github.com/whatwg/fetch/issues/376">cache state</a> is not "`local`", set |registration|’s [=service worker registration/last update check time=] to the current time.
+        1. [=Extract a MIME type=] from the |response|'s [=response/header list=]. If this MIME type (ignoring parameters) is not a [=JavaScript MIME type=], return a [=network error=].
+        1. If |response|'s [=response/type=] is not "`error`", and |response|'s [=response/status=] is an <a>ok status</a>, then:
+            1. [=map/Set=] |serviceWorker|'s [=service worker/script resource map=][|request|'s [=request/url=]] to |response|.
+            1. Set |serviceWorker|'s [=classic scripts imported flag=].
+        1. Return |response|.
```
mattto: To aid my understanding, what exactly happens in the error response case? Does it result in an error so the installation/update fails, or does it mean we basically treat it like the importScripts() call never happened?

jungkees: An error response makes it throw a "`NetworkError`" `DOMException` in [fetch a classic worker-imported script](https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-classic-worker-imported-script) step 4. That again makes Update reject the job promise in https://github.com/w3c/ServiceWorker/pull/1283/files#diff-27b79860afe28f01aed4f1f6228367faR2467. This particular behavior hasn't been changed.

jungkees:
@jakearchibald,

> The job stuff is making things increasingly difficult to read IMO. I wonder if parallel queues would let us achieve the same without a state object being passed around.

I think we should figure out to what extent we could refactor it. I tried some bits to make the related algorithm instances serialized using parallel queues here: https://github.com/w3c/ServiceWorker/pull/1229. But it wasn't that I intended to replace the whole job queue structure with the parallel queue. What concerns me about changing it now is the implementations have already implemented the job queues as in the spec. Let's take a look what we can do.

jungkees:
BTW, I think we need some additional tests for this change. Does anyone have bandwidth to help with writing tests?

docs/index.bs
mattto: Ah I see. That means if you do try { importScripts('fail.js'); } and catch the exception, you can later try to do importScripts('fail.js') and it'll retry again. I think that matches our implementation too.

(Acknowledged that this hasn't changed, just confirming my understanding.)

jungkees: Yes, if you catch the exception, it just installs it without populating the cache. When the service worker starts up later, `importScript('fail.js')` will get a network error again that throws and be caught again.

jungkees:
Thanks for reviewing. I'll update the commit description when merging the final snapshot.

I also checked out the tests and found they already cover the behavior this PR addresses:
 - https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/import-scripts-resource-map.https.html is testing the multiple same script imports case.
 - https://github.com/w3c/web-platform-tests/blob/master/service-workers/service-worker/update-bytecheck.https.html is testing no new service worker's created when both of the main script and the imported scripts are byte-for-byte identical. I think we might want to add more assertions to check other cases, e.g. whether it continues with Install if only imported scripts were changed. But I'll merge this PR as we at least have the test file in place.

jungkees:
I'll merge this PR. @ylafon is helping out with the travis.yml settings for later works.

PR #1293, opened 2018-03-27
Make Client.postMessage to unloaded client not throw

### LINKED ISSUE #1291: Client.postMessage() throwing for unloaded client is problematic
Currently `Client.postMessage()` says:

```
2. Let destination be the ServiceWorkerContainer object whose
service worker client is the context object’s service worker client,
or null if no match is found.

3. If destination is null, throw an "InvalidStateError" DOMException.
```

This is problematic for a few reasons:

1. Most `postMessage()` implementations don't throw.  They just silently consume the message if the other end is no longer listening.  For example Worker.postMessage() and MessageChannel, etc.
2. The lookup of the destination is presented as synchronous in the spec, but it must be async in any cross-process browser implementation.

If we wanted we could change postMessage() to return a promise, but its probably better to just silently eat the message.

I'm filing this issue in regards to https://github.com/w3c/web-platform-tests/pull/10153

jungkees:
https://github.com/w3c/ServiceWorker/commit/cf834f052bc0aed0795fbf29cbf1c6f251dd5c9d
clarified the target object can be null, and it must throw in that case.

But https://github.com/w3c/ServiceWorker/issues/1291 pointed out that we
cannot implement that behavior without blocking the service worker
process in multi-process browser architectures.

This change makes the control return right away if the target client has
been unloaded.

Fixes #1291.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1293.html" title="Last updated on May 24, 2019, 11:53 PM UTC (b23637c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1293/b3a1328...b23637c.html" title="Last updated on May 24, 2019, 11:53 PM UTC (b23637c)">Diff</a>


jungkees:
@wanderview, does https://github.com/w3c/ServiceWorker/pull/1293/files#diff-27b79860afe28f01aed4f1f6228367faR1033 work for you? Chromium checks if the client is available when the request is getting through the browser process: https://cs.chromium.org/chromium/src/content/browser/service_worker/service_worker_version.cc?type=cs&q=ServiceWorkerVersion::PostMessageToClient&l=1155.

/cc @mattto

docs/index.bs
```diff
@@ -1028,10 +1028,11 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       The <dfn method for="Client"><code>postMessage(|message|, |transfer|)</code></dfn> method *must* run these steps:
 
         1. Let |sourceSettings| be the <a>context object</a>'s <a>relevant settings object</a>.
-        1. Let |destination| be the {{ServiceWorkerContainer}} object whose [=ServiceWorkerContainer/service worker client=] is the <a>context object</a>'s [=Client/service worker client=], or null if no match is found.
-        1. If |destination| is null, <a>throw</a> an "{{InvalidStateError}}" {{DOMException}}.
         1. Let |serializeWithTransferResult| be <a abstract-op>StructuredSerializeWithTransfer</a>(|message|, |transfer|). Rethrow any exceptions.
-        1. Add a <a>task</a> that runs the following steps to |destination|'s [=ServiceWorkerContainer/client message queue=]:
+        1. Let |targetClient| be the [=context object=]'s [=Client/service worker client=], or null if it does not exist (i.e. if the [=/service worker client=] has been [=Handle Service Worker Client Unload|unloaded=]).
```
jakearchibald: I'm not sure what "or null if it does not exist" means. Have we defined how to check for existence? If so we should link to that. Also, is it ok to perform this existence check on the main thread?

jungkees: @jakearchibald, I pushed a commit to address these points. PTAL.

PR #1294, opened 2018-03-28
Configure script's settings and realm for service workers

### LINKED ISSUE #1013: Sort out settings object for service worker module scripts

jungkees:
Service workers have a different script execution model from other
workers. Update algorithm fetches a service worker script, and Run
Service Worker algorithm runs the script when installing it and whenever
functional events and message events need to be dispatched.

This change:
 - Passes null as the value of script settings object/module map
   settings object argument to fetch a classic worker script/fetch a
   module worker script graph algorithm, respectively, called in Update
   algorithm, instead of the placeholder argument "the to-be-created
   environment settings object".
 - Sets the script's settings to the environment settings object created
   in Run Service Worker algorithm and the script's record.[[Realm]] to
   that settings object's Realm.

Fixes #1013.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2375,10 +2375,10 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |updatedResourceMap| be an [=ordered map=] where the [=map/keys=] are [=/URLs=] and the [=map/values=] are [=/responses=].
       1. Switching on |job|'s <a>worker type</a>, run these substeps with the following options:
           <!-- TODO: reorganize algorithm so that the worker environment is created before fetching happens -->
-          : "<code>classic</code>"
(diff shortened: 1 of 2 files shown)


jungkees:
@domenic, passing null to those algorithms breaks the current _create a classic script_ and _create a module script_ algorithm steps. I'm not sure if I still need to create an environment settings object and pass it although the script's settings will be replaced by the another settings object in Run Service Worker eventually.

As far as I can tell, script's settings and script's record.[[Realm]] are referenced when the script is evaluated. So, even if we don't set them in the fetching phase, it wouldn't make any side effects, right?

PR #1315, opened 2018-05-29
Clients.get: block on reserved clients.

mattto:
@jakearchibald  @jungkees Can you review this? The index.html looks huge because it includes the spec recompile commit.

***
<a href="https://pr-preview.s3.amazonaws.com/mattto/ServiceWorker/pull/1315.html" title="Last updated on Jun 7, 2018, 3:46 AM GMT (08ad67e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1315/9a6b738...mattto:08ad67e.html" title="Last updated on Jun 7, 2018, 3:46 AM GMT (08ad67e)">Diff</a>


docs/index.bs
```diff
@@ -1148,6 +1148,11 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. Run these substeps <a>in parallel</a>:
             1. For each [=/service worker client=] |client| whose [=service worker client/origin=] is the <a lt="same origin">same</a> as the associated [=ServiceWorkerGlobalScope/service worker=]'s [=environment settings object/origin=]:
                 1. If |client|'s [=environment/id=] is not |id|, continue to the next iteration of the loop.
+                1. If |client|'s [=environment/execution ready flag=] is unset, then:
+                    1. Wait for the [=/HTTP fetch=] steps using the [=/request=] that |client| is the associated [=request/reserved client=] of to finish.
```
jakearchibald: This seems a bit hand-wavey, since there isn't an explicit connection between a client & it's fetch.

Adding this connection would likely be really hard, since fetch doesn't really return an instance (this has been a problem [elsewhere](https://github.com/whatwg/fetch/issues/536#issuecomment-330482027)).

Feels like it would be easier to turn the *execution ready flag* into an enum, eg *execution state*, which could be "pending", "ready" or "failed". Then, this part of the spec would be explicit, as you'd be waiting for that enum to change to "failed" (in which case you'd reject and abort) or "ready" (in which case you'd continue).

jakearchibald: Ohh, if we have a standard way of describing "goes away", I agree that would fit nicely.

wanderview: Its probably not clearly defined for windows in the html spec.  I think defining when to set this error state would be equivalent to defining discarding the client on failure, though.

wanderview: Note, we also already depend on client discarding in other places in the service worker spec.  We activate waiting workers when the last client "goes away".  If that last controlled client is a reserved client being navigated, then we have this same situation.

Not saying we shouldn't spec it better, but just highlighting that we already depend on this concept pretty heavily.

mattto: Yep that's all fair. I indeed wanted to wait for "client goes away" and tried to sweep it under the rug by waiting for HTTP fetch to finish.

That's a good point about activation. In the SW spec this happens on "Handle Service Worker Client Unload" which is run when a "service worker client unloads by unloading or terminating". I'm guessing reserved clients don't go through "unload" as per the HTML spec though so I can't just hook into that yet.

wanderview: Would it not fall under "terminating"?

Also consider that reserved clients can "go away" even when an error is not encountered.  When you hit a cross-origin redirect the reserved client is discarded and a new one is created.

mattto: Oh, it probably means to link to worker termination in the HTML spec.

mattto: I think probably here "process a navigate response" <https://html.spec.whatwg.org/multipage/browsing-the-web.html#process-a-navigate-response> we can add something like "discard reservedEnvironment" and have the service worker spec wait for that.

wanderview: To clarify, the reserved client is not discarded on all redirects.  Redirecting cross-origin, though, must result in a new client being reserved with the new origin to avoid violating same-origin restrictions.  So at that point the old reserved client can be discarded.

mattto: @wanderview Hum, that's not what I thought. I filed https://github.com/w3c/ServiceWorker/issues/1316

FYI: created https://github.com/whatwg/html/pull/3723

mattto:
I'm having trouble rebasing. I will make a new PR.

wanderview:
Did we decide that we wanted to reject if the reserved client is discarded?  The original PR has "NetworkError" there, but not sure that is quite right.

We could also just resolve `undefined` like `clients.get()` does when it doesn't find a matching client.

mattto:
Closing this in favor of https://github.com/w3c/ServiceWorker/pull/1323 (sorry about my git troubles)

PR #1259, opened 2018-01-12
Clean up Client API to not expose reserved clients

jungkees:
To reflect the decision to not expose reserved clients
(https://github.com/w3c/ServiceWorker/issues/1216) and to spec
.resultingClientId and .targetClientId in Nightly only, this:

* Removes
  - Client.reserved and its associated reserved state (V1, Nightly)
  - ClientQueryOptions.includeReserved (V1, Nightly)
  - FetchEvent.reservedClientId (V1)
  - FetchEvent.targetClientId (V1)
  - FetchEventInit.reservedClientId (V1)
  - FetchEventInit.targetClientId (V1)

* Changes
  - FetchEvent.reservedClientId to FetchEvent.resultingClientId (Nightly)
  - FetchEventInit.reservedClientId to FetchEventInit.resultingClientId
   (Nightly)
  - Handle Fetch to set FetchEvent.clientId for a navigation request to
   the empty string (V1)

* Corrects
  - matchedClients with clientObjects in Clients.matchAll() (V1, Nightly)

Related issue: https://github.com/w3c/ServiceWorker/issues/1245.

This also cleans up sort condition steps in Clients.matchAll() that
fixes https://github.com/w3c/ServiceWorker/issues/1080 (V1, Nightly)

(Changes for the Clients interface's methods will be addressed as
separate PRs.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1259.html" title="Last updated on Jan 12, 2018, 11:14 AM GMT (5ab28e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1259/8db81d8...5ab28e8.html" title="Last updated on Jan 12, 2018, 11:14 AM GMT (5ab28e8)">Diff</a>

Issue #1316, opened 2018-05-30
Is resultingClientId preserved over redirects?


mattto:
Splitting off of https://github.com/w3c/ServiceWorker/pull/1315

On my reading of the spec, I thought the reserved client gets set one time at the beginning of a navigation, and it never changes until navigation finishes.

In the Chrome we were just going to not give reserved clients URLs (or rather, we'd essentially ignore whatever URL it currently has until navigation settles). But we would preserve resultingClientId over cross-origin redirects. I believe a team here wanted this at least for same-origin redirects but probably don't care about cross-origin. Is there any reason we'd prefer to preserve or discard the id for cross-origin ones?

annevk:
Currently https://html.spec.whatwg.org/#process-a-navigate-fetch ends up resetting the reserved client for each redirect.

When and where is the reserved client exposed? It seems bad if different origins can get hold of the same unique identifier, so if you're proposing changing the above setup to not reset it for each redirect, we should make sure that's not possible.

mattto:
Reserved clients are never exposed. Clients.get(id) resolves once the navigation finishes or rejects once the navigation failed. It would also reject if the resulting client is cross-origin to the caller. So I was thinking just having the same unique identifier would be safe.

In #process-a-navigate-fetch, I thought the relevant part (navigation follows a redirect) is step 7:
"If response has a location URL and it is either failure or a URL whose scheme is an HTTP(S) scheme, then set response to the result of performing HTTP-redirect fetch using request and response and then run this step again."

I read that as running step 7 again, i.e., step 5 only happens once.

annevk:
You're reading that correctly, I was mistaken. That means the URL never gets updated though. It also means the CSP check only gets applied once. That's almost certainly a bug.

wanderview:
I believe we agreed at a previous f2f to reset the resulting client when we perform a cross-origin redirect.  Because clients are tied to their origin, something like this must happen.  Otherwise we risk exposing the same client ID to two different origin service workers.

For example, a fetch event for origin A, then a redirect to origin B, and then a fetch event on origin B.  The origin A and origin B should see different client IDs, IMO.

I'll look later today to try to find where we discussed that before.

wanderview:
Please see https://github.com/w3c/ServiceWorker/issues/1031#issuecomment-291056646.

wanderview:
[...]

PR #1323, opened 2018-06-07
Clients.get: block on reserved clients.

mattto:
Clients.get: block on reserved clients.

Follow-up to 07e9487 (#1259).

***
<a href="https://pr-preview.s3.amazonaws.com/mattto/ServiceWorker/pull/1323.html" title="Last updated on Jun 13, 2018, 3:31 PM GMT (fdd9c51)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1323/70e76c5...mattto:fdd9c51.html" title="Last updated on Jun 13, 2018, 3:31 PM GMT (fdd9c51)">Diff</a>


mattto:
@jakearchibald @wanderview @jungkees  please take a look (I failed to link to the newly added steps in the HTML spec, but maybe need to just update Bikeshed or wait a little longer).

annevk:
Did you explicitly add `data-export=""` to the definitions in HTML you wanted to use here?

docs/index.bs
```diff
@@ -1147,28 +1154,11 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-                        1. Set |focusState| to the result of running the <a>has focus steps</a> with |browsingContext|'s <a>active document</a> as the argument.
-                        1. If |client| is a <a>window client</a>, set |ancestorOriginsList| to |browsingContext|'s <a>active document</a>'s <a>relevant global object</a>'s {{Location}} object's [=Location/ancestor origins list=]'s associated list.
-                    1. Wait for |task| to have executed.
-                    1. Let |windowClient| be the result of running <a>Create Window Client</a> algorithm with |client|, |visibilityState|, |focusState|, and |ancestorOriginsList| as the arguments.
-                    1. Resolve |promise| with |windowClient| and abort these steps.
+                1. If |client|'s [=environment/id=] is not |id|, [=continue=].
+                1. Wait for either |client|'s [=environment/execution ready flag=] to be set or for |client|'s [=discarded flag=] to be set.
+                1. If |client|'s [=environment/execution ready flag=] is set, then invoke [=Resolve Get Client Promise=] with |client| and |promise|.
+                1. Else, resolve |promise| with undefined.
```
wanderview: nit: Technically I think you could leave this "resolve with undefined" out as the one following the loop would take effect.  What you have might be clearer about intent, though.

docs/index.bs
```diff
@@ -3216,6 +3206,40 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+          1. Let |visibilityState| be null.
+          1. Let |focusState| be false.
+          1. Let |ancestorOriginsList| be the empty list.
+          1. If |client| is an [=environment settings object=], set |browsingContext| to |client|'s [=environment settings object/global object=]'s [=/browsing context=].
+          1. Else, set |browsingContext| to |client|’s [=environment/target browsing context=].
+          1. <a>Queue a task</a> |task| to run the following substeps on |browsingContext|'s <a>event loop</a> using the <a>user interaction task source</a>:
+              1. Set |visibilityState| to |browsingContext|'s <a>active document</a>'s {{Document/visibilityState}} attribute value.
+              1. Set |focusState| to the result of running the <a>has focus steps</a> with |browsingContext|'s <a>active document</a> as the argument.
+              1. If |client| is a <a>window client</a>, set |ancestorOriginsList| to |browsingContext|'s <a>active document</a>'s <a>relevant global object</a>'s {{Location}} object's [=Location/ancestor origins list=]'s associated list.
+          1. Wait for |task| to have executed.
```
wanderview: I realize this part of the PR is mostly preserving the original algorithm, but should we check to see if the client was discarded or destroyed between running the task and getting the result back?  Or otherwise say that if any error occurs here we resolve undefined, etc?

docs/index.bs
mattto: Ah actually I think I have this wrong: the break will break out of the "for each" loop only and we'll double resolve the promise after the loop. I'll adjust this to "abort these steps" which I think breaks out of the  "Run these substeps".

docs/index.bs
mattto: Ohhh that's a great point. Yes I'll add that.

mattto:
@annevk Updating Bikeshed fixed it!

mattto:
Thanks all, I've updated the PR.

jungkees:
@mattto, thanks for the commits! I think they look good. Waiting on either the execution ready flag set or the discarded flag set sounds good to me.

I just want to clarify who will invoke setting the client's discarded flag step. And we may need to change some related WTP tests for this change, right?

mattto:
@jungkees Thanks for the review! As for WPT tests, this is being actively developed in Chrome so we'll have tests soon (I find it'd be hard to write WPT tests without a working implementation to test the test, but I see others have been able to do so.)

The discarded flag is set by HTML: https://html.spec.whatwg.org/multipage/webappapis.html#environment-discarding-steps

docs/index.bs
```diff
@@ -222,6 +222,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A <dfn export id="dfn-service-worker-client" for="">service worker client</dfn> is an [=environment=].
 
+    A [=/service worker client=] has an associated <dfn export>discarded flag</dfn>. It is initially unset.
+
+   This specification defines the following [=environment discarding steps=] for a [=/service worker client=] |client|:
+        1. Set |client|'s [=discarded flag=].
+
+    Note: Implementations can essentially discard clients whose [=discarded flag=] is set.
```
jakearchibald: Nit: lose "essentially". Also, is this defined somewhere?

docs/index.bs
```diff
@@ -3216,6 +3204,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+      : Input
+      :: |client|, a [=/service worker client=]
+      :: |promise|, a <a>promise</a>
+
+      : Output
+      :: none
+
+      1. If |client| is an [=environment settings object=], then:
+          1. If |client| is not a <a>secure context</a>, reject |promise| with a "{{SecurityError}}" {{DOMException}} and abort these steps.
```
jakearchibald: For consistency, can we go with `[=secure context=]`?

docs/index.bs
```diff
@@ -3216,6 +3204,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      1. Else:
+          1. If |client|’s <a>creation URL</a> is not a <a>potentially trustworthy URL</a>, reject |promise| with a "{{SecurityError}}" {{DOMException}} and abort these steps.
```
jakearchibald: Same here, can we use `[=term=]` instead of `<a>term</a>`?

docs/index.bs
```diff
@@ -3216,6 +3204,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      1. If |client| is an [=environment settings object=] and is not a [=window client=], then:
+          1. Let |clientObject| be the result of running <a>Create Client</a> algorithm with |client| as the argument.
```
jakearchibald: `[=Create Client=]`

jakearchibald: There are a few more in this algo too.

docs/index.bs
mattto: Removed essentially. Does "this" mean  [=environment discarding steps=]? They are defined in HTML: https://html.spec.whatwg.org/multipage/webappapis.html#environment-discarding-steps

docs/index.bs
mattto: Yep! This whole block was just copied from the original. Done.

docs/index.bs
mattto: All done.

docs/index.bs
```diff
@@ -222,6 +222,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A <dfn export id="dfn-service-worker-client" for="">service worker client</dfn> is an [=environment=].
 
+    A [=/service worker client=] has an associated <dfn export>discarded flag</dfn>. It is initially unset.
+
+   This specification defines the following [=environment discarding steps=] for a [=/service worker client=] |client|:
```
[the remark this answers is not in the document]

mattto:
Thanks @jakearchibald , I've updated the PR.

docs/index.bs
```diff
@@ -222,6 +222,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A <dfn export id="dfn-service-worker-client" for="">service worker client</dfn> is an [=environment=].
 
+    A [=/service worker client=] has an associated <dfn export>discarded flag</dfn>. It is initially unset.
+
+   Each [=/service worker client=] has the following [=environment discarding steps=]:
+        1. Set |client|'s [=discarded flag=].
+
+    Note: Implementations can discard clients whose [=discarded flag=] is set.
```
jakearchibald: "Implementations can discard clients" sounds normative, but maybe it's defined somewhere else? Otherwise, add it to the definition of the flag. Eg:

```
A [=/service worker client=] has an associated <dfn export>discarded flag</dfn>. It is initially unset. User agents may discard clients whose [=discarded flag=] is set.
```

(it looks like we tend to say "user agents" rather than "implementations")

docs/index.bs
```diff
@@ -3216,6 +3204,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+          1. Else, set |browsingContext| to |client|’s [=environment/target browsing context=].
+          1. [=Queue a task=] |task| to run the following substeps on |browsingContext|'s [=event loop=] using the [=user interaction task source=]:
+              1. Set |visibilityState| to |browsingContext|'s [=active document=]'s {{Document/visibilityState}} attribute value.
+              1. Set |focusState| to the result of running the [=has focus steps=] with |browsingContext|'s [=active document=] as the argument.
+              1. If |client| is a [=window client=], set |ancestorOriginsList| to |browsingContext|'s [=active document=]'s [=relevant global object=]'s {{Location}} object's [=Location/ancestor origins list=]'s associated list.
+          1. Wait for |task| to have executed.
+          1. If |client|'s [=discarded flag=] is set, resolve |promise| with undefined and abort these steps.
+          1. Let |windowClient| be the result of running [=Create Window Client=] algorithm with |client|, |visibilityState|, |focusState|, and |ancestorOriginsList| as the arguments.
+          1. Resolve |promise| with |windowClient|.
```
jakearchibald: This might be a silly question, but is it possible for a client to end up representing another origin, or is the client discarded and a new one created in that case?

docs/index.bs
mattto: Ah no, it was just supposed to be a hint for implementors. It just says "don't worry about this flag. you don't have to keep state around for a client with this flag set". I can just remove the non-normative note if it's confusing.

docs/index.bs
mattto: That's a great question! See https://github.com/w3c/ServiceWorker/issues/1316, it turns out the HTML and/or Fetch spec has a bug around clients persisting across cross-origin redirects. I was planning on tackling that next.

In the case where the client is discarded, it's expected to never reach execution ready so will never be exposed.

docs/index.bs
jakearchibald: Ohhh I get it now. Yeah that's fine then.

docs/index.bs
jakearchibald: Gotcha. In that case, LGTM.

docs/index.bs
jungkees: @mattto, so what do you expect the implementations actually do when the discarded flag is set? That bit sounds like a normative to me. Let me know if closing/terminating of the clients is rather implicit so we don't have to specify. 

(And now we auto-bikeshed the .bs file in Nightly. So you can just upload .bs without .html next time. For this PR, don't worry about it.)

mattto: @jungkees I think implementations would resolve any open clients.get(id) promise with undefined, then completely forget about that id. If clients.get(id) were called again, it would resolve with undefined immediately because the id wouldn't be found. So they don't need to keep state of the client around. The flag is just to make the spec algorithm work.

Sounds like I should just remove the note.

jungkees: Let's keep the note. I'll merge the PR. Maybe you can update the note later with what you explained if that'll make it clearer.

jungkees:
Thanks @mattto!

mattto:
Thanks!

PR #1327, opened 2018-06-19
Resolve the skipWaiting() promise after activation completes.

### LINKED ISSUE #1187: skipWaiting() promise should resolve after promotion to .active
As spec'd the skipWaiting() promise doesn't seem really useful. I think it might be an oversight.

I'd expect the promise to resolve after the worker has been promoted to .active and has state 'activating'.

But currently the spec is:
2. Invoke Try Activate with service worker's containing service worker registration.
3. Resolve promise with undefined.

And Try Activate early returns in some cases:
- existing active worker is still activating
- the existing active worker has pending events

So if Try Activate early returned, we resolve the promise before activating.

Also am I parsing this sentence correctly: "The result of running Service Worker Has No Pending Events with registration’s active worker is true, and no service worker client is using registration or registration’s waiting worker's skip waiting flag is set." means "A && (B || C)" where A = "no pending events", B = "no client" and C = "skip waiting flag".

mattto:
The Activate algorithm's final step was to run Update Worker State
to make the worker `activated`. However Update Worker State
just queues a task to do so. So the task would run after
skipWaiting() resolved its promise, which wasn't the intent of
the spec.

Fixes #1187.


mattto:
@jungkees @wanderview @jakearchibald please take a look

mattto:
WPT change at https://github.com/web-platform-tests/wpt/pull/11566

TravisGit:
Pull request

PR #1329, opened 2018-06-24
Add Client.frameType back

jungkees:
This change adds Client.frameType back with the reasons:
 - Implementations maintain it.
 - It can be accessed outside of fetch event handlers

This change also improved the steps of the algorithms that Create Window Client
invokes by removing unnecessary parallel steps.

Fixes: https://github.com/w3c/ServiceWorker/issues/732.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1329.html" title="Last updated on Jun 26, 2018, 5:35 AM GMT (3c7ff5c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1329/ff36205...3c7ff5c.html" title="Last updated on Jun 26, 2018, 5:35 AM GMT (3c7ff5c)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1087,11 +1087,11 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. Let |serviceWorkerEventLoop| be the [=current global object=]'s [=event loop=].
         1. Let |promise| be a new <a>promise</a>.
         1. [=Queue a task=] to run the following steps on the [=context object=]'s associated [=Client/service worker client=]'s [=responsible event loop=] using the [=user interaction task source=]:
-            1. Run the [=focusing steps=] with the current [=/browsing context=].
-            1. Let |frameType| be the result of running [=Get Frame Type=] with the current [=/browsing context=].
-            1. Let |visibilityState| be the current [=/browsing context=]'s [=active document=]'s {{Document/visibilityState}} attribute value.
-            1. Let |focusState| be the result of running the [=has focus steps=] with the current [=/browsing context=]'s [=active document=].
-            1. Let |ancestorOriginsList| be the current [=/browsing context=]'s [=active document=]'s [=relevant global object=]'s {{Location}} object's [=Location/ancestor origins list=]'s associated list.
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -1068,54 +1084,45 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-                1. Set |focusState| to the result of running the <a>has focus steps</a> with |browsingContext|'s <a>active document</a> as the argument.
-                1. Set |ancestorOriginsList| to |browsingContext|'s <a>active document</a>'s <a>relevant global object</a>'s {{Location}} object's [=Location/ancestor origins list=]'s associated list.
-            1. Wait for |task| to have executed.
-            1. Let |windowClient| be the result of running <a>Create Window Client</a> algorithm with the <a>context object</a>'s associated [=Client/service worker client=], |visibilityState|, |focusState|, and |ancestorOriginsList| as the arguments.
-            1. If |windowClient|'s <a>focus state</a> is true, resolve |promise| with |windowClient|.
-            1. Else, reject |promise| with a <code>TypeError</code>.
+        1. [=Queue a task=] to run the following steps on the [=context object=]'s associated [=Client/service worker client=]'s [=responsible event loop=] using the [=user interaction task source=]:
+            1. Run the [=focusing steps=] with the current [=/browsing context=].
```
mattto: Should "this be the [=context object=]'s associated [=/browsing context=]", for clarity and consistency with line 1089? (I would think current browsing context refers to the service worker's browsing context, if that exists)

jungkees: Thanks for reviewing it! I used the *current* as a browsing context has an event loop, and the queued task is running in the browsing context's event loop. https://html.spec.whatwg.org/#navigating-auxiliary-browsing-contexts-in-the-dom used the *current* as well.

But I agree the use of the *current* here can be ambiguous as we don't have a definition for the *current* browsing context. And we have a clearer definition, the [=WindowClient/browsing context=]. I'll change them using the [=WindowClient/browsing context=].

jungkees: Done.

Issue #1245, opened 2017-12-11
Spec FetchEvent.resultingClientId/targetClientId


jungkees:
We decided to [not expose reserved client](https://github.com/w3c/ServiceWorker/issues/1216) that closed #1228 #1215 #1123 #1078 #1035 #1034 #1031 #1030.

V1 removes the related web exposed features:
- `Client.reserved`
- `ClientQueryOptions.includeReserved`
- `FetchEvent.reservedClientId`
- `FetchEvent.targetClientId`
- `FetchEventInit.reservedClientId`
- `FetchEventInit.targetClientId`

V2 will spec `FetchEvent.resultingClientId`, `FetchEvent.targetClientId`, and the underlying behavior including the excusion ready flag.

monica-ch:
[Bug triage] Issue still exists

Issue #1091, opened 2017-03-23
consider Client behavior for windows where initial about:blank is replaced with a loaded document


wanderview:
Our current plan is to expose the environment creation URL as `Client.url` when the environment is marked execution ready.  This is a bit confusing for the case where the environment has multiple documents; like when an initial about:blank is replaced with a loaded document.  See:

https://github.com/whatwg/html/issues/2456

My read of the spec and my initial implementation so far expose the about:blank URL briefly as `Client.url`.  In addition, the environment is marked execution ready twice.  Once as about:blank and once as the final loaded URL.  The second time its marked execution ready this basically amounts to updating the creation URL.

Is this intended behavior for the clients API?  I suspect perhaps the intent was not to mark the Client as non-reserved until after the final loaded document is execution ready.

wanderview:
>  I suspect perhaps the intent was not to mark the Client as non-reserved until after the final loaded document is execution ready.

Talking with @bzbarsky, it may be very hard to determine if an initial about:blank will indeed be followed up another document.  And I don't think we can just ignore about:blank completely because programmatically constructed iframes are supposed to inherit their parent's service worker controller, etc.  We need to represent them as Client objects.

jakearchibald:
In this iframe case:

* client: the about:blank doc
* reserved client: null
* target client: null

wanderview:
@jakearchibald, I'm kind of confused by https://github.com/w3c/ServiceWorker/issues/1091#issuecomment-291048685.  I don't recall this being the outcome from the f2f meeting.

I thought we said we were going to do:

* client: the client initiating the load just like we do for non-replacement cases
* reserved client: the client currently attached to the about:blank window
* target client: I don't remember, but the about:blank window would make sense to me.

If you don't want the existing about:blank to be "reserved" since its already execution ready, we could also do:

* client: the client initiating the load just like we do for non-replacement cases
* reserved client: null
* target client: I don't remember, but the about:blank window would make sense to me.

So in:

```
// foo.html
let f = document.createElement('iframe');
document.body.appendChild(f);

f.contentWindow.someGlobalStatus = 'yuck';

f.src = 'frame.html';
```

I guess I would always expect `FetchEvent.clientId` to reference the foo.html parent window here.

It also seems to make sense to make `FetchEvent.targetClientId` reference the current about:blank iframe client.

[...]

PR #1333, opened 2018-07-01
Change name from targetClientId to replacesClientId

jungkees:
We had decided (and I forgot) to change the name of FetchEvent.targetClientId to
.replacesClientId to clarify the meaning that this client is a to be replaced
client: https://github.com/w3c/ServiceWorker/issues/1091#issuecomment-342569083.

Related issue: https://github.com/w3c/ServiceWorker/issues/1245.

Fetch PR: https://github.com/whatwg/fetch/pull/774.
HTML PR: https://github.com/whatwg/html/pull/3788.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1333.html" title="Last updated on Jul 1, 2018, 11:25 PM GMT (b6bb29b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1333/ff36205...b6bb29b.html" title="Last updated on Jul 1, 2018, 11:25 PM GMT (b6bb29b)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1343,7 +1343,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         readonly attribute Promise&lt;any&gt; preloadResponse;
         readonly attribute DOMString clientId;
         readonly attribute DOMString resultingClientId;
-        readonly attribute DOMString targetClientId;
+        readonly attribute DOMString replacesClientId;
 
         void respondWith(Promise&lt;Response&gt; r);
       };
@@ -1354,7 +1354,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         Promise&lt;any&gt; preloadResponse;
         DOMString clientId = "";
         DOMString resultingClientId = "";
-        DOMString targetClientId = "";
+        DOMString replacesClientId = "";
       };
     </pre>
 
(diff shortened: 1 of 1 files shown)


jungkees:
I'll make PRs for Fetch and HTML changes.

annevk:
This can be reviewed and landed (if it passes review). Fetch and HTML are updated.

annevk:
(It might take a day for Bikeshed to have the new terms indexed though.)

jungkees:
Thanks @annevk. I'll merge this tomorrow.

foolip:
I'm coming across this issue via BCD: https://github.com/mdn/browser-compat-data/pull/10717

It looks like this property shipped in Safari in April 2018, and then was renamed after that. I can't find any WebKit bug filed. This looks like an unfortunate mishap, but does someone who was involved with this know more, and can someone file that WebKit bug? Historical tests in WPT would be great too.

jakearchibald:
Yeah this isn't great. If memory serves, we were under the impression that no one had shipped this. I guess there was some sort of miscommunication.

I'll create the tests and file the WebKit issue. Thanks for pointing this out.

foolip:
Thanks @jakearchibald!

I think that this sort of thing could be avoided with a faster and more reliable flow of support data from specs and releases into BCD. If there had already been a green row in https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent#browser_compatibility already, something like this might have occurred:
- Nobody checked MDN, so the change was made
- The spec link in the BCD entry disappeared from [webref](https://github.com/w3c/webref) so this was flagged
- Someone reviewing bad spec links in BCD finds their way back here within a week or two and points it out
- The relevant implementers are looped in, maybe the change is reverted or maybe it sticks, but *something* happens.

jakearchibald:
https://github.com/web-platform-tests/wpt/pull/29213

foolip:
If my research in https://github.com/mdn/browser-compat-data/pull/10774 is correct, no browser has actually shipped `replacesClientId` yet. Given that, is it worth considering just renaming it back to `targetClientId`?

PR #1335, opened 2018-07-12
Fix syntax error in IDL

lukebjerring:
`AbstractWorker` is an `interface mixin`, not an `interface`, so needs to be pull in via `includes` not `implements`.

Found while authoring https://github.com/web-platform-tests/wpt/pull/11960


jungkees:
Thanks @lukebjerring!

PR #1340, opened 2018-08-08
Make Run Service Worker return after initial script evaluation.

mattto:
Otherwise Run Service Worker would not return until the worker
thread was terminated.

Also fix formatting so steps for 17 are numbered 17.1 and 17.2
instead of 17.1 and 17.1.

***
<a href="https://pr-preview.s3.amazonaws.com/mattto/ServiceWorker/pull/1340.html" title="Last updated on Aug 8, 2018, 3:37 AM GMT (a138ce9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1340/8e6e33b...mattto:a138ce9.html" title="Last updated on Aug 8, 2018, 3:37 AM GMT (a138ce9)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2675,11 +2675,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. For each |eventType| of |settingsObject|'s [=environment settings object/global object=]'s associated list of <a>event listeners</a>' event types:
               1. [=set/Append=] |eventType| to |workerGlobalScope|'s associated [=ServiceWorkerGlobalScope/service worker=]'s <a>set of event types to handle</a>.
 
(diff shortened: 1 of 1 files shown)


mattto:
@jakearchibald @jungkees please take a look.

jungkees:
@mattto, Run Service Worker step 4 creates a parallel execution environment and makes the rest of the steps in that parallel thread. This is modeled after https://html.spec.whatwg.org/#run-a-worker.

> Also fix formatting so steps for 17 are numbered 17.1 and 17.2 instead of 17.1 and 17.1.

Thanks for spotting this. It seems you need one more space char before "Note" in the change. Could you check in on it?

mattto:
Thanks! I missed that. My motivation was to have Run Service Worker return the result of running the script, since the callsite expects it. I'll make a new PR for that.

PR #1341, opened 2018-08-10
Abstracting cache matching

jakearchibald:
We have a somewhat similar `matchAll` method in background fetch https://wicg.github.io/background-fetch/#background-fetch-registration and I'd like to use the same matching we use in service worker.

I've abstracted out the part that can determine true/false for a request and a stored request/response. In background fetch we may not have the response yet, so it's optional.

If you're happy with this PR, I'll port it to v1.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1341.html" title="Last updated on Sep 12, 2018, 10:11 AM GMT (da43529)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1341/59c46fb...da43529.html" title="Last updated on Sep 12, 2018, 10:11 AM GMT (da43529)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1671,7 +1671,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         boolean ignoreSearch = false;
         boolean ignoreMethod = false;
         boolean ignoreVary = false;
-        DOMString cacheName;
       };
     </pre>
 
@@ -1912,11 +1911,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     <pre class="idl">
       [SecureContext, Exposed=(Window,Worker)]
       interface CacheStorage {
-        [NewObject] Promise&lt;any&gt; match(RequestInfo request, optional CacheQueryOptions options);
+        [NewObject] Promise&lt;any&gt; match(RequestInfo request, optional MultiCacheQueryOptions options);
         [NewObject] Promise&lt;boolean&gt; has(DOMString cacheName);
         [NewObject] Promise&lt;Cache&gt; open(DOMString cacheName);
         [NewObject] Promise&lt;boolean&gt; delete(DOMString cacheName);
         [NewObject] Promise&lt;sequence&lt;DOMString&gt;&gt; keys();
+      };
+
+      dictionary MultiCacheQueryOptions : CacheQueryOptions {
+        DOMString cacheName;
       };
     </pre>
     <!--FIXME(jungkees): set method is not entirely dropped yet. Promise&lt;any&gt; set(DOMString key, Cache val);-->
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -1656,20 +1656,23 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     <pre class="idl">
       [SecureContext, Exposed=(Window,Worker)]
       interface Cache {
-        [NewObject] Promise&lt;any&gt; match(RequestInfo request, optional CacheQueryOptions options);
-        [NewObject] Promise&lt;FrozenArray&lt;Response&gt;&gt; matchAll(optional RequestInfo request, optional CacheQueryOptions options);
+        [NewObject] Promise&lt;any&gt; match(RequestInfo request, optional MultiCacheQueryOptions options);
```
wanderview: Can you explain why methods on the Cache object would take a `cacheName` value via the `MultiCacheQueryOptions`?  It seems like the Cache object should be locked to itself and not be able to query other cache objects.

wanderview: Oh, I guess we always allowed `cacheName` at the webidl and would internally throw in those cases.

I guess I find it confusing to have a separate dictionary name with a more restrictive set of options, but not use it here.

jakearchibald: Agreed. I didn't want to change existing interfaces, but I guess this is backwards compatible. Will fix.

jungkees: Yes, this would be backwards compatible.

jungkees:
LGTM. I'm happy that you're going to port it to V1. Glad we can re-use the part of the algorithm by the abstraction indeed.

jakearchibald:
Addressed comments & ported to v1.

jakearchibald:
@jungkees any objections to this landing?

docs/index.bs
```diff
@@ -3194,37 +3197,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-          1. Let |fieldValues| be the [=list=] containing the elements corresponding to the [=http/field-values=] of the <a>Vary</a> header.
-          1. Let |isMatched| be true.
-          1. For each |fieldValue| in |fieldValues|:
-              1. If |fieldValue| matches "`*`", or the [=combined value=] given |fieldValue| and |cachedRequest|'s [=request/header list=] does not match the [=combined value=] given |fieldValue| and |request|'s [=request/header list=], then:
-                  1. Set |isMatched| to false.
-                  1. [=Break=].
-          1. If |isMatched| is true, [=list/append=] |cachedRequest|/|cachedResponse| to |resultList|.
+          1. Let |cachedRequest| be |requestResponse|'s request.
+          1. Let |cachedResponse| be |requestResponse|'s response.
+          1. If [=Request Matches Cached Item=] with |requestQuery|, |cachedRequest|, |cachedResponse|, and |options| returns true, then add |cachedRequest|/|cachedResponse| to |resultList|.
```
jungkees: It seems this step lost semantics of using the "copy" of the request and the "copy" of the response?

docs/v1/index.bs
```diff
@@ -3040,37 +3043,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-          1. Let |fieldValues| be the [=list=] containing the elements corresponding to the [=http/field-values=] of the <a>Vary</a> header.
-          1. Let |isMatched| be true.
-          1. For each |fieldValue| in |fieldValues|:
-              1. If |fieldValue| matches "`*`", or the [=combined value=] given |fieldValue| and |cachedRequest|'s [=request/header list=] does not match the [=combined value=] given |fieldValue| and |request|'s [=request/header list=], then:
-                  1. Set |isMatched| to false.
-                  1. [=Break=].
-          1. If |isMatched| is true, [=list/append=] |cachedRequest|/|cachedResponse| to |resultList|.
+          1. Let |cachedRequest| be |requestResponse|'s request.
+          1. Let |cachedResponse| be |requestResponse|'s response.
+          1. If [=Request Matches Cached Item=] with |requestQuery|, |cachedRequest|, |cachedResponse|, and |options| returns true, then add |cachedRequest|/|cachedResponse| to |resultList|.
```
jungkees: It seems this step lost semantics of using the "copy" of the request and the "copy" of the response?

jungkees:
@jakearchibald, please check my additional comments.

jakearchibald:
@jungkees good catch! I've added copying back.

Issue #1374, opened 2018-12-10
Update shouldn't fail due to MIME check or 404 for the incumbent's worker scripts


mattto:
The update check spiders the script resource map of the incumbent worker scripts, but then fails immediately if one script has a non-OK status or fails the MIME check. This means if the incumbent worker is using import-script.js which now 404s, and the updated worker doesn't use import-script.js, the update will fail.

Instead these checks shouldn't fail the whole update, since it's not known whether the new worker will use them.

> For each url → storedResponse of newestWorker’s script resource map:
> ...
> 7. Extract a MIME type from the fetchedResponse’s header list. If this MIME type (ignoring parameters) is not a JavaScript MIME type, asynchronously complete these steps with a network error.
> 8. If fetchedResponse’s type is "error", or fetchedResponse’s status is not an ok status, asynchronously complete these steps with a network error.

mattto:
Proposal: When spidering |newestWorker|'s script resource map, always write the response into |updatedResourceMap|. On MIME failure, network error, or byte-for-byte body diff, set |hasUpdatedResources| to true. This ensures the diff translates into starting the installing worker, which is correct since the worker scripts are different. When the installing worker does importScripts() a MIME failure or network failure will throw SecurityError/NetworkError so in effect the installation will usually fail. We also need to remove unused resources from the installing worker's map that ended up not being used, or these will be used in the next update check against this new worker. So augment the script resource map with a flag about whether the script was loaded. After installation finishes, drop unused scripts from the map.

mattto:
Ah I think I found Chrome's (in-progress) implementation is doing the ""no change" and continue on to the next script" already: https://cs.chromium.org/chromium/src/content/browser/service_worker/service_worker_single_script_update_checker.cc?rcl=e2f59f49da06e49c476bd0d32913d7349d61a0d4&l=175

makotoshimazu:
Yeah, it's implemented as no change tentatively. [this](https://cs.chromium.org/chromium/src/content/browser/service_worker/service_worker_single_script_update_checker.cc?l=176) is the latest code, which treats network error differently from 200 and identical case, but it behaves similarly for now.
Treating network error on comparison as "no change" sounds reasonable to me too.

tingshao:
[...]

PR #1377, opened 2018-12-19
Don't fail update upon error fetching the incumbent worker's import s…

mattto:
…cripts.

To include imported scripts in the byte-for-byte comparison, the Update
algorithm fetches the incumbent worker's scripts in order to
detect if there is a difference. But it would abort the update upon
MIME type error of if fetching the script failed. This is undesirable
because the new worker might not use these scripts at all, e.g.,
the scripts may have legitimitely been removed from the server.

To fix this, just set the updated flag upon detecting a difference
and continue with the Update.

Addresses issue #1374.

***
<a href="https://pr-preview.s3.amazonaws.com/mattto/ServiceWorker/pull/1377.html" title="Last updated on Mar 5, 2019, 12:50 AM UTC (17e3d8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1377/c0aa494...mattto:17e3d8c.html" title="Last updated on Mar 5, 2019, 12:50 AM UTC (17e3d8c)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -161,17 +161,19 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A <a>script resource</a> has an associated <dfn export for="script resource" id="dfn-referrer-policy">referrer policy</dfn> (a [=/referrer policy=]). It is initially the empty string.
 
-    A [=/service worker=] has an associated <dfn export id="dfn-script-resource-map">script resource map</dfn> which is an <a>ordered map</a> where the keys are [=/URLs=] and the values are [=pairs=] consisting of a [=/response=] and a <dfn export id="dfn-script-resource-used-flag">used flag</dfn>.
+    A [=/service worker=] has an associated <dfn export id="dfn-script-resource-map">script resource map</dfn> which is an <a>ordered map</a> where the keys are [=/URLs=] and the values are [=/responses=].
 
-    Note: The [=used flag=] is only used to prune unused resources from a new worker's map after installation, that were populated based on the old worker's map during the update check.
+    A [=/service worker=] has an associated  <dfn export id="dfn-set-of-used-scripts">set of used scripts</dfn> (a [=ordered set|set=]) whose [=list/item=] is a [=/URL=]. It is initially empty.
+
+    Note: The [=set of used scripts=] is only used to prune unused resources from a new worker's map after installation, that were populated based on the old worker's map during the update check.
 
     A [=/service worker=] has an associated <dfn export id="dfn-skip-waiting-flag">skip waiting flag</dfn>. Unless stated otherwise it is unset.
 
     A [=/service worker=] has an associated <dfn export id="dfn-classic-scripts-imported-flag">classic scripts imported flag</dfn>. It is initially unset.
 
-    A [=/service worker=] has an associated <dfn export id="dfn-set-of-event-types-to-handle">set of event types to handle</dfn> (a [=ordered set|set=]) whose [=list/item=] is an <a>event listener</a>'s event type. It is initially an empty set.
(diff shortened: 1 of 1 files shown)


mattto:
@jakearchibald @jungkees please review.

docs/index.bs
```diff
@@ -2058,21 +2062,20 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
             * The [=current global object=]'s [=force bypass cache for importscripts flag=] is set.
             * |registration|'s [=last update check time=] is not null and the time difference in seconds calculated by the current time minus |registration|’s [=last update check time=] is greater than 86400.
         1. Let |response| be the result of [=fetch|fetching=] |request|.
-        1. Set |response| to |response|'s [=unsafe response=].
         1. If |response|’s [=response/cache state=] is not "`local`", set |registration|’s [=service worker registration/last update check time=] to the current time.
-        1. [=Extract a MIME type=] from the |response|'s [=response/header list=]. If this MIME type (ignoring parameters) is not a [=JavaScript MIME type=], return a [=network error=].
-        1. If |response|'s [=response/type=] is not "`error`", and |response|'s [=response/status=] is an <a>ok status</a>, then:
-            1. [=map/Set=] |serviceWorker|'s [=script resource map=][|request|'s [=request/url=]] to |response|.
-            1. Set |serviceWorker|'s [=classic scripts imported flag=].
```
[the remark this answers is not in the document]
wanderview: I'm not sure this matches the behavior implemented in firefox.  Gecko treats errors importScripts as "no change" in the update algorithm:

https://searchfox.org/mozilla-central/rev/c21d6620d384dfb13ede6054015da05a6353b899/dom/serviceworkers/ServiceWorkerScriptCache.cpp#725-729

There is some discussion in the bug about why this was chosen here:

https://bugzilla.mozilla.org/show_bug.cgi?id=1290951#c16

@asutherland, what do you think about this PR?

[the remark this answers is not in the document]
mattto: Thanks for flagging, I'll study https://bugzilla.mozilla.org/show_bug.cgi?id=1290951. I was also considering cases like `try { importScripts('import.js'); }` which might not be a good justification.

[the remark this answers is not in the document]
mattto: Hmm.. let me return this discussion to the bug. Will follow up there.

mattto:
Hi @jakearchibald I've revised this PR per #1374. Please take another look.

docs/index.bs
```diff
@@ -163,13 +163,17 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker=] has an associated <dfn export id="dfn-script-resource-map">script resource map</dfn> which is an <a>ordered map</a> where the keys are [=/URLs=] and the values are [=/responses=].
 
+    A [=/service worker=] has an associated  <dfn export id="dfn-set-of-used-scripts">set of used scripts</dfn> (a [=ordered set|set=]) whose [=list/item=] is a [=/URL=]. It is initially empty.
```
jakearchibald: I think "It is initially a new [=ordered set|set=]" is the correct language here (and below).

mattto: Thank you! I see the infra spec examples has instances of "Let result be an empty list." or "Let tokens be a list of strings, initially empty.". But I see it also has "Let list be a new list." OK... I guess we'll do "new set"... seems wordy though.

jakearchibald: Yeah, it makes more sense when something can be an ordered set or null. Eg:

> A thing has an associated **whatever** (a set or null). It is initially empty.

Since it can be a set or null, it isn't totally clear that "initially empty" means a new empty set.

mattto: OK I see.

Issue #1382, opened 2019-01-21
Cache's put() says it clones the response


annevk:
I thought the pattern is that the developer needs to clone the response?

jakearchibald:
That was my understanding too.

annevk:
Step 7 and 8 of https://w3c.github.io/ServiceWorker/#cache-put are confusing therefore, but perhaps I'm missing something?

jakearchibald:
I think step 8 is trying to mark the original request "read", but it looks like a developer could still have a reference to the original body before it's replaced in 8.2.

annevk:
Aaah, that's what it's trying to do. That's indeed wrong. Simply reading the body of the passed in response (and not doing any cloning) would also fix @ytausky's #1383.

monica-ch:
[Bug triage] Fixed by https://github.com/w3c/ServiceWorker/pull/1384

PR #1384, opened 2019-01-22
Correcting the use of types and simplifying the response body read.

### LINKED ISSUE #1383: Cache.put() does not read response's body
A response's body is never explicitly read, so it's unclear what would be stored in the cache for a response with an arbitrary `ReadableStream`.

jakearchibald:
This is to fix #1383 and #1382.

If it's ok, I'll update v1 too.

cc @ytausky.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1384.html" title="Last updated on Jan 30, 2019, 11:36 PM UTC (b462a94)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1384/f0bd826...b462a94.html" title="Last updated on Jan 30, 2019, 11:36 PM UTC (b462a94)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1843,13 +1843,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. Let |realm| be the [=context object=]'s [=relevant realm=].
         1. [=Upon fulfillment of=] |bodyReadPromise|:
             1. Let |cacheJobPromise| be [=a new promise=].
-            1. Run the following substeps [=in parallel=]:
+            1. Return |bodyReadPromise| and run the following substeps [=in parallel=]:
                 1. Let |errorData| be null.
                 1. Invoke [=Batch Cache Operations=] with |operations|. If this [=throws=] an exception, set |errorData| to the exception.
                 1. [=Queue a task=], on |cacheJobPromise|'s [=relevant settings object=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to perform the following substeps:
                     1. If |errorData| is null, resolve |cacheJobPromise| with undefined.
                     1. Else, reject |cacheJobPromise| with a [=exception/create|new=] [=exception=] with |errorData| and a user agent-defined [=exception/message=], in |realm|.
-                1. Return |cacheJobPromise|.
         1. Return |bodyReadPromise|.
     </section>
 
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1842,7 +1842,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. Let |realm| be the [=context object=]'s [=relevant realm=].
         1. Return the result of the [=Upon fulfillment|fulfillment=] of |bodyReadPromise|:
             1. Let |cacheJobPromise| be [=a new promise=].
-            1. Return |bodyReadPromise| and run the following substeps [=in parallel=]:
+            1. Return |cacheJobPromise| and run these steps [=in parallel=]:
                 1. Let |errorData| be null.
                 1. Invoke [=Batch Cache Operations=] with |operations|. If this [=throws=] an exception, set |errorData| to the exception.
                 1. [=Queue a task=], on |cacheJobPromise|'s [=relevant settings object=]'s [=responsible event loop=] using the [=DOM manipulation task source=], to perform the following substeps:
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1832,7 +1832,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
             1. Let |reader| be the result of [=ReadableStream/get a reader|getting a reader=] for |stream|.
             1. Set |bodyReadPromise| to the result of [=Read all bytes|reading all bytes=] from |stream| with |reader|.
 
-            Note: This ensures that |response|'s [=response/body=] is [=Body/locked=], and we have a full buffered copy of the body in |clonedResponse|. An implementation could optimise by streaming directly to disk rather than memory.
+            Note: This ensures that |response|'s [=response/body=] is [=Body/locked=], and we have a full buffered copy of the body in |clonedResponse|. An implementation could optimize by streaming directly to disk rather than memory.
 
         1. Let |operations| be an empty [=list=].
         1. Let |operation| be a [=cache batch operation=].
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1832,7 +1832,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
             1. Let |reader| be the result of [=ReadableStream/get a reader|getting a reader=] for |stream|.
             1. Set |bodyReadPromise| to the result of [=Read all bytes|reading all bytes=] from |stream| with |reader|.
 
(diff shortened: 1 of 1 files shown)


docs/index.bs
```diff
@@ -1809,42 +1809,48 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-            1. Set |r| to |request|'s [=Request/request=].
-            1. If |r|'s [=request/url=]'s [=url/scheme=] is not one of "`http`" and "`https`", or |r|'s [=request/method=] is not \`<code>GET</code>\`, return [=a promise rejected with=] a `TypeError`.
-        1. Else if |request| is a string, then:
-            1. Set |r| to the associated [=Request/request=] of the result of invoking the initial value of {{Request}} as constructor with |request| as its argument. If this [=throws=] an exception, return [=a promise rejected with=] that exception.
-            1. If |r|'s [=request/url=]'s [=url/scheme=] is not one of "`http`" and "`https`", return [=a promise rejected with=] a `TypeError`.
-        1. If |response|'s associated [=Response/response=]'s [=response/status=] is `206`, return [=a promise rejected with=] a `TypeError`.
-        1. If |response|'s associated [=Response/response=]'s [=response/header list=] contains a <a>header</a> [=header/named=] \`<code>Vary</code>\`, then:
+      <dfn method for="Cache"><code>put(|requestArg|, |responseObject|)</code></dfn> method *must* run these steps:
+
+        1. Let |request| be null.
```
jakearchibald: I wanted to get rid of single char variable names, especially as |r| could be request/response/etc etc.

docs/index.bs
```diff
@@ -1809,42 +1809,48 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-        1. If |request| is a {{Request}} object, then:
-            1. Set |r| to |request|'s [=Request/request=].
-            1. If |r|'s [=request/url=]'s [=url/scheme=] is not one of "`http`" and "`https`", or |r|'s [=request/method=] is not \`<code>GET</code>\`, return [=a promise rejected with=] a `TypeError`.
-        1. Else if |request| is a string, then:
-            1. Set |r| to the associated [=Request/request=] of the result of invoking the initial value of {{Request}} as constructor with |request| as its argument. If this [=throws=] an exception, return [=a promise rejected with=] that exception.
-            1. If |r|'s [=request/url=]'s [=url/scheme=] is not one of "`http`" and "`https`", return [=a promise rejected with=] a `TypeError`.
-        1. If |response|'s associated [=Response/response=]'s [=response/status=] is `206`, return [=a promise rejected with=] a `TypeError`.
-        1. If |response|'s associated [=Response/response=]'s [=response/header list=] contains a <a>header</a> [=header/named=] \`<code>Vary</code>\`, then:
+      <dfn method for="Cache"><code>put(|requestArg|, |responseObject|)</code></dfn> method *must* run these steps:
```
jakearchibald: Using |requestArg| and |requestObject| so |request| and |response| can be the concepts rather than the object.

This cleared up a lot of misuse of types in this algorithm.

annevk: requestObject for consistency? (I've also never seen arg used so explicitly.)

docs/index.bs
```diff
@@ -1809,42 +1809,48 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-            1. [=Read all bytes=] from |dummyStream| with |reader|.
+        1. If |response|'s [=response/body=] is [=Body/disturbed=] or [=Body/locked=], return [=a promise rejected with=] a `TypeError`.
+        1. Let |clonedResponse| be a [=response/clone=] of |response|.
+        1. Let |bodyReadPromise| be [=a promise resolved with=] undefined.
+        1. If |response|'s [=response/body=] is non-null, run these substeps:
+            1. Let |stream| be |response|'s [=response/body=]'s [=body/stream=].
+            1. Let |reader| be the result of [=ReadableStream/get a reader|getting a reader=] for |stream|.
+            1. Set |bodyReadPromise| to the result of [=Read all bytes|reading all bytes=] from |stream| with |reader|.
+
+            Note: This ensures that |response|'s [=response/body=] is [=Body/locked=], and we have a full buffered copy of the body in |clonedResponse|.
```
annevk: I don't understand this. It seems there shouldn't be a clone at all. And if the stream is already tainted you'd throw.

docs/index.bs
jakearchibald: It might be a string (see 1.2).

docs/index.bs
jakearchibald: I'm creating a clone so the algorithm has a non-consumed copy that can be stored.

We're already throwing if the stream is tainted, but this algorithm needs to taint the stream.

wanderview: At the very least I think there should be a note indicating that this is there to simulate reading the body in order to write it to disk.  We don't actually require tee'ing the data in memory before writing the bits on to disk.

annevk: Interesting. What happens if the stream is rejected or a signal is aborted?

It also seems good to indicate that any failures are already dealt with for the original stream so this new stream should stream without error to disk assuming no hardware issues.

jakearchibald: @annevk `|bodyReadPromise|` is returned at the end of this algorithm, so any rejection there will become a rejection of `cache.put()`.

@wanderview good point. I'll expand on that.

annevk: How does the user of the API observe cacheJobPromise in that case?

docs/index.bs
```diff
@@ -1809,42 +1809,48 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-            1. [=Read all bytes=] from |dummyStream| with |reader|.
+        1. If |response|'s [=response/body=] is [=Body/disturbed=] or [=Body/locked=], return [=a promise rejected with=] a `TypeError`.
+        1. Let |clonedResponse| be a [=response/clone=] of |response|.
+        1. Let |bodyReadPromise| be [=a promise resolved with=] undefined.
+        1. If |response|'s [=response/body=] is non-null, run these substeps:
+            1. Let |stream| be |response|'s [=response/body=]'s [=body/stream=].
+            1. Let |reader| be the result of [=ReadableStream/get a reader|getting a reader=] for |stream|.
+            1. Set |bodyReadPromise| to the result of [=Read all bytes|reading all bytes=] from |stream| with |reader|.
+
+            Note: This ensures that |response|'s [=response/body=] is [=Body/locked=], and we have a full buffered copy of the body in |clonedResponse|. An implementation could optimise by streaming directly to disk rather than memory.
```
annevk: optimize*

jakearchibald: boooo

jakearchibald:
I think eventually we want a way to pipe a stream to disk and later get a readable for it.

docs/index.bs
jakearchibald: `cacheJobPromise` is returned in the fulfillment handler of `bodyReadPromise`

annevk: But the original version of this API returned cacheJobPromise, no?
```
bodyReadPromise.then(x => cacheJobPromise);
return bodyReadPromise
```
is not equal to
```
return cacheJobPromise
```
unless I'm missing something.

jakearchibald: oh ffs you're right. Fixing…

jakearchibald: (my excuse is jetlag)

Issue #1405, opened 2019-05-06
Discuss the maxScopeString and Service-Worker-Allowed in security considerations


jyasskin:
[`Service-Worker-Allowed`](https://w3c.github.io/ServiceWorker/#service-worker-allowed) and the [*maxScopeString* in Update](https://w3c.github.io/ServiceWorker/#update-algorithm) exist to protect sites that allow different users to operate within different paths but don't want those users to overwrite each other's content. This could use some discussion in the [Security Considerations](https://w3c.github.io/ServiceWorker/#security-considerations) since it's not particularly obvious that there is this protection or what its limits are.

mattto:
Thanks, added in #1409.

Issue #1307, opened 2018-04-23
Service-Worker-Allowed can be cross-origin to the script URL.


mattto:
I noticed we don't have WPT tests for Service-Worker-Allowed and started upstreaming Chrome's.

While doing that I added test cases for when SWA is an absolute URL, and found that Chrome accepts SWA that is cross-origin to the script URL, since it only looks at the path of the SWA. That seems a bit surprising. For example, this works if run on https://my-origin.com:
```
// Set the scope to an upper path of the script location
// Response included "Service-Worker-Allowed : https://another-origin.com/"
navigator.serviceWorker.register("/js/sw.js", { scope: "/" }).then(() => {
  console.log("Install succeeded as the max allowed scope was overriden to '/'.");
});
```
That seems to agree with the spec which has these steps:

1. Let maxScope be the result of parsing serviceWorkerAllowed with job’s script url.
2. Set maxScopeString to "/" concatenated with the strings in maxScope’s path (including empty strings), separated from each other by "/".

I believe that parsing an absolute URL with a cross-origin base URL ends up ignoring base URL, based on results from `new URL()` (correct me if I'm wrong...).

I think we should error on cross-origin SWA from the script URL though. Firefox seems to be doing this. Here is error output when script URL is same-origin as the scope at resources/empty-worker.js, and SWA was a cross-origin:

Failed to register a ServiceWorker: The path of the provided scope ‘http://127.0.0.1:8001/service-workers/service-worker/resources/this-scope-is-normally-allowed’ is not under the max scope allowed ‘https://www1.web-platform.test:8444/’. Adjust the scope, move the Service Worker script, or use the Service-Worker-Allowed HTTP header to allow the scope.

aleemb:
@mattto We tripped into a similar issue, however, in our case everything was on the same origin. When doing,

    navigator.serviceWorker.register("/js/sw.js", { scope: "/" })

This failed with the error `Failed to register a ServiceWorker: The path of the provided scope...`

The fix is to add the `Service-Worker-Allowed` header seems to trip up quite a few people. Is there any reason to add this header if the request is **not** cross-origin?

annevk:
@aleemb that's off-topic for this issue. It seems like the error message provides the reason (if you can inject a service worker script somewhere we don't want to easily let it take over the entire domain).

mattto:
[...]

PR #1409, opened 2019-05-22
Refinements and description of Service-Worker-Allowed.

mattto:
* Require Service-Worker-Allowed to be same-origin to the script URL
  (#1307)
* Add non-normative explanation of Service-Worker-Allowed (#1405) and
  other mitigations. The text is highly inspired by
  https://infrequently.org/2014/12/psa-service-workers-are-coming/.

***
<a href="https://pr-preview.s3.amazonaws.com/mattto/ServiceWorker/pull/1409.html" title="Last updated on May 29, 2019, 12:56 AM UTC (0eadf93)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1409/86d9414...mattto:0eadf93.html" title="Last updated on May 29, 2019, 12:56 AM UTC (0eadf93)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2124,6 +2124,26 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
   </section>
 
   <section>
+    <h3 id="path-restriction">Path restriction</h3>
+
+    *This section is non-normative.*
+
+    In addition to the [[#origin-restriction|origin restriction]], service workers are restricted by the [=url/path=] of the service worker script. For example, a service worker script at <code>https://www.example.com/~bob/sw.js</code> can be registered for the [=scope=] <code>https://www.example.com/~bob</code> but not for the scope <code>https://www.example.com</code> or <code>https://www.example.com/~alice</code>. This provides some protection for sites that host multiple-user content in separated directories on the same origin. However, the path restriction is not considered a hard security boundary, as only origins are. Sites are encouraged to use different origins to securely isolate segments of the site if appropriate.
+
+    Servers can break the path restriction by setting a [=Service-Worker-Allowed=] header on the service worker script.
+  </section>
+
+  <section>
+    <h3 id="script-request">Service worker script request</h3>
+
+    *This section is non-normative.*
+
+    To further defend against malicious registration of a service worker on a site, this specification requires that:
+    * The [=Service-Worker=] header is present on service worker script requests, and
+    * Service worker scripts are served with a [=JavaScript MIME type=].
+  </section>
+
+  <section>
     <h3 id="implementer-concerns">Implementer Concerns</h3>
 
     *This section is non-normative.*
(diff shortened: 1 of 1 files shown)


mattto:
@jakearchibald wdyt?

docs/index.bs
```diff
@@ -2119,6 +2119,26 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
   </section>
 
+  <section>
+    <h3 id="path-restriction">Path restriction</h3>
+
+    *This section is non-normative.*
+
+    In addition to the [[#origin-restriction|origin restriction]], service workers are restricted by the [=url/path=] of the service worker script. For example, a service worker script at <code>https://www.example.com/~bob/sw.js</code> can be registered for the [=scope=] <code>https://www.example.com/~bob</code> but not for the scope <code>https://www.example.com</code> or <code>https://www.example.com/~alice</code>. This provides some protection for sites that host multiple-user content in separated directories on the same origin. However, the path restriction is not considered a hard security boundary, as only origins are. Sites are encouraged to use different origins to securely isolate segments of the site if appropriate.
```
jakearchibald:
```suggestion
    In addition to the [[#origin-restriction|origin restriction]], service workers are restricted by the [=url/path=] of the service worker script. For example, a service worker script at <code>https://www.example.com/~bob/sw.js</code> can be registered for the [=scope=] <code>https://www.example.com/~bob/</code> but not for the scope <code>https://www.example.com/</code> or <code>https://www.example.com/~alice/</code>. This provides some protection for sites that host multiple-user content in separate directories on the same origin. However, the path restriction is not considered a hard security boundary, as only origins are. Sites are encouraged to use different origins to securely isolate segments of the site if appropriate.
```

Adding a trailing slash to the URLs, as `https://www.example.com/~bob/sw.js` won't be able to register for scope `https://www.example.com/~bob`.

docs/index.bs
```diff
@@ -2119,6 +2119,26 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+    Servers can break the path restriction by setting a [=Service-Worker-Allowed=] header on the service worker script.
```
jakearchibald:
```suggestion
    Servers can remove the path restriction by setting a [=Service-Worker-Allowed=] header on the service worker script.
```

Just a nit. Feel free to ignore.

docs/index.bs
```diff
@@ -2482,9 +2502,10 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
           1. Else:
               1. Let |maxScope| be the result of <a lt="URL parser">parsing</a> |serviceWorkerAllowed| with |job|'s [=job/script url=].
-              1. Set |maxScopeString| to "<code>/</code>" concatenated with the strings in |maxScope|'s [=url/path=] (including empty strings), separated from each other by "<code>/</code>".
+              1. If |maxScope|'s [=url/origin=] is |job|'s [=job/script url=]'s [=url/origin=]:
+                1. Set |maxScopeString| to "<code>/</code>" concatenated with the strings in |maxScope|'s [=url/path=] (including empty strings), separated from each other by "<code>/</code>".
           1. Let |scopeString| be "<code>/</code>" concatenated with the strings in |scopeURL|'s [=url/path=] (including empty strings), separated from each other by "<code>/</code>".
-          1. If |scopeString| starts with |maxScopeString|, do nothing.
+          1. If |maxScopeString| is not null and |scopeString| starts with |maxScopeString|, do nothing.
```
jakearchibald: Nit: Seems like a good opportunity to flip this around.

```
1. If |maxScopeString| is null or |scopeString| does not start with |maxScopeString|, then:
    1. Invoke [=Reject Job Promise=]…
```

docs/index.bs
mattto: Thanks, done.

docs/index.bs

docs/index.bs
mattto: Good point, done.

Issue #1408, opened 2019-05-20
Should register() trigger byte-for-byte checking when WorkerType is updated?


makotoshimazu:
```js
// (1) Register sw.js as classic script.
await navigator.serviceWorker.register('sw.js', {type: 'classic'});
// (2) Register sw.js as module script with the same script.
await navigator.serviceWorker.register('sw.js', {type: 'module'});
```
Should (2) install a new ServiceWorker and invoke install/activate event?

mattto:
I think I'd expect a new service worker to be installed and the byte-to-byte check skipped as if the URL changed.

See also https://github.com/w3c/ServiceWorker/issues/1359.

makotoshimazu:
Thanks!
The current spec seems to have another edge case where install/activate events are not triggered as follows, and it also can be solved by adding a "skip byte-to-byte checking" flag.

```js
// ==== script in something.html ====
// Register a.js, and it imports b.js too.
await navigator.serviceWorker.register('a.js');
// Register b.js, but b.js is identical to the one we previously put into the script cache map.
await navigator.serviceWorker.register('b.js');

// ==== a.js ====
a = true;
if (typeof b === 'undefined') {
  importScripts('b.js');
}

// ==== b.js ====
b = true;
if (typeof a === 'undefined') {
  importScripts('a.js');
}
```

For example, does changing [[Register]] like this and refers to the flag in the [[Update]] algorithm make sense?

5. If registration is not null, then:
    ....
    4. If newestWorker is not null, job's script url doesn't equal to newestWorker's script url or job's worker type doesn't equal to newestWorker's worker type, then:
        1. Set job's skip byte-for-byte comparison to true.

mattto:
Addressed in https://github.com/w3c/ServiceWorker/pull/1411

Issue #1359, opened 2018-10-04
Should add a “script type” check in the register algorithm.


d0iasm:
In the current spec, we finish a new register job when it equals _newestWorker_. Spec says ([Register 5. 3.](https://w3c.github.io/ServiceWorker/#register-algorithm) ): 
> 3. If newestWorker is not null, **job’s script url equals newestWorker’s script url, and job’s update via cache mode's value equals registration’s update via cache mode**, then:
>      1. Invoke Resolve Job Promise with job and registration.
>      2. Invoke Finish Job with job and abort these steps.

However, a new register job should be alive when its script url/update via cache equal newestWorker’s ones but its script type does not equal newestWorker’s one. I think script type should also be checked.

mattto:
Yes, this looks like just an oversight. It should be easy to patch the spec. Did you have WPT ready to test this?

d0iasm:
> Did you have WPT ready to test this?
Yes. I just added a new test for updating the registration with different script type.
https://github.com/web-platform-tests/wpt/pull/13692

d0iasm:
Just concern:
If in the case that
1. A user register SW as classic script type.
2. A user update it to module script type and the main script using module specific syntax(like “import ‘sw2.js’;”).
3. Only run Update Job  without running Register Job (It would happen when a user accesses a page in scope but not a top-level page).
4. Fail update because of an evaluation error.
Update Job keeps to fail until a user accesses a top-level page having navigator.serviceWorker.register(‘sw2.js’, {type: ‘module’});

Is this behavior intentional?

mattto:
> Is this behavior intentional?

If I'm understanding the concern, it's that the server has changed the script from a classic script to a module script and so Soft Update would fail because the registration is still set to classic script.

I think that's fine. To change the script type you should probably change the script URL to avoid this problem. I'm assuming there is no plan to do something like add info to HTTP response headers to let the browser know whether it's a module or classic script? @domenic

annevk:
There's https://github.com/whatwg/html/pull/3205, but that's really about confirming the request and hasn't gotten support (though maybe it could be used for this as well).

mattto:
Addressed in https://github.com/w3c/ServiceWorker/pull/1411.

Issue #1358, opened 2018-10-04
Should add “script type” and “update via cache”  when check two jobs are equivalent.


d0iasm:
In the current spec, we enqueue a new job to _jobQueue_ if it is not equivalent to the last job of a  _jobQueue_. Spec says ( [Schedule Job 6.](https://w3c.github.io/ServiceWorker/#schedule-job-algorithm) ):
> 1. Let lastJob be the element at the back of jobQueue.
> 2 .If job is **equivalent** to lastJob and lastJob’s job promise has not settled, append job to lastJob’s list of equivalent jobs.
> 3. Else, set job’s containing job queue to jobQueue, and enqueue job to jobQueue.

Also **equivalent** means in the spec ( [dfn-job-equivalent](https://w3c.github.io/ServiceWorker/#dfn-job-equivalent) ):
> Two jobs are equivalent when their job type is the same and:
> 	- **For register and update jobs, both their scope url and the script url are the same.**
> 	- For unregister jobs, their scope url is the same.

I think script type and update via cache should also be the same for equivalent register and update jobs. Because we might ignore a new job which has different script type/update via cache if we do not add this check.

mattto:
Yes this also looks like an oversight. Did you have WPT for this also?

mattto:
Addressed in https://github.com/w3c/ServiceWorker/pull/1411

Issue #1189, opened 2017-08-21
Does changing updateViaCache mint a new ServiceWorkerRegistration or update them all?


mattto:
Suppose you have something like this:
```javascript
var reg;
navigator.serviceWorker.getRegistration(url).then(() => { reg = doc; });
reg.updateViaCache;  // 'all';
register(sw, {updateViaCache: 'none'}).then(r => {
  reg === r;  // true?
  reg.updateViaCache; // 'all' or 'none'?
});
```

I'm leaning toward updating all existing ServiceWorkerRegistration objects, because otherwise ServiceWorkerGlobalScope#registration would have to mutate to a brand new registration object after register() settled, which doesn't feel right.

However, updating them all is a bit troublesome and seems to lead to race conditions (one of them has been updated but the others haven't). Would it be worth considering changing reg.updateViaCache to a promise similar to NavigationPreloadManager state?

@wanderview what does Firefox do?

mattto:
(The spec doesn't seem to address this, we get to "Invoke Update algorithm passing job as the argument" but nothing there updates updateViaCache AFAICT)

jakearchibald:
Good catch. The other properties are live, so it'd be confusing to make other properties a snapshot.

We could add `getState()` which we could add other settings to later.

```js
const { updateViaCache } = await reg.getState();
```

wanderview:
We only just recently landed this in firefox nightly.  There I think you will get:

```
var reg = await navigator.serviceWorker.getRegistration(url);
reg.updateViaCache;  // 'all';
var r = await register(sw, {updateViaCache: 'none'});
reg === r;  // false
reg.updateViaCache; // 'all'
r.updateViaCache; // 'none'
```

I think this behavior is determined by step 5.3 here in the spec:

https://w3c.github.io/ServiceWorker/#register-algorithm

You should skip the short-circuit because the cache mode is different and get a new registration.

wanderview:
And the update algorithm gets the cache mode out of the newest worker in step 3 and step 7.2 here:

https://w3c.github.io/ServiceWorker/#update-algorithm

So I'm not sure I understand the particular problem here.

wanderview:
Errr... my last comment makes no sense because `updateViaCache` is on the registration, not the worker.

I think changing the `updateViaCache` property is analogous to changing the `scriptURL`.  The problem is we put `updateViaCache` on the registration instead of the worker itself.

wanderview:
[...]

PR #1411, opened 2019-06-03
Improve "update via cache" and "worker type" in register algorithms.

mattto:
This makes the following changes:
* updateViaCache is a live property. This is covered by the WPT
registration-updateviacache.https.html. However we are missing tests for a
register() that rejects with a new updateViaCache property. The desired
behavior is to only update the property if register() resolves.
* Changing worker type bypasses the byte-for-byte update check. This is covered
by WPT update-registration-with-type.https.html.
* Adds Worker type and updateViaCache to the job equivalence check. This
probably can have a WPT for successive register() calls that vary these
properties don’t get coalesced into a single job.

Addresses #1189, #1408, #1359, and #1358.

***
<a href="https://pr-preview.s3.amazonaws.com/mattto/ServiceWorker/pull/1411.html" title="Last updated on Jun 3, 2019, 7:50 AM UTC (872100e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1411/bf8f037...mattto:872100e.html" title="Last updated on Jun 3, 2019, 7:50 AM UTC (872100e)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2516,9 +2516,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Set |updatedResourceMap|[|url|] to |response|.
           1. If |response|'s [=response/cache state=] is not "`local`", set |registration|'s [=last update check time=] to the current time.
           1. Set |hasUpdatedResources| to true if any of the following are true:
-            1. |newestWorker| is null.
-            1. |newestWorker|'s [=service worker/script url=] is not |url| or |newestWorker|'s [=service worker/type=] is not |job|'s [=worker type=].
-            1. |newestWorker|'s [=script resource map=][|url|]'s [=response/body=] is not byte-for-byte identical with |response|'s [=response/body=].
+            * |newestWorker| is null.
+            * |newestWorker|'s [=service worker/script url=] is not |url| or |newestWorker|'s [=service worker/type=] is not |job|'s [=worker type=].
+            * |newestWorker|'s [=script resource map=][|url|]'s [=response/body=] is not byte-for-byte identical with |response|'s [=response/body=].
           1. If |hasUpdatedResources| is false and |newestWorker|'s [=classic scripts imported flag=] is set, then:
 
               Note: The following checks to see if an imported script has been updated, since the main script has not changed.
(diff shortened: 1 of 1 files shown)


mattto:
Adding the missing tests at https://github.com/web-platform-tests/wpt/pull/17131

docs/index.bs
```diff
@@ -637,7 +637,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       Note: The {{ServiceWorkerContainer/register(scriptURL, options)}} method creates or updates a [=/service worker registration=] for the given [=service worker registration/scope url=]. If successful, a [=/service worker registration=] ties the provided |scriptURL| to a [=service worker registration/scope url=], which is subsequently used for <a lt="handle fetch">navigation matching</a>.
 
-      <dfn method for="ServiceWorkerContainer"><code>register(|scriptURL|, |options|)</code></dfn> method *must* run these steps:
+      The <dfn method for="ServiceWorkerContainer"><code>register(|scriptURL|, |options|)</code></dfn> method *must* run these steps:
```
jakearchibald: Hah we're pretty inconsistent about this. Fine with adding them as we go along. I agree "The" is better.

docs/index.bs
```diff
@@ -2515,14 +2515,16 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Let |url| be |request|'s [=request/url=].
           1. Set |updatedResourceMap|[|url|] to |response|.
           1. If |response|'s [=response/cache state=] is not "`local`", set |registration|'s [=last update check time=] to the current time.
-          1. Let |map| be |newestWorker|'s [=script resource map=] if |newestWorker| is not null, and null otherwise.
-          1. If |map| is null or |map|[|url|]'s [=response/body=] is not byte-for-byte identical with |response|'s [=response/body=], set |hasUpdatedResources| to true.
-          1. Else if |newestWorker|'s [=classic scripts imported flag=] is set, then:
+          1. Set |hasUpdatedResources| to true if any of the following are true:
+            1. |newestWorker| is null.
+            1. |newestWorker|'s [=service worker/script url=] is not |url| or |newestWorker|'s [=service worker/type=] is not |job|'s [=worker type=].
+            1. |newestWorker|'s [=script resource map=][|url|]'s [=response/body=] is not byte-for-byte identical with |response|'s [=response/body=].
```
jakearchibald: Nit: Since it's "any of the following", these should start "*" not "1.", since the order doesn't matter.

docs/index.bs
mattto: Thanks, agreed.

docs/index.bs

mattto:
Updated, and also revised slightly to set updateViaCache even when there is no byte-for-byte change (though we can discuss whether updateViaCache should trigger the update at all in another issue).

mattto:
Thanks, filed https://github.com/w3c/ServiceWorker/issues/1414 !

Issue #1204, opened 2017-10-06
resurrected service workers don't fire an activate event which may break expectations


wanderview:
Consider a service worker that looks like this:

```js
self.addEventListener('activate', evt => evt.waitUntil(clients.claim()));
```

And then a main page that waits for a window to be controlled:

```js
navigator.serviceWorker.register('sw.js');
await waitForControlled(existingWindow);
```

This works fine if the SW is installed fresh.  If the registration resurrects a service worker that was previously unregistered, then this code will break.  The activate event will not fire and the claim() will not occur.

Should we consider firing another `activate` event or some other new event when the service worker is resurrected like this?  It seems that the intent of restoring the SW to the active slot is that it should be equivalent to registering the new worker.

n8schloss:
I think that makes sense, this is a hard edge case for many developers to reason about. I do think that if we fire 'activate' again that we probably want to fire 'install' again too. I could see developers getting confused about 'activate' firing in this case but not 'install'.

NekR:
> If the registration resurrects a service worker that was previously unregistered, then this code will break.

What does this mean exactly? How does this happen?

wanderview:
> What does this mean exactly? How does this happen?

When you call `unregister()` on a registration with an active worker that is still controlling clients, then the worker doesn't die immediately.  It gets an uninstalling flag and will normally go away once the last client closes.  If someone calls register() with the same params before that happens, though, then SW is "resurrected".  We just remove the uninstalling flag and let it keep on living.  While the uninstalling flag is set, though, the worker misses out on controlling clients, doesn't show in getRegistrations(), etc.

> I think that makes sense, this is a hard edge case for many developers to reason about. I do think that if we fire 'activate' again that we probably want to fire 'install' again too. I could see developers getting confused about 'activate' firing in this case but not 'install'.

Well, if we are going to do that, maybe we should just get rid of resurrection completely and make a new installing worker in this case.  Once a worker is doomed as uninstalling it would never come back.

Also, right now we have an invariant that SW firing the install event are reg.installing.  That wouldn't hold for what you are proposing here and could break things.  (Unlikely, though.)

NekR:
From JS perspective is sounds like ServiceWorker is being removed completely and then added again. I mean, that's certainly unexpected, especially if that worker doesn't control anything anymore (if I got it right) and isn't showing in any API to be alive.

I is how I think it's working (I haven't thought much uninstalling process before):

* SW registered
[...]

PR #1415, opened 2019-06-03
Kill resurrection

jakearchibald:
Fixes https://github.com/w3c/ServiceWorker/issues/1204.
Tests https://github.com/web-platform-tests/wpt/pull/17139.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1415.html" title="Last updated on Jun 4, 2019, 5:59 AM UTC (88f1b6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1415/c1f3619...88f1b6a.html" title="Last updated on Jun 4, 2019, 5:59 AM UTC (88f1b6a)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -210,8 +210,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker registration=] has an associated <dfn export id="dfn-update-via-cache">update via cache mode</dfn>, which is "`imports`", "`all`", or "`none`". It is initially set to "`imports`".
 
-    A [=/service worker registration=] has an associated <dfn export id="dfn-uninstalling-flag">uninstalling flag</dfn>. It is initially unset.
-
     A [=/service worker registration=] has one or more <dfn export id="dfn-service-worker-registration-task-queue">task queues</dfn> that back up the <a>tasks</a> from its <a>active worker</a>'s <a>event loop</a>'s corresponding [=/task queues=]. (The target task sources for this back up operation are the <a>handle fetch task source</a> and the <a>handle functional event task source</a>.) The user agent dumps the <a>active worker</a>'s <a>tasks</a> to the [=/service worker registration=]'s [=service worker registration/task queues=] when the <a>active worker</a> is <a lt="terminate service worker">terminated</a> and <a lt="queue a task">re-queues those tasks</a> to the <a>active worker</a>'s <a>event loop</a>'s corresponding [=/task queues=] when the <a>active worker</a> spins off. Unlike the [=/task queues=] owned by <a>event loops</a>, the [=/service worker registration=]'s [=service worker registration/task queues=] are not processed by any <a>event loops</a> in and of itself.
 
     A [=/service worker registration=] has an associated <dfn export>{{NavigationPreloadManager}}</dfn> object.
@@ -219,6 +217,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     A [=/service worker registration=] has an associated <dfn export>navigation preload enabled flag</dfn>. It is initially unset.
 
     A [=/service worker registration=] has an associated <dfn export>navigation preload header value</dfn>, which is a [=byte sequence=]. It is initially set to \`<code>true</code>\`.
+
+    A [=/service worker registration=] is said to be <dfn export id="dfn-service-worker-registration-uninstalled">unregistered</dfn> if [=scope to registration map=][[=service worker registration/scope url=]] is not this [=/service worker registration=].
 
     <section>
       <h4 id="service-worker-registration-lifetime">Lifetime</h4>
@@ -678,7 +678,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. Run the following steps [=in parallel=]:
             1. Let |registrations| be a new [=list=].
             1. [=map/For each=] |scope| → |registration| of [=scope to registration map=]:
-                1. If the [=url/origin=] of the result of [=URL parser|parsing=] |scope| is the [=same origin|same=] as |client|'s [=environment settings object/origin=], and |registration|'s [=uninstalling flag=] is unset, then [=append=] |registration| to |registrations|.
+                1. If the [=url/origin=] of the result of [=URL parser|parsing=] |scope| is the [=same origin|same=] as |client|'s [=environment settings object/origin=], then [=append=] |registration| to |registrations|.
             1. [=Queue a task=] on |promise|'s [=relevant settings object=]'s [=responsible event loop=], using the [=DOM manipulation task source=], to run the following steps:
                 1. Let |registrationObjects| be a new [=list=].
                 1. [=list/For each=] |registration| of |registrations|, [=append=] the {{ServiceWorkerRegistration}} object associated with |registration| to |registrationObjects|.
@@ -1297,6 +1297,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                 1. If |client| is not a [=secure context=], [=continue=].
                 1. Let |registration| be the result of running <a>Match Service Worker Registration</a> algorithm passing |client|'s <a>creation URL</a> as the argument.
                 1. If |registration| is not the [=ServiceWorkerGlobalScope/service worker=]'s <a>containing service worker registration</a>, [=continue=].
+
+                    Note: |registration| will be null if the [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=] is [=service worker registration/unregistered=].
+
                 1. If |client|'s <a>active service worker</a> is not the [=ServiceWorkerGlobalScope/service worker=], then:
                     1. Invoke <a>Handle Service Worker Client Unload</a> with |client| as the argument.
                     1. Set |client|'s <a>active service worker</a> to [=ServiceWorkerGlobalScope/service worker=].
@@ -1360,8 +1363,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         1. Upon [=upon fulfillment|fulfillment=] or [=upon rejection|rejection=] of |promise|, [=queue a microtask=] to run these substeps:
             1. Decrement |event|'s [=ExtendableEvent/pending promises count=] by one.
             1. If |event|'s [=ExtendableEvent/pending promises count=] is 0, then:
-                1. Let |registration| be the [=context object=]'s [=relevant global object=]'s associated [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=].
-                1. If |registration|'s [=uninstalling flag=] is set, invoke [=Try Clear Registration=] with |registration|.
+                1. Let |registration| be the [=current global object=]'s associated [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=].
+                1. If |registration| is [=service worker registration/unregistered=], invoke [=Try Clear Registration=] with |registration|.
                 1. If |registration| is not null, invoke [=Try Activate=] with |registration|.
 
       The user agent *should not* [=terminate service worker|terminate=] a [=/service worker=] if [=Service Worker Has No Pending Events=] returns false for that [=/service worker=].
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -518,12 +518,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
       <h4 id="navigator-service-worker-unregister">{{ServiceWorkerRegistration/unregister()}}</h4>
 
-      Note: The {{ServiceWorkerRegistration/unregister()}} method unregisters the [=/service worker registration=]. It is important to note that the currently <a>controlled</a> [=/service worker client=]'s <a>active service worker</a>'s <a>containing service worker registration</a> is effective until all the [=/service worker clients=] (including itself) using this [=/service worker registration=] unload. That is, the {{ServiceWorkerRegistration/unregister()}} method only affects subsequent <a lt="navigate">navigations</a>.
+      Note: The {{ServiceWorkerRegistration/unregister()}} method unregisters the [=/service worker registration=]. It is important to note that the currently [=controlled=] [=/service worker client=]'s [=active service worker=]'s [=containing service worker registration=] is effective until all the [=/service worker clients=] (including itself) using this [=/service worker registration=] unload. That is, the {{ServiceWorkerRegistration/unregister()}} method only affects subsequent [=navigate|navigations=].
```
jakearchibald: Just converting to [=bikeshed shorthand=]. Couldn't help myself. Sorry.

docs/index.bs
```diff
@@ -1336,28 +1339,35 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
       <h4 id="wait-until-method">{{ExtendableEvent/waitUntil()|event.waitUntil(f)}}</h4>
 
-      {{ExtendableEvent/waitUntil()}} method extends the lifetime of the event.
+      Note: {{ExtendableEvent/waitUntil()}} method extends the lifetime of the event.
 
       <dfn method for="ExtendableEvent"><code>waitUntil(|f|)</code></dfn> method *must* run these steps:
 
-        1. If the {{Event/isTrusted}} attribute is false, [=throw=] an "{{InvalidStateError}}" {{DOMException}}.
+        1. Let |event| be the [=context object=].
+        1. [=ExtendableEvent/Add lifetime promise=] |f| to |event|.
```
jakearchibald: I moved this into its own algorithm as we have it copy-pasted in the spec twice, and it's something other specs may want to do.

docs/index.bs
```diff
@@ -1336,28 +1339,35 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
             Note: The [=ExtendableEvent/pending promises count=] is incremented even if the given promise has already been settled. The corresponding count decrement is done in the microtask queued by the reaction to the promise.
 
-        1. Upon [=upon fulfillment|fulfillment=] or [=upon rejection|rejection=] of |f|, [=queue a microtask=] to run these substeps:
-            1. Decrement the [=ExtendableEvent/pending promises count=] by one.
-            1. Let |registration| be the [=context object=]'s [=relevant global object=]'s associated [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=].
-            1. If |registration|'s [=uninstalling flag=] is set, invoke [=Try Clear Registration=] with |registration|.
-            1. If |registration| is not null, invoke [=Try Activate=] with |registration|.
+        1. Upon [=upon fulfillment|fulfillment=] or [=upon rejection|rejection=] of |promise|, [=queue a microtask=] to run these substeps:
+            1. Decrement |event|'s [=ExtendableEvent/pending promises count=] by one.
+            1. If |event|'s [=ExtendableEvent/pending promises count=] is 0, then:
```
jakearchibald: Seems pointless to do the next steps if the count is > 0.

docs/index.bs
```diff
@@ -2605,7 +2606,10 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
       1. Invoke [=Resolve Job Promise=] with |job| and |registration|.
-      1. <a>Queue a task</a> to <a>fire an event</a> named <code>updatefound</code> at all the {{ServiceWorkerRegistration}} objects for all the [=/service worker clients=] whose <a>creation URL</a> <a lt="Match Service Worker Registration">matches</a> |registration|'s [=service worker registration/scope url=] and all the [=/service workers=] whose <a>containing service worker registration</a> is |registration|.
+      1. Let |settingsObjects| be all [=environment settings objects=] whose [=environment settings object/origin=] is |registration|'s [=service worker registration/scope url=]'s origin.
+      1. For each |settingsObject| of |settingsObjects|, [=queue a task=] on |settingsObject| to run the following steps:
+          1. Let |registrationObjects| be every {{ServiceWorkerRegistration}} object in |settingsObject|'s [=environment settings object/realm execution context=], whose [=ServiceWorkerRegistration/service worker registration=] is |registration|.
+          1. For each |registrationObject| of |registrationObjects|, [=fire an event=] on |registrationObject| named `updatefound`.
```
jakearchibald: I needed to fix the "Match Service Worker Registration" stuff here anyway, so I tried to make event firing a bit more explicit at the same time.

jakearchibald:
Test PR coming soon

jakearchibald:
It's worth noting that this surfaces more of the weirdness in `navigator.serviceWorker.ready`.

1. Register SW.
1. await `navigator.serviceWorker.ready`.
1. Unregister & reregister.

Now you're in a state where `navigator.serviceWorker.ready` has settled, but the service worker in your scope isn't actually ready.

docs/index.bs
```diff
@@ -1435,18 +1445,10 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       <dfn method for="FetchEvent"><code>respondWith(|r|)</code></dfn> method *must* run these steps:
 
+        1. Let |event| be the [=context object=].
         1. If the <a>dispatch flag</a> is unset, [=throw=] an "{{InvalidStateError}}" {{DOMException}}.
         1. If the [=FetchEvent/respond-with entered flag=] is set, [=throw=] an "{{InvalidStateError}}" {{DOMException}}.
```
mattto: Should these be "event's dispatch flag..." and "event's respond-with entered flag" instead of "the flag"?

docs/index.bs
```diff
@@ -220,6 +218,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker registration=] has an associated <dfn export>navigation preload header value</dfn>, which is a [=byte sequence=]. It is initially set to \`<code>true</code>\`.
 
+    A [=/service worker registration=] is said to be <dfn export id="dfn-service-worker-registration-unregistered">unregistered</dfn> if [=scope to registration map=][[=service worker registration/scope url=]] is not this [=/service worker registration=].
```
mattto: nitty: is it required to say "this registration's scope url" to qualify "scope url"?

docs/index.bs
mattto: Thanks I've been doing the same :)

docs/index.bs
mattto: Looks nicer.

docs/index.bs
mattto: Agreed.

docs/index.bs
```diff
@@ -2454,7 +2455,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       :: none
 
       1. Let |registration| be the result of running the <a>Get Registration</a> algorithm passing |job|'s [=job/scope url=] as the argument.
-      1. If |registration| is null or |registration|'s <a>uninstalling flag</a> is set, then:
+      1. If |registration| is null or |registration| is [=service worker registration/unregistered=], then:
```
mattto: Can Get Registration return an unregistered registration?

docs/index.bs
mattto: Is a |settingsObject| a task source? Other parts of the spec seem to use language like "relevant settings object ’s responsible event loop and the DOM manipulation task source."

docs/index.bs
```diff
@@ -3177,7 +3178,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Set |matchingScope| to the result of <a lt="URL parser">parsing</a> |matchingScopeString|.
           1. Assert: |matchingScope|'s [=url/origin=] and |clientURL|'s [=url/origin=] are [=same origin=].
       1. Let |registration| be the result of running <a>Get Registration</a> algorithm passing |matchingScope| as the argument.
-      1. If |registration| is not null and |registration|'s <a>uninstalling flag</a> is set, return null.
       1. Return |registration|.
```
mattto: Should we collapse these two steps into just "Return the result of running..."

docs/index.bs
```diff
@@ -3187,15 +3187,14 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
 
       1. Run the following steps atomically.
       1. Let |scopeString| be the empty string.
       1. If |scope| is not null, set |scopeString| to [=URL serializer|serialized=] |scope| with the *exclude fragment flag* set.
-      1. Let |registration| be null.
       1. [=map/For each=] |key| → |value| of <a>scope to registration map</a>:
-          1. If |scopeString| matches |key|, set |registration| to |value|.
-      1. Return |registration|.
+          1. If |scopeString| matches |key|, then return |value|.
+      1. Return null.
```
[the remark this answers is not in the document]
mattto: Nice.

docs/index.bs
jakearchibald: Yeah, probably for the best. Will change.

docs/index.bs
jakearchibald: Yeah, I got a bit lazy since I wasn't touching these lines, but I should fix this up while I'm here.

docs/index.bs
jakearchibald: Hah, no. Good catch.

docs/index.bs
jakearchibald: Ugh yes, I forgot the [=responsible event loop=]. I'll add a task source too.

docs/index.bs
```diff
@@ -3007,7 +3005,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. If |registration| is null, then:
           1. Invoke <a>Resolve Job Promise</a> with |job| and false.
           1. Invoke <a>Finish Job</a> with |job| and abort these steps.
-      1. Set |registration|'s <a>uninstalling flag</a>.
+      1. [=map/Remove=] [=scope to registration map=][|job|'s [=job/scope url=]].
       1. Invoke <a>Resolve Job Promise</a> with |job| and true.
       1. Invoke [=Try Clear Registration=] with |registration|.
```
jungkees: *This comment should not block merging the PR.*

@mattto, I just saw https://chromium.googlesource.com/chromium/src.git/+/fafebc8a59d87a8c6f0620a91ee7cb3a0d8331eb. Do you think we should use Clear Registration over Try Clear Registration here as well?

Also, we didn't do "drop the active version as a controller of any clients" in the spec in any case. Because of the changes this PR makes, any reload (and subsequent navigations for sure) won't be able to use the unregistered registration again. So, the clients having the active worker of the unregistered registration as a controller will be isolated in their own scenario within their lifetime. Do you think we still need it given the changes in this PR?

mattto: We could discuss in an issue, but I think dropping the active version is orthogonal to this PR. The point of dropping the active version is a kind of "unclaim" or "immediate unregister" that's been proposed in other issues. In the case of the Chrome CL, we did this because it's an emergency force delete situation when something's gone wrong with the service worker (e.g., disk corruption).

jungkees: That sounds reasonable. I agree "unclaim" is a separate issue we can discuss, if necessary. I am fine with Unregister using Try Clear Registration as in this PR too.

jakearchibald:
I'll file bugs once this lands.

jakearchibald:
Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=971571 (cc @mattto @SteveBeckerMSFT) 
Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1557244 (cc @asutherland)
WebKit: https://bugs.webkit.org/show_bug.cgi?id=198602 (cc @youennf)

PR #1416, opened 2019-06-04
Central algorithm for updating service worker state and registrations

### LINKED ISSUE #1273: Should the worker be removed from the registration *before* its state is set to "redundant"?
While investigating some web-platform-tests failures on WebKit, a possible difference in behavior between Firefox/Chrome and the specification.

For example, when the install step fails, the spec says:
If installFailed is true, then:
1. Run the Update Worker State algorithm passing registration’s installing worker and redundant as the arguments.
2. Run the Update Registration State algorithm passing registration, "installing" and null as the arguments.

Notice the order, we first set the worker state to redundant, then we set the registration's installing worker to null. This means that when the statechange event is fired on this worker, registration.installing is not set to null yet (according to my understanding of the spec).

However, local testing shows that in Firefox and Chrome, registration.installing already returns null when the statechange event is fired at the worker (with state being redundant).

Should the specification be updated to match the behavior of browsers? Or did I misinterpret the specification?

I believe there is a similar issue in the "Clear Registration" algorithm. There is a WPT test that seems to assume that when clearing the registration, and the waiting worker becomes redundant, then all the registration's installing/waiting/active worker should be null. This test seems to pass in Chrome/Firefox.
However, my understanding is that the spec says:
1. Set installing worker state to redundant & schedule a task to fire statechange event at worker
2. Schedule a task to update ServiceWorkerRegistration(s) installing worker to null
3. Set waiting worker state to redundant & schedule a task to fire statechange event at worker
4. Schedule a task to update ServiceWorkerRegistration(s) waiting worker to null
5. Set active worker state to redundant & schedule a task to fire statechange event at worker
6. Schedule a task to update ServiceWorkerRegistration(s) active worker to null

So when the statechange is fired at the waiting worker, based on the spec, I would not expect registration.waiting / registration.active to become null right away. Instead, I would have expected registration.waiting to return null after 1 event loop iteration, and  registration.active even later.

jakearchibald:
Fixes #1273.

My aim here was to create a central place that applies updates to workers, registrations, and their various JS objects. This means the rest of the spec doesn't have to try to guess the correct order.

I update all the JS objects before firing events, so objects should be in the current state once the related events fire.

Additionally, I wanted to create a single task (per environment) for firing these events.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1303,7 +1303,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                 1. If |client|'s <a>active service worker</a> is not the [=ServiceWorkerGlobalScope/service worker=], then:
                     1. Invoke <a>Handle Service Worker Client Unload</a> with |client| as the argument.
                     1. Set |client|'s <a>active service worker</a> to [=ServiceWorkerGlobalScope/service worker=].
-                    1. Invoke <a>Notify Controller Change</a> algorithm with |client| as the argument.
+                    1. [=Queue a task=] on |client|'s [=responsible event loop=] using the [=DOM manipulation task source=] to [=fire an event=] named `controllerchange` at the {{ServiceWorkerContainer}} object that |client| is [=ServiceWorkerContainer/service worker client|associated=] with.
             1. Resolve |promise| with undefined.
         1. Return |promise|.
     </section>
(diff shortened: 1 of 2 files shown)


jakearchibald:
I'll write tests for this. We can tweak the order of events if need be. I tried to do it 'bubbling', so the leaf thing (the service worker) gets the event first, then the registration, then `navigator.serviceWorker`.

wanderview:
Do you expect this to change any previously defined behavior?  Or do you believe its a refactor without functional changes.

jakearchibald:
If things were implemented as spec'd, you'd get an event per task for related things. Eg if you have two registration objects representing the `ServiceWorkerRegistration`, things like `updatefound` would happen in different tasks per objects. It's now part of the same task.

Same if a worker becomes `activating` which also results in a controller change, and the previous worker becoming `redundant`. Those will now be dispatched in the same task whereas it was previously three tasks.

Using a single task for events that are part of the same action seems to make sense. As for the exact event order, I'm open to tweaking that if all browsers seem to behave the same way.

wanderview:
The statechange stuff at least matches what I implemented in firefox:

https://searchfox.org/mozilla-central/rev/f8b11433159cbc9cc80500b3e579d767473fa539/dom/serviceworkers/ServiceWorkerRegistration.cpp#396-489

Firefox does not run the updatefound event in the same task currently, but is guaranteed to be after the statechange events.  So that kind of matches here.

I'm unsure of the controllerchange event right now.

Getting all this right in the face of cross-process implementations is pretty tricky and I spent a lot of time getting firefox to match the current tests.  I expect anything that changes behavior may have a high implementation cost to adjust to.  I just want to make sure we're not making behavioral changes that don't have a solid reason to avoid churning implementations.

jakearchibald:
If the cost looks too high I can add an additional task for `controllerchange`. I'd rather not, but it isn't a big deal. Consistent event ordering is the most important thing.

wanderview:
Is the multiple-task vs single-task difference easily observable to script?  I expect trying to get statechange, updatefound, and controllerchange in the same task will be nearly impossible with the current firefox implementation.

wanderview:
cc @asutherland

jakearchibald:
> Is the multiple-task vs single-task difference easily observable to script?

You could use a message port to queue a task, or `requestAnimationFrame` which *might* happen before the next task. It isn't guaranteed though.

With these changes, you're guaranteed *not* to have other tasks or `requestAnimationFrame` between these related events.

wanderview:
I guess I'm not sure guaranteeing consistency with tasks posted from other task sources is worth the implementation pain this would cause.

jakearchibald:
If it looks bad in the tests I can split it back out into tasks but keep the central algorithm for ordering. It'd look weird at that point so I'd add a note saying it's for legacy reasons.

wanderview:
I found the code that guarantees that controllerchange from a worker going active will occur after the statechange events:

https://searchfox.org/mozilla-central/rev/f8b11433159cbc9cc80500b3e579d767473fa539/dom/serviceworkers/ServiceWorkerRegistrationInfo.cpp#329-335

The TransitionWaitingToActive() will queue a task for the statechange stuff and then UpdateClientControllers() will queue a task for controllerchange events.  Of course, there is nothing here for updatefound because updatefound is not fired for activating a worker.

mattto:
Do we have an idea which tests will change?

Chrome also does most things in separate tasks. Each change is typically its own IPC, and each IPC is its own task. So Chrome acts like Firefox in that statechange is one task and controllerchange is another. Chrome also doesn't have a good system yet of changing all the JS objects first and then firing the events. Not really sure yet how difficult it would be to change these.

wanderview:
Another slight difference is that firefox can resolve its ready promise after statechange events fire but before controllerchange/updatefound fire.

Related code:

https://searchfox.org/mozilla-central/rev/f8b11433159cbc9cc80500b3e579d767473fa539/dom/serviceworkers/ServiceWorkerContainer.cpp#572-577
https://searchfox.org/mozilla-central/rev/f8b11433159cbc9cc80500b3e579d767473fa539/dom/serviceworkers/ServiceWorkerRegistration.cpp#157-166

jakearchibald:
Isn't it inefficient using many tasks to update objects like this?

wanderview:
I don't think these events fire enough to really worry about efficiency.  Its not a hot path.

I think we probably queue tasks this way because the spec used to be written that way and that influenced design decisions.  Reversing those design decisions might be costly, though.

Also, some of the existing tests may require some ordering that is not clearly defined in the spec.  To be honest at some point I was just trying to get the WPT tests to pass.

jakearchibald:
I've started going through the tests and looking at the event ordering and other assumptions.

* `activate-event-after-install-state-change.https`
    * Asserts `registration.active` is null once sw state becomes `installed` on new registration.
* `activation-after-registration.https`
    * Asserts `registration.installing.state` is `installing` on a new registration.
* `activation.https`
    * After worker becomes `installed`, it waits on `wait_for_activation_on_dummy_scope` then asserts `registration.waiting` is set. 
    * Asserts `registration.active` is a given worker once that worker's state is `activating`.
    * Asserts `active`, `waiting`, `installing` on a registration are null once the waiting worker's state becomes `redundant` on an unregistered registration.
    * `wait_for_activation_on_dummy_scope` shouldn't be needed with this PR, but maybe leave tests as they are?
* `active.https`
    * Asserts `registration.active.state` is `activating` once `registration.installing`'s state becomes `activating`.
* `claim-affect-other-registration.https`
    * Once `registration.installing` becomes state `activated`, asserts `.installing` and `.waiting` are null, and `.active` is the worker that was previously `registration.installing`.
    * `registration.installing` is set when calling `register` with an existing scope but new url & the promise resolves.
    * Once `registration.installing` becomes state `installed`, asserts `.waiting` is the worker that was previously `registration.installing`.
    * Once `registration.waiting` becomes state `activated`, asserts `.active` is the worker that was previously `registration.waiting`, and old `.active` worker is now state `redundant`.

I'm now as far as `indexeddb.https` and haven't encountered anything new/different. I imagine most of the stuff will be in the registration tests.

The main thing to note is that `registration` is expected to be updated by the time the state change event fires on one of the workers.

I'll press on tomorrow.

wanderview:
> The main thing to note is that registration is expected to be updated by the time the state change event fires on one of the workers.

Yea.  Just for extra implementation context, trying to keep the separate registration and worker objects in a coherent state was very tricky for me implementing cross-process support.  In the end I basically had to treat the worker objects as children of the registration and funnel all IPC traffic for them through the registration.

jakearchibald:
That seems totally reasonable given that a worker cannot transfer between registrations.

wanderview:
> That seems totally reasonable given that a worker cannot transfer between registrations.

Its a little bit wonky because there are cases where a document may have a ServiceWorker object, but not have its corresponding ServiceWorkerRegistration object.  For example, a client receives a postMessage from the SW and so have the msg.source attribute.  My solution ends up creating the ServiceWorkerRegistration even though its not exposed in these cases.

jakearchibald:
Out of curiosity, if it had been spec'd like this PR from the start, would it have been easier, or just the same?

wanderview:
Maybe?  Not sure.  Its definitely easier to read and understand, though.

docs/index.bs
```diff
@@ -3068,88 +3049,74 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+          1. If `updateFound` is true, then [=list/for each=] |registrationObject| of |registrationObjects|, [=fire an event=] on |registrationObject| named `updatefound`.
+          1. If |newActiveWorker| is true:
+              1. Let |container| be the {{ServiceWorkerContainer}} object within |settingsObject|'s [=environment settings object/realm execution context=].
+              1. Let |readyPromise| be |container|'s [=ServiceWorkerContainer/ready promise=].
+              1. If |settingsObject| [=/uses=] |registration|, then:
+                  1. Set |settingsObject|'s [=environment/active service worker=] to |activeWorker|.
+                  1. [=fire an event=] named `controllerchange` on |container|.
+              1. <span id="update-state-resolve-ready-step">If the [=Match Service Worker Registration|matching=] [=/service worker registration=] for |settingsObject|'s [=environment/creation URL=] is |registration|, and |readyPromise| is pending, then [=/resolve=] |readyPromise| with a new {{ServiceWorkerRegistration}} object representing |registration| in |readyPromise|'s [=relevant Realm=].</span>
```
jakearchibald: This shouldn't be a new {{ServiceWorkerRegistration}}. It should return the same object if one exists. This is kinda hand-wavey in the spec.

jakearchibald:
😈 If I'm really trying to cut down the number of tasks, pages on the same event loop should share a task. This would make things consistent between iframes that have synchronous script access.

jakearchibald:
Ok, I've done going through every test, but I didn't find any that require these events to be posted in separate tasks, *except* `updatefound`, which happens in a task after `register()` resolves.

jakearchibald:
When an oldWorker is active, and a newWorker replaces it, Chrome and Safari fire `statechange` on oldWorker first. Firefox fires it on newWorker first.

We should spec one or the other.

wanderview:
> Firefox fires it on newWorker first.

This should be somewhat straightforward to fix in this code here:

https://searchfox.org/mozilla-central/rev/f8b11433159cbc9cc80500b3e579d767473fa539/dom/serviceworkers/ServiceWorkerRegistration.cpp#405-454

jakearchibald:
The original issue (https://github.com/w3c/ServiceWorker/issues/1273) is about consistency of state, so I guess that's the important bit. 

https://w3c.github.io/ServiceWorker/#installation-algorithm steps 17 & 18 leave the possibility of JavaScript seeing the same worker in `.installing` and `.waiting`.

There are many places where the a worker becomes redundant, but it's still in `.installing`/`.waiting`/`.active`.

https://w3c.github.io/ServiceWorker/#activation-algorithm steps 3 & 4 leaves the possibility of JavaScript seeing the same worker in `.waiting` and `.active`.

Even if we swap things around (as originally proposed in https://github.com/w3c/ServiceWorker/issues/1273), you can still end up with JavaScript seeing a worker in `.waiting` that has a state of `installing`.

If objects are updating across multiple tasks, we're always going to have this problem.

How about:

1. For a single update (like a worker being promoted to active, and the current active worker becoming redundant), we update all related JS objects across globals that share the event loop.
1. Related events then fire, which may be split across tasks, as long as they fire before the next state update.

@mattto 

> Chrome also doesn't have a good system yet of changing all the JS objects first and then firing the events. Not really sure yet how difficult it would be to change these.

I can't think of a better way to achieve consistency here. Is it _too difficult_?

@wanderview @asutherland Do you have a feel about this in terms of Mozilla's code base.

@youennf @cdumez Do you have feelings on this in terms of WebKit?

jakearchibald:
Summary for folks arriving here for the first time:

When *newWorker* (a service worker) is promoted from "waiting" to "active", overwriting *oldWorker* (another service worker), the following happens:

1. *oldWorker*'s state becomes `"redundant"`.
1. *newWorker*'s state becomes `"activating"`.
1. `registration.waiting` becomes `null`.
1. `registration.active` becomes *newWorker*.

The current version of the spec performs this update over 4 separate tasks in a single realm, meaning JavaScript can run in between and observe a half-updated state. All updates to JS objects & related events follow this pattern.

I'm proposing we switch to a single task per event loop, where the objects are updated, then the events are fired.

@wanderview suspects this would be a massive amount of work for Firefox. @matto is looking into it for Chrome.

If it makes things easier for implementers, it's probably fine to fire events over a series of tasks. However, I think we should aim to have *all* JS objects updated before firing any of the related events.

Issue #1200, opened 2017-09-25
is the browser required to start the ServiceWorker for postMessage() if there is no message event handler?


wanderview:
Consider this code:

```
// main window script
let bc = new BroadcastChannel('sw_side_channel');
sw.onmessage = evt => console.log('service worker started up');
navigator.serviceWorker.controller.postMessage('ping');

// service worker script
let bc = new BroadcastChannel('sw_side_channel');
bc.postMessage('pong');
self.onmessage = null;  // no direct message event handler
```

Is the browser required to start the service worker if its not running and a postMessage() is fired at while there is no message event handler?  Worker startup is observable thanks to things like BroadcastChannel or Clients API.

Or can the browser do similar optimizations like we do for FetchEvent where we don't start the worker if there is no message handler?

jakearchibald:
I think we should do it for all events.

https://w3c.github.io/ServiceWorker/#handle-functional-event-algorithm covers it for "functional events", but "message" isn't one of those. We should be consistent here.

wanderview:
Hmm, I didn't realize this mechanism had been expanded beyond FetchEvent.  Does chrome do this for other events yet?

jakearchibald:
Not sure about implementation. I'm looking at changing that functional event algorithm btw, but not that particular behaviour https://github.com/w3c/ServiceWorker/pull/1199

wanderview:
I just thought making event handlers observable at all was contentious so we limited it to fetch.

jakearchibald:
I don't have strong feelings here. Happy to make it just fetch.

mattto:
Chrome just does the optimization for fetch currently.

jakearchibald:
Action: We should make this behaviour optional in the spec, but allow it for any top level event, including `message`.

jakearchibald:
Action: As before. Make it a *may*.

mattto:
Addressed in https://github.com/w3c/ServiceWorker/pull/1417

mattto:
At https://github.com/whatwg/dom/pull/653 @annevk argues it should be limited to fetch or should be a _must_ rather than a _may_.

yoshisatoyanagisawa:
https://w3c.github.io/ServiceWorker/#should-skip-event may work?

yoshisatoyanagisawa:
Then, may be addressed by https://github.com/w3c/ServiceWorker/pull/1417 ?

monica-ch:
[...]

PR #1417, opened 2019-06-05
 Unify the event skipping algorithm when there is no event handler.

mattto:
This makes {install,activate,fetch,message} events and
functional events use the same algorithm to determine
whether to skip firing the event, and makes it a *may*
rather than *must*.

It also unifies the concept of a stale registration to reduce
duplicated text.

Addresses #1200

***
<a href="https://pr-preview.s3.amazonaws.com/mattto/ServiceWorker/pull/1417.html" title="Last updated on Jun 6, 2019, 3:41 AM UTC (be8eb55)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1417/7f2b6e1...mattto:be8eb55.html" title="Last updated on Jun 6, 2019, 3:41 AM UTC (be8eb55)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2934,7 +2934,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           The |task| *must* use |activeWorker|'s [=event loop=] and the [=handle functional event task source=].
 
       1. Wait for |task| to have executed or been discarded.
-      1. If |registration| is [=stale=], then run the [=Soft Update=] algorithm with |registration|.
+      1. If |registration| is [=stale=], then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
 
       <div class="example">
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -2926,7 +2934,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           The |task| *must* use |activeWorker|'s [=event loop=] and the [=handle functional event task source=].
 
       1. Wait for |task| to have executed or been discarded.
-      1. If the time difference in seconds calculated by the current time minus |registration|'s [=last update check time=] is greater than 86400, invoke [=Soft Update=] algorithm with |registration|.
+      1. If |registration| is [=stale=], then run the [=Soft Update=] algorithm with |registration|.
```
jakearchibald: Should this also be in parallel?

mattto: Done. I noticed this too and thought I'd just keep the existing behavior, but yea it looks right to do it in parallel. I'm also wondering if there's a point to "Wait for |task| to have executed..." but will leave it for now.

mattto:
Thanks, also propagated changes to v1.

Issue #1403, opened 2019-05-02
Missing task queuing


Ms2ger:
In https://w3c.github.io/ServiceWorker/#run-service-worker-algorithm , step 4.13 says

> If callbackSteps is provided, run them with evaluationStatus on the original thread that invoked this algorithm, while continuing in parallel with these steps.

That's not really something that can work. Anne suggests queuing a task instead.

mkruisselbrink:
It's not entirely clear to me why that can't work? It sort of reads to me like the calling thread is supposed to be "blocked" while steps 4.1 - 4.12 are executed. Only after 4.13 is there a "continue in parallel", so before that everything is just sequenced?

annevk:
How can you run something on the original thread without scheduling that run in some way?

mkruisselbrink:
The way I read it the original thread is just waiting around when we reach step 4, until that reaches step 4.13. So there is nothing to schedule, since the thread is just waiting.

But I do agree that this does seem unclear. Also "Run Service Worker" isn't consistently used the same way. Definitely for the cases where Run Service Worker is invoked synchronously from idl methods, as for example in post message, blocking the calling thread until the worker script has been evaluated doesn't make sense, so at least for those cases (ones where no callback steps are passed in to run service worker) my interpretation doesn't make sense.

And since afaict callback steps actually only seemed to be used in one place (in the Update algorithm), some kind of scheduling probably makes sense. But I'm not sure what task queue tasks would get posted to. It's not like the "update" algorithm is running on any particular task queue (actually I have no idea where the "Run Job" algorithm is posting tasks to. All these tasks really aren't associated with any particular agent, but afaict event loops are always associated to agents?

annevk:
Using HTML's parallel queue might make sense in such cases.

But also, HTML creates these kind of threads in parallel, not synchronously from, e.g., `postMessage()`. That seems wrongish?

mattto:
Yes my intent was that Update would block until the callback is run with the result of starting the service worker.

For postMessage(), doesn't it still make sense to wait for the service worker to have started before continuing? I.e., the service worker's global object must be set for the postMessage() steps to work.

I'm guessing we want Run Service Worker to block until the service worker starts (just before "Run the responsible event loop" is called) and return the evaluation status. Is there a rigorous way to specify that?

Something like:

Run Service Worker:

3. Let `evaluationStatus` be null.
[...]

Issue #1146, opened 2017-05-18
"Unregister" operation may be interrupted by `ServiceWorker#postMessage`


jugglinmike:
Consider the following code:

```js
// Precondition: `registration`is a valid ServiceWorkerRegistration object
// whose "active" worker has no pending events
registration.unregister()
  .then(function() {
      registration.active.postMessage('');
    });
```

"unregistration" involves the following sequence:

1. **Queue a task** to resolve the promise (via Resolve Job Promise)
2. If the active worker has no pending events (via Try Clear Registration)
   1. Terminate the active worker (via Clear Registration)
      1. Set the worker's "closing" flag to `true`
   2. **Queue a task** to set the `state` attribute of the active worker to
      "redundant" (via Update Worker State)
   3. **Queue a task** to set the registration's `active` property to `null`
      (via Update Registration State)

This means the `state` attribute is not set until *after* the promise has been
resolved and any subsequent microtasks processed. Here, the `postMessage`
invocation is part of such a microtask, so when it references the worker's
`state`, it receives the value "active". Execution of the `postMessage`
algorithm proceeds to step 4, where Run Service Worker is invoked. This creates
a new ServiceWorkerGlobalScope object, effectively undoing the prior invocation
of Terminate Worker.

Chromium's behavior more or less adheres to this interpretation, although the
worker's script is only evaluated one time (I would expect a second evaluation
for that invocation of Run Service Worker). Firefox throws an error named
`NS_ERROR_FAILURE` from the `postMessage` invocation. In either case, this
seems like a spec bug--the "unregister" operation is being interrupted, but the
worker state is still ultimately set to "redundant."

Distinguishing between "user-requested un-registration" and "user-agent
requested termination" seems necessary in order to decide whether to throw or
re-start the worker. Making this determination may not be possible from an
arbitrary `postMessage` invocation without the introduction of additional
state or a modification to the schedule of operations in Clear Registration.

jungkees:
Thanks for reviewing to this detail. [Unregister](https://w3c.github.io/ServiceWorker/#unregister) algorithm being invoked means the update job is scheduled and run. So, we set the [uninstalling flag](https://w3c.github.io/ServiceWorker/#dfn-uninstalling-flag) to not allow getting this registration or running any action on it from this point. (Only a scheduled register job can flip this.) Also, we try to clear registration at the earliest possible (via Try Clear Registration in many points where the registration's workers are idle.)

[...]

PR #1419, opened 2019-06-06
Check for Run Service Worker failure and disallow starting redundant workers.

mattto:
This cleans up Run Service Worker to block until the worker is running or fails to start. If running, the Completion record is stashed in a new variable. Callsites are updated to check if the worker is running before proceeding. It also disallows starting a redundant worker. postMessage is dropped silently if starting the service worker failed.

Addresses #1146 and #1403.

***
<a href="https://pr-preview.s3.amazonaws.com/mattto/ServiceWorker/pull/1419.html" title="Last updated on Jun 10, 2019, 1:35 AM UTC (b729b47)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1419/d65ed67...mattto:b729b47.html" title="Last updated on Jun 10, 2019, 1:35 AM UTC (b729b47)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -175,7 +175,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A [=/service worker=] has an associated <dfn export id="dfn-set-of-extended-events">set of extended events</dfn> (a [=ordered set|set=]) whose [=list/item=] is an {{ExtendableEvent}}. It is initially a new [=ordered set|set=].
 
-    A [=/service worker=] is said to be <dfn>running</dfn> if its [=event loop=] is running and its [=service worker/global object=]'s [=WorkerGlobalScope/closing=] flag is false.
+    A [=/service worker=] has an associated <dfn>start status</dfn> which can be null or a [=Completion=]. It is initially null.
+
+    A [=/service worker=] is said to be <dfn>running</dfn> if its [=event loop=] is running.
 
     <section>
       <h4 id="service-worker-lifetime">Lifetime</h4>
@@ -390,7 +392,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       The <dfn method for="ServiceWorker"><code>postMessage(|message|, |options|)</code></dfn> method *must* run these steps:
 
-        1. If the {{ServiceWorker/state}} attribute value of the <a>context object</a> is {{"redundant"}}, [=throw=] an "{{InvalidStateError}}" {{DOMException}}.
         1. Let |serviceWorker| be the [=/service worker=] represented by the <a>context object</a>.
         1. If the result of running the [=Should Skip Event=] algorithm with "message" and |serviceWorker|, is true, then return.
         1. Run these substeps [=in parallel=]:
@@ -2585,8 +2586,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Append |url| to |worker|'s [=set of used scripts=].
       1. Set |worker|'s <a>script resource</a>'s <a>HTTPS state</a> to |httpsState|.
       1. Set |worker|'s <a>script resource</a>'s [=script resource/referrer policy=] to |referrerPolicy|.
-      1. Set |evaluationStatus| to the result of running the [=Run Service Worker=] algorithm with |worker| and the *force bypass cache for importscripts flag* set if |job|'s [=job/force bypass cache flag=] is set.
-      1. If |evaluationStatus| is an [=abrupt completion=] or |evaluationStatus|.\[[Value]] is empty, then:
+      1. Run the [=Run Service Worker=] algorithm with |worker| and the *force bypass cache for importscripts flag* set if |job|'s [=job/force bypass cache flag=] is set.
+      1. If |worker| is not [=running=] or its [=start status=] is an [=abrupt completion=], then:
           1. Invoke [=Reject Job Promise=] with |job| and `TypeError`.
           1. If |newestWorker| is null, invoke [=Clear Registration=] algorithm passing |registration| as its argument.
           1. Invoke [=Finish Job=] with |job|.
@@ -2649,7 +2650,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                       1. <span id="install-settle-step">Wait until |e| is not [=ExtendableEvent/active=].</span>
                       1. If |e|'s [=ExtendableEvent/timed out flag=] is set, or the result of [=waiting for all=] of |e|'s [=extend lifetime promises=] rejected, set |installFailed| to true.
 
-                  If |task| is discarded or the script has been aborted by the <a lt="Terminate Service Worker">termination</a> of |installingWorker|, set |installFailed| to true.
+                  If |task| is discarded, set |installFailed| to true.
 
               1. Wait for |task| to have executed or been discarded.
               1. Wait for the step labeled *WaitForAsynchronousExtensions* to complete.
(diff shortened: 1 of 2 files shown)


mattto:
There were too many conflicts trying to patch to v1 so I'll give that a go later.

mattto:
Hm, I think it doesn't actually work to make Run Service Worker blocking because then Terminate Service Worker can never be entered while the worker is starting up.

It seems we have to make everything asynchronous or hand-wave that the user agent can abort starting the worker at any time. Or maybe keep it non-blocking and do something like:


1. Run the "Run Service Worker" algorithm.
1. In parallel:
    1. Wait for service worker to be running or to have been aborted by the Terminate Service Worker algorithm.
    1. If service worker is not running, return.
    1. ... continue with dispatching the event etc.

mattto:
We also need to make Run Service Worker re-entrant :(

mattto:
Ah I may have fallen into a trap. The Terminate Service Worker algorithm probably should be specified to "run the following steps in parallel with the worker's main loop" so it runs on the worker's loop not the main algorithm loop that Run Service Worker is invoked on, and can be entered at any time.

mattto:
OK revised.

jakearchibald:
https://infra.spec.whatwg.org/#algorithm-conditional-abort might be handy if you need to abort steps.

docs/index.bs
```diff
@@ -2717,12 +2725,20 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       : Input
       :: |serviceWorker|, a [=/service worker=]
       :: *force bypass cache for importscripts flag*, an optional flag unset by default
-      :: optional |callbackSteps|, an algorithm which takes a [=Completion=]
 
+      <div class="note">
+        Note: This algorithm typically blocks until the service worker is [=running=]. Most callsites can check for success or failure by checking if the service worker is [=running=] after this algorithm returns.
+
+        This specification generally treats an [=abrupt completion=] due to an uncaught exception the same as a normal completion. That is, if the worker throws an exception during initial script evaluation, it is still considering running and can receive events. Some callsites distinguish between these scenarios by examining the [=start status=] of the [=running=] worker.
```
jakearchibald: Would it be cleaner to return *failure* or the start status from this algorithm?

That means callers could use

1. If the result of invoking [=Run Service Worker=] with |someServiceWorker| is *failure*, then return.

Or:

1. If the result of invoking [=Run Service Worker=] with |someServiceWorker| is *failure* or an [=abrupt completion=], then….

docs/index.bs
```diff
@@ -2717,12 +2725,20 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+      1. If |serviceWorker| is [=running=], then return.
+      1. If |serviceWorker|'s [=state=] is *redundant*, then return.
+      1. Assert: |serviceWorker|'s [=start status=] is null.
       1. Let |script| be |serviceWorker|'s [=service worker/script resource=].
       1. Assert: |script| is not null.
-      1. If |serviceWorker| is already running, this algorithm must have been invoked previously. If |callbackSteps| is provided, run them with the same value as the previous time, and abort these steps.
-      1. Create a separate parallel execution environment (i.e. a separate thread or process or equivalent construct), and run the following substeps in that context:
+      1. Let |startFailed| be false.
+      1. Create a separate parallel execution environment (i.e. a separate thread or process or equivalent construct), and run the following substeps in that context [=in parallel=]:
```
jakearchibald: I'm not sure we need [=in parallel=] here. The steps are already running in "that context" which is a "separate parallel execution environment". Otherwise it could read like you're running steps in that context, then in parallel, which would take you off the JavaScript thread for that context.

jakearchibald: May as well replace "substeps" with "steps".

docs/index.bs
```diff
@@ -2757,9 +2773,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
           1. If |serviceWorker| is an <a>active worker</a>, and there are any <a>tasks</a> queued in |serviceWorker|'s <a>containing service worker registration</a>'s [=service worker registration/task queues=], <a lt="queue a task">queue</a> them to |serviceWorker|'s <a>event loop</a>'s [=/task queues=] in the same order using their original <a>task sources</a>.
           1. Let |evaluationStatus| be the result of <a lt="run a classic script">running the classic script</a> |script| if |script| is a <a>classic script</a>, otherwise, the result of <a lt="run a module script">running the module script</a> |script| if |script| is a [=module script=].
-
-              Note: In addition to the usual possibilities of returning a value or failing due to an exception, this could be prematurely aborted by the <a>terminate service worker</a> algorithm.
-          1. If |callbackSteps| is provided, run them with |evaluationStatus| on the original thread that invoked this algorithm, while continuing in parallel with these steps.
+          1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort the rest of these steps.
```
jakearchibald: I think "abort these steps" is fine.

docs/index.bs
```diff
@@ -2780,14 +2797,16 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       :: None
 
       1. If |serviceWorker| is not running, abort these steps.
```
jakearchibald: Link to your "running" definition?

docs/index.bs
```diff
@@ -2857,8 +2876,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
           1. Return null.
       1. If |activeWorker|'s <a>state</a> is *activating*, wait for |activeWorker|'s <a>state</a> to become *activated*.
-      1. Invoke <a>Run Service Worker</a> algorithm with |activeWorker| as the argument.
-      1. <a>Queue a task</a> |task| to run the following substeps:
+      1. Run the [=Run Service Worker=] algorithm with |activeWorker|.
```
jakearchibald: Do you prefer "Run the" to "Invoke the"? Happy to go with "Run the".

docs/index.bs
```diff
@@ -2921,7 +2941,10 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. If |registration| is [=stale=], then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
           2. Return.
       1. If |activeWorker|'s [=state=] is *activating*, wait for |activeWorker|'s [=state=] to become *activated*.
-      1. Invoke [=Run Service Worker=] algorithm with |activeWorker| as the argument.
+      1. Run the [=Run Service Worker=] algorithm with |activeWorker|.
+      1. If |activeWorker| is not [=running=], then:
+          1. If |registration| is [=stale=], then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
+          2. Return.
```
jakearchibald: Ohh good catch!

docs/index.bs
mattto: Oh yes that would be much simpler... I'll do that.

docs/index.bs
mattto: Agreed, it's redundant, I remember thinking about this and somehow decided to put it in. I'll remove.

docs/index.bs
mattto: I wasn't sure what to do here. My definition doesn't work because it will abort the steps if the worker is starting up and hasn't started its event loop. On the other hand, the "running" used in the existing spec isn't defined. I wonder if we can just delete this step? The HTML equivalent doesn't have it:
https://html.spec.whatwg.org/multipage/workers.html#terminate-a-worker

Or else say something like, "If the steps of the "Create a separate parallel execution environment" step in the Run Service Worker algorithm are not being executed, abort these steps".  ?

docs/index.bs
mattto: Ah, meant to bring this up earlier. There's some inconsistency around "Invoke" vs "Run" and also "with xx, yy, and zz" vs "passing xx, yy, and zz as arguments" and other variations. I planned to just always go with the shortest which is "Run the AA algorithm with xx, yy, and zz." But not sure if there is a style guideline here.

I guess with Run in the name of algorithm, it looks a bit awkward.

docs/index.bs
jakearchibald: Deleting the step seems best I think

docs/index.bs
jakearchibald: It probably doesn't matter much. In other specs I avoid it by making the link text an action, like "Let foo be the result of getting a service worker object…". But I guess we should keep stuff in the algorithm section more like a function.

"Run" or "invoke" seems fine here. No point trying to solve the inconsistencies in this PR.

mattto:
Thanks, addressed all the comments I think. I'll look at backporting to v1 next time.

mattto:
Hi Jake, I made a couple changes since last time:
- For postMessage, moved getting incumbentSettings and serializing (which possibly rethrows the exception) out of the parallel steps, since I don't think you can do those in parallel.
- Backported to v1 which needed some work since the "callbackSteps" part of Run Service Worker which this replaces was never in v1 to begin with.

I don't see how to compile v1 so I'm not sure it works, but looking at https://www.w3.org/TR/service-workers-1 it hasn't been built since 2017 (?).

I'll go ahead and merge this but if you see something wrong we can fix it.

PR #1424, opened 2019-06-07
Fixing max scope to include trailing slash.

### LINKED ISSUE #1423: maxScopeString definition is missing the trailing '/'
The spec has:

> Set maxScopeString to "/" concatenated with the strings, except the last string that denotes the script’s file name, in job’s script url's path (including empty strings), separated from each other by "/".

This seems to omit the trailing "/". Therefore a service worker in /foo/bar/sw.js gets maxScopeString of /foo/bar instead of the intended /foo/bar/.

jakearchibald:
Fixes #1423

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1424.html" title="Last updated on Jun 7, 2019, 8:36 AM UTC (2f3e32c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1424/4a50271...2f3e32c.html" title="Last updated on Jun 7, 2019, 8:36 AM UTC (2f3e32c)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2502,12 +2502,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Let |scopeURL| be |registration|'s [=service worker registration/scope url=].
           1. Let |maxScopeString| be null.
           1. If |serviceWorkerAllowed| is null, then:
-              1. Set |maxScopeString| to "<code>/</code>" concatenated with the strings, except the last string that denotes the script's file name, in |job|'s [=job/script url=]'s [=url/path=] (including empty strings), separated from each other by "<code>/</code>".
+              1. Let |resolvedScope| be the result of [=URL parser|parsing=] "`./`" with a [=base URL=] of |job|'s [=job/script url=].
+              1. Set |maxScopeString| to "`/`", followed by the strings in |resolvedScope|'s [=url/path=] (including empty strings), separated from each other by "`/`".
           1. Else:
               1. Let |maxScope| be the result of <a lt="URL parser">parsing</a> |serviceWorkerAllowed| with |job|'s [=job/script url=].
               1. If |maxScope|'s [=url/origin=] is |job|'s [=job/script url=]'s [=url/origin=], then:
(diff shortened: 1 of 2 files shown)


jakearchibald:
TODO: Drop a note about how this works with the trailing slash

docs/index.bs
```diff
@@ -2502,12 +2502,13 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Let |scopeURL| be |registration|'s [=service worker registration/scope url=].
           1. Let |maxScopeString| be null.
           1. If |serviceWorkerAllowed| is null, then:
-              1. Set |maxScopeString| to "<code>/</code>" concatenated with the strings, except the last string that denotes the script's file name, in |job|'s [=job/script url=]'s [=url/path=] (including empty strings), separated from each other by "<code>/</code>".
+              1. Let |resolvedScope| be the result of [=URL parser|parsing=] "`./`" with a [=base URL=] of |job|'s [=job/script url=].
+              1. Set |maxScopeString| to "`/`", followed by the strings in |resolvedScope|'s [=url/path=] (including empty strings), separated from each other by "`/`".
           1. Else:
-              1. Let |maxScope| be the result of <a lt="URL parser">parsing</a> |serviceWorkerAllowed| with |job|'s [=job/script url=].
+              1. Let |maxScope| be the result of [=URL parser|parsing=] |serviceWorkerAllowed| with a [=base URL=] of |job|'s [=job/script url=].
```
mattto: optional nit: "job's script url's base URL" reads nicer to me. If it's "a base URL" it feels like there's multiple possible base URLs.

jakearchibald: In this case "a base URL" refers to the argument of "URL parser" https://url.spec.whatwg.org/#concept-url-parser.

I could omit it, so it's "Let |maxScope| be the result of [=URL parser|parsing=] |serviceWorkerAllowed| with |job|'s [=job/script url=]."

Or I could be more specific, like "Let |maxScope| be the result of [=URL parser|parsing=] |serviceWorkerAllowed| using |job|'s [=job/script url=] as the [=base URL=]."

Any preference?

mattto: I totally missed that, sorry. I would prefer: "Let |maxScope| be the result of [=URL parser|parsing=] |serviceWorkerAllowed| using |job|'s [=job/script url=] as the [=base URL=]."

PR #1426, opened 2019-06-07
Adding self.serviceWorker to service workers.

### LINKED ISSUE #1077: Find a way for service workers to get their current state
A service worker instance currently has no way to get it's current lifecycle state. There's a bit of a hack one can do where you keep track of the current event as a global variable but that's error prone and doesn't capture the waiting periods correctly.

Imo, the best way to solve this would be to just allow a service worker to get it's ServiceWorker object somehow. Maybe via `self.worker` or `self.self` or something like that.

jakearchibald:
Fixes #1077.

Tests https://github.com/web-platform-tests/wpt/pull/17285

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1426.html" title="Last updated on Jun 12, 2019, 9:14 AM UTC (ad80aea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1426/7e17924...ad80aea.html" title="Last updated on Jun 12, 2019, 9:14 AM UTC (ad80aea)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -926,6 +926,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       interface ServiceWorkerGlobalScope : WorkerGlobalScope {
         [SameObject] readonly attribute Clients clients;
         [SameObject] readonly attribute ServiceWorkerRegistration registration;
+        [SameObject] readonly attribute ServiceWorker serviceWorker;
 
         [NewObject] Promise&lt;void&gt; skipWaiting();
 
@@ -952,6 +953,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       <h4 id="service-worker-global-scope-registration">{{ServiceWorkerGlobalScope/registration}}</h4>
 
       The <dfn attribute for="ServiceWorkerGlobalScope"><code>registration</code></dfn> attribute *must* return the {{ServiceWorkerRegistration}} object that represents the [=ServiceWorkerGlobalScope/service worker=]'s <a>containing service worker registration</a>.
+    </section>
+
+    <section>
+      <h4 id="service-worker-global-scope-serviceworker">{{ServiceWorkerGlobalScope/serviceWorker}}</h4>
+
+      The <dfn attribute for="ServiceWorkerGlobalScope"><code>serviceWorker</code></dfn> attribute *must* return the result of [=getting the service worker object=] that represents the [=context object=]'s [=ServiceWorkerGlobalScope/service worker=] in the [=context object=]'s [=relevant settings object=].
     </section>
 
     <section algorithm="service-worker-global-scope-skipwaiting">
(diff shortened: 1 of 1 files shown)


docs/index.bs
```diff
@@ -1084,7 +1106,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
             1. Let |destination| be the {{ServiceWorkerContainer}} object whose associated [=ServiceWorkerContainer/service worker client=] is |targetClient|.
             1. Add a [=task=] that runs the following steps to |destination|'s [=ServiceWorkerContainer/client message queue=]:
                 1. Let |origin| be the [=Unicode serialization of an origin|Unicode serialization=] of |sourceSettings|'s [=environment settings object/origin=].
-                1. Let |source| be a {{ServiceWorker}} object, which represents the [=ServiceWorkerGlobalScope/service worker=] associated with |sourceSettings|'s [=environment settings object/global object=].
+                1. Let |source| be the result of [=getting the service worker object=] that represents |sourceSettings|'s [=environment settings object/global object=]'s [=ServiceWorkerGlobalScope/service worker=] in |targetClient|.
```
mattto: I'm still wrapping my head around Realms, settings objects, etc, so a naive question: is there a more direct way to get to the service worker? It looks like we're doing [=context object=]'s [=relevant settings object=]'s [=global object=]'s service worker. Is it equivalent to do [=context object=]'s [=relevant global object=]'s service worker, or is that something else?

jakearchibald: Yeah, that's fair, I'll switch to that. I think it means adding `1. Let |contextObject| be the [=context object=].` at the start, since I don't think you can access the context object once you're in parallel.

jakearchibald:
> if it's not too much of a hassle, it might also be worth splitting this into two PRs: one with the "getting the service worker object" algorithm and one that adds self.serviceWorker

Good point. I'll split it up.

jakearchibald:
Ok, this is only about `self.serviceWorker` now. It'll fail to build properly because "getting the service worker object" isn't defined.

jakearchibald:
Assuming we want `self.serviceWorker` to always be available, during the very first run of a service worker it won't yet be in the 'installing' state. Looks like we're going to expose the 'parsed' state for the first time.

jakearchibald:
Tests https://github.com/web-platform-tests/wpt/pull/17285

mattto:
Seems OK to expose "parsed". I'm a bit curious now how that name was chosen. Is it better than something like "initial"? Really naive question but is JavaScript specified to do parsing before starting execution?

jakearchibald:
> Seems OK to expose "parsed". I'm a bit curious now how that name was chosen. Is it better than something like "initial"?

I never thought about changing the name. "initial" seems ok. It could also be "pending", "pendingInstall", "beforeInstall", "preinstall".

> Really naive question but is JavaScript specified to do parsing before starting execution?

I think there are cases where v8 defers parsing parts of some functions, but the top level has been parsed at this point.

wanderview:
Right, if there was a syntax error you would not have gotten to the point of executing the script.

jakearchibald:
I think I'm favouring "preinstall". Anyone hating that?

jakearchibald:
Let's stick with "parsed". It's past tense like "installed" & "activated".

PR #1440, opened 2019-06-23
Create a Kill Service Workers algorithm

asakusuma:
Intended to be a killswitch mechanism for all service workers associated with an origin. In the short term, designed to be used directly by Clear Site Data.

1. "Unclaim" all service workers for the origin. In other words, ensure no clients are controlled by any service worker under the origin
2. Terminate all in-flight tasks, but allow fetch tasks to fallback to network
3. Immediately unregister all registrations under the origin

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2819,6 +2819,52 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Return |serviceWorker|'s [=start status=].
   </section>
 
+    <section algorithm>
+    <h3 id="terminate-task-queue-algorithm"><dfn>Terminate Task Queue</dfn></h3>
+      : Input
+      :: |event loop|, an <a>event loop</a>
+      : Output
+      :: None
+
+      1. For each |task| in |event loop|'s corresponding [=/task queues=]
+        1. If |task|'s <a>task source</a> is <a>handle fetch task source</a>
+            1. Let |client| be |task|'s |request|'s [=request/client=].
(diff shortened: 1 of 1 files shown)


jakearchibald:
I'm digging into this. It feels like we need to define what happens when various algorithms are cancelled halfway through. Eg, what happens if a service worker is killed half way through install, fetch, etc.

asakusuma:
I agree that we need to define what happens to in-flight tasks. However, for the purpose of a killswitch, it would be nice if that definition was centralized/global, rather than relying on the individual design of each algorithm. Is it possible to just drop every task on the floor, have fetches fallback to network, nuke a top level state container, and let GC take over?

asakusuma:
TODO:
- [ ] Either specify a time limit for pending promises, or update [wait until](https://www.w3.org/TR/service-workers-1/#wait-until-method) language to allow killing pending promises without a grace period
- [ ] [Close](https://notifications.spec.whatwg.org/#closing-a-notification) any notifications associated with service workers

jakearchibald:
We'd also need to do something about ongoing [background fetches](https://wicg.github.io/background-fetch/).

This is why I think we need to spec this a little more explicitly rather than try and hand-wave it.

However, these aren't problems created by Clear Site Data. We just haven't been clear about how registrations are cleaned up so far.

asakusuma:
Makes sense. I've added the background fetches to the TODO.

I'm hoping we can apply the [fetch group concept](https://github.com/w3c/ServiceWorker/pull/1443#issuecomment-506665817) and do something similar for service workers.

PR #1444, opened 2019-06-25
Handle asynchronous modules (top-level await).

### LINKED ISSUE #1407: Top-level await integration for ServiceWorkers running modules
The [top-level await proposal](https://github.com/tc39/proposal-top-level-await) would make module evaluation result in a Promise, rather than synchronous success or failure. This would affect the [Update](https://w3c.github.io/ServiceWorker/#update-algorithm) algorithm, which takes the steps in 16.1 if there is such a failure.

Does anyone have such concerns about this evaluation being asynchronous?

If this seems alright, then I'll make sure we have a PR for ServiceWorker after top-level await reaches Stage 3. It looks like there would be a little bit of refactoring on the SW side, so I'll wait for things to settle down editorially on the TC39 side before making more churn with dependent specs.

cc @domenic @guybedford @mylesborins

Ms2ger:
Fixes #1407.

***
<a href="https://pr-preview.s3.amazonaws.com/Ms2ger/ServiceWorker/pull/1444.html" title="Last updated on May 25, 2021, 1:43 PM UTC (51e01b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1444/4b5d041...Ms2ger:51e01b8.html" title="Last updated on May 25, 2021, 1:43 PM UTC (51e01b8)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -51,6 +51,8 @@ spec: push; urlPrefix: https://w3c.github.io/push-api/
         text: push; url: h-the-push-event
 
 spec: ecma-262; urlPrefix: http://tc39.github.io/ecma262/
+    type: abstract-op
+        text: NormalCompletion; url: sec-normalcompletion
     type: dfn
         text: Assert; url: sec-algorithm-conventions
         text: [[Call]]; url: sec-ecmascript-function-objects-call-thisargument-argumentslist
@@ -64,6 +66,8 @@ spec: ecma-262; urlPrefix: http://tc39.github.io/ecma262/
         text: execution context; url: sec-execution-contexts
         text: abrupt completion; url: sec-completion-record-specification-type
         text: Completion; url: sec-completion-record-specification-type
+        text: Module Record; url: sec-abstract-module-records
+        text: Cyclic Module Record; url: sec-cyclic-module-records
 
 spec: page-visibility; urlPrefix: https://www.w3.org/TR/page-visibility/
     type: enum; text: VisibilityState; url: VisibilityState
@@ -83,6 +87,9 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: delay the load event; for: document; url: delay-the-load-event
         urlPrefix: origin.html
             text: creating a policy container from a fetch response
+        urlPrefix: webappapis.html
+            text: module map; url: module-map
+            text: resolve a module specifier; url: resolve-a-module-specifier
 
 spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
     type: dfn
@@ -2710,16 +2717,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                       Note: The control does not break the loop in this step to continue with all the imported scripts to populate the cache.
           1. Asynchronously complete these steps with |response|.
 
-          If the algorithm asynchronously completes with null, then:
+          When the algorithm asynchronously completes, continue the rest of these steps, with |script| being the asynchronous completion value.
 
-              1. Invoke [=Reject Job Promise=] with |job| and `TypeError`.
+      1. If |script| is null or [=Is Async Module=] with |script|'s [=script/record=], |script|'s [=script/base URL=], and « » is true, then:
+          1. Invoke [=Reject Job Promise=] with |job| and `TypeError`.
 
-                  Note: This will do nothing if [=Reject Job Promise=] was previously invoked with "{{SecurityError}}" {{DOMException}}.
+              Note: This will do nothing if [=Reject Job Promise=] was previously invoked with "{{SecurityError}}" {{DOMException}}.
 
-              1. If |newestWorker| is null, then [=map/remove=] [=scope to registration map=][|scopeURL|, [=URL serializer|serialized=]].
-              1. Invoke <a>Finish Job</a> with |job| and abort these steps.
-
-          Else, continue the rest of these steps after the algorithm's asynchronous completion, with |script| being the asynchronous completion value.
+          1. If |newestWorker| is null, then [=map/remove=] [=scope to registration map=][|scopeURL|, [=URL serializer|serialized=]].
+          1. Invoke [=Finish Job=] with |job| and abort these steps.
       1. If |hasUpdatedResources| is false, then:
           1. Set |registration|'s [=service worker registration/update via cache mode=] to |job|'s [=job/update via cache mode=].
           1. Invoke [=Resolve Job Promise=] with |job| and |registration|.
@@ -2940,8 +2946,17 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Create a new {{WorkerLocation}} object and associate it with |workerGlobalScope|.
           1. If the <a>run CSP initialization for a global object</a> algorithm returns "<code>Blocked</code>" when executed upon |workerGlobalScope|, set |startFailed| to true and abort these steps.
           1. If |serviceWorker| is an <a>active worker</a>, and there are any <a>tasks</a> queued in |serviceWorker|'s <a>containing service worker registration</a>'s [=service worker registration/task queues=], <a lt="queue a task">queue</a> them to |serviceWorker|'s <a>event loop</a>'s [=/task queues=] in the same order using their original <a>task sources</a>.
-          1. Let |evaluationStatus| be the result of <a lt="run a classic script">running the classic script</a> |script| if |script| is a <a>classic script</a>, otherwise, the result of <a lt="run a module script">running the module script</a> |script| if |script| is a [=module script=].
-          1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Let |evaluationStatus| be null.
+          1. If |script| is a [=classic script=], then:
+              1. Set |evaluationStatus| to the result of [=run a classic script|running the classic script=] |script|.
+              1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Otherwise, if |script| is a [=module script=], then:
+              1. Let |evaluationPromise| be the result of [=run a module script|running the module script=] |script|, with report errors set to false.
+              1. Assert: |evaluationPromise|.\[[PromiseState]] is not "pending".
+              1. If |evaluationPromise|.\[[PromiseState]] is "rejected":
+                  1. Set |evaluationStatus| to [$ThrowCompletion$](|evaluationPromise|.\[[PromiseResult]]).
+              1. Otherwise:
+                  1. Set |evaluationStatus| to [$NormalCompletion$](undefined).
           1. If the script was aborted by the [=Terminate Service Worker=] algorithm, set |startFailed| to true and abort these steps.
           1. Set |serviceWorker|'s [=start status=] to |evaluationStatus|.
           1. If |script|'s <a>has ever been evaluated flag</a> is unset, then:
@@ -3602,6 +3617,32 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. [=Throw=] the exception.
 
           Note: When an exception is [=throw|thrown=], the implementation does undo (roll back) any changes made to the cache storage during the batch operation job.
+  </section>
+
+  <section algorithm>
+    <h3 id="is-async-module-algorithm"><dfn>Is Async Module</dfn></h3>
+
+      : Input
+      :: |record|, a [=Module Record=]
+      :: |moduleMap|, a [=/module map=]
+      :: |base|, a [=/URL=]
+      :: |seen|, a [=/set=] of [=/URLs=]
+      : Output
+      :: a boolean
+
+      1. If |record| is not a [=Cyclic Module Record=], then:
+          1. Return false.
+      1. If |record|.\[[Async]] is true, then:
+          1. Return true.
+      1. [=list/For each=] string |requested| of |record|.\[[RequestedModules]]:
+          1. Let |url| be the result of [=resolve a module specifier|resolving a module specifier=] given |base| and |requested|.
+          1. If |url| is *failure*:
+              1. Return false.
+          1. If |seen| does not [=set/contain=] |url|, then:
+              1. [=set/Append=] |url| to |seen|.
+              1. If [=Is Async Module=] for |moduleMap|[|url|]'s [=script/record=], |moduleMap|, |base|, and |seen| is true, then:
+                  1. Return true.
+      1. Return false.
   </section>
 </section>
 
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3636,9 +3636,10 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Return true.
       1. [=list/For each=] string |requested| of |record|.\[[RequestedModules]]:
           1. Let |url| be the result of [=resolve a module specifier|resolving a module specifier=] given |base| and |requested|.
+          1. Assert: |url| is never failure, because [=resolve a module specifier|resolving a module specifier=] must have been previously successful with these same two arguments.
           1. If |seen| does not [=set/contain=] |url|, then:
               1. [=set/Append=] |url| to |seen|.
-              1. If |moduleMap| does not [=map/contain=] |url|, then:
+              1. If |moduleMap|[|url|] does not have a [=script/record=], then:
                   1. Return false.
               1. If [=Is Async Module=] for |moduleMap|[|url|]'s [=script/record=], |moduleMap|, |base|, and |seen| is true, then:
                   1. Return true.
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3636,10 +3636,10 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Return true.
       1. [=list/For each=] string |requested| of |record|.\[[RequestedModules]]:
           1. Let |url| be the result of [=resolve a module specifier|resolving a module specifier=] given |base| and |requested|.
-          1. If |url| is *failure*:
-              1. Return false.
           1. If |seen| does not [=set/contain=] |url|, then:
               1. [=set/Append=] |url| to |seen|.
+              1. If |moduleMap| does not [=map/contain=] |url|, then:
+                  1. Return false.
               1. If [=Is Async Module=] for |moduleMap|[|url|]'s [=script/record=], |moduleMap|, |base|, and |seen| is true, then:
                   1. Return true.
       1. Return false.


jakearchibald:
Is top-level await only allowed in module scripts?

Ms2ger:
Yes, that's correct.

docs/index.bs
```diff
@@ -2801,8 +2803,21 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-          1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Let |evaluationStatus| be null.
+          1. If |script| is a <a>classic script</a>, then:
+              1. Set |evaluationStatus| to the result of <a lt="run a classic script">running the classic script</a> |script|.
+              1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Otherwise, if |script| is a [=module script=], then:
+              1. Let |evaluationPromise| be the result of <a lt="run a module script">running the module script</a> |script|, with report errors set to false.
+              1. If |evaluationPromise|.\[[PromiseState]] is "fulfilled":
+                  1. Set |evaluationStatus| to [$NormalCompletion$](|evaluationPromise|.\[[PromiseResult]]).
```
littledan: I'm not sure why we need this synchronous check for "fulfilled"; isn't this path handled just as well by line 2816?

Ms2ger: To set `evaluationStatus`. I haven't quite figured out what that's used for, though.

Ms2ger: Except that the promise can only resolve with undefined, so you're right.

jakearchibald:
> would it be OK to skip the synchronous error check, and just rely on the error being reported a bit later,

I don't think so. When we first install a service worker, if it throws an error during initial execution, we discard it.

annevk:
I'm probably missing something, but how could the timing be observed?

Can you add listeners before an `await`? At what point are they invoked?

jakearchibald:
Here's how I think this works: (@littledan: correct me if I'm wrong)

When we run a service worker, we consider it as 'evaluated' and ready to receive events after synchronous execution + microtasks.

# Errors & evaluation

This error will prevent a service worker from installing:

```js
throw Error('…');
```

As will this:

```js
await undefined;

throw Error('…');
```

As will this:

```js
import './foo.js';

// If `./foo.js` is:
throw Error('…');
```

However, this error will not prevent a service worker from installing:

```js
await new Promise(r => setTimeout(r));

throw Error('…');
```

…as the error has happened after 'evaluation'.

# Adding listeners

```js
// This is fine
addEventListener('fetch', func);

// This is fine
await undefined;
addEventListener('fetch', func);

// This throws
await new Promise(r => setTimeout(r));
addEventListener('fetch', func);
```

However, the throw is asynchronous, so it wouldn't prevent install.

# Dispatching events

We consider the service worker ready to receive events once it's evaluated:

```js
addEventListener('fetch', () => {
  console.log('A');
});

await new Promise(r => setTimeout(r, 5000));
console.log('B');
```

The console may log `A` before `B`.

# Concerns

```js
import './third-party-script-v1.js';

addEventListener('install', event => …);
addEventListener('fetch', event => …);
```

The above works as the developer expects. However, later, they update to `'./third-party-script-v2.js'`. This include a top-level await that waits beyond a microtask.

Now, their service worker updates successfully, but their `install` and `fetch` listeners are absent. They'll show as an error in the console, but the developer may miss them.

annevk:
Thank you! "Concerns" seems pretty bad from a debugging perspective, i.e., action at a distance.

To address @littledan's concern, we could perhaps race the _evaluationPromise_ with another promise, but maybe that introduces issues for the remainder of the algorithm. (I was also thinking of this as a way to allow top-level await, but only for a limited number of ticks.)

This whole setup still seems rather suboptimal for the design of service workers.

jakearchibald:
> I was also thinking of this as a way to allow top-level await, but only for a limited number of ticks

What would we do if it exceeded that number of ticks?

> This whole setup still seems rather suboptimal for the design of service workers.

Yeah, it still feels like we'd have to tell developers "you can do this but oh god please don't do this".

Looking at the PR, we could fail the install of a service worker if its `|evaluationPromise|.\[[PromiseState]]` is "pending". This would effectively prevent top-level `await` in service workers at install time. A top-level await could be introduced later, using something like `if (Date.now() > …)`, in which case we'd behave as above.

annevk:
I was thinking that if you exceed the ticks it's equivalent to an evaluation error. So you're allowed to do some minor `queueMicrotask()`-type stuff, but nothing else.

(At some point we need to have a conversation with TC39 about language extensions and the different types of host environments. Not everything that is great for documents/dedicated workers/servers necessarily makes sense for workets/service workers.)

jakearchibald:
> I was thinking that if you exceed the ticks it's equivalent to an evaluation error. So you're allowed to do some minor `queueMicrotask()`-type stuff, but nothing else.

I thought we already got microtasks for free since they're handled just after "executing as script". The bit I'd be worried about is:

```js
addEventListener('fetch', …);
await usuallyFast();
```

…which means you have a service worker that *usually* works, but sometimes doesn't.

littledan:
Yeah, these are the semantics I was suggesting. Another possibility (based on a suggestion from @annevk) would be to have a per-agent variable which sets whether TLA will currently be allowed. We could make the module link algorithm check this flag, and make the flag disable TLA just while initializing the module, and reenable it later. How does this sound?

Ms2ger:
With the changes in https://github.com/tc39/proposal-top-level-await/pull/128 and https://github.com/whatwg/html/pull/4352, I believe this implements the new semantics.

annevk:
Is there no way to avoid looking at the internals of a promise? If we really need to do that perhaps provide an abstraction in IDL? At the moment TC39 will only alert IDL and HTML of changes to such things (if it happens at all).

docs/index.bs
```diff
@@ -3513,6 +3530,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+      : Input
+      :: |record|, a [=Module Record=]
+      :: |moduleMap|, a [=/module map=]
+      :: |base|, a [=/URL=]
+      :: |seen|, a [=/set=] of [=/URLs=]
+      : Output
+      :: a boolean
+
+      1. If |record| is not a [=Cyclic Module Record=], then:
```
jakearchibald: I'm struggling to figure out what a "Cyclic Module Record" is. What's the difference between a Cyclic Module Record and a non-Cyclic Module Record? How does each relate to a module?

littledan: JavaScript modules (Source Text Module Records) had a bunch of their functionality for traversing the module graph refactored up into Cyclic Module Records, so that WebAssembly modules could share the logic. Not all modules use this infrastructure, for example, JSON modules are leaves so they don't bother.

littledan:
@jakearchibald I think we're missing tests for this as well as https://github.com/whatwg/html/pull/4352 , making it a rather big project. I'd welcome help here!

docs/index.bs
```diff
@@ -2873,8 +2879,19 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-          1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Let |evaluationStatus| be null.
+          1. If |script| is a <a>classic script</a>, then:
+              1. Set |evaluationStatus| to the result of <a lt="run a classic script">running the classic script</a> |script|.
+              1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Otherwise, if |script| is a [=module script=], then:
+              1. Let |evaluationPromise| be the result of <a lt="run a module script">running the module script</a> |script|, with report errors set to false.
+              1. Assert: |evaluationPromise|.\[[PromiseState]] is not "pending".
+              1. If |evaluationPromise|.\[[PromiseState]] is "rejected":
+                  1. Set |startFailed| to true and abort these steps.
```
hiroshige-g, line 2890: I think we should set |evaluationStatus| to an abrupt completion here, i.e. shouldn't set |startFailed| to true, because in classic scripts (and module scripts before this PR) don't set |startFailed| on abrupt completion.

docs/index.bs
```diff
@@ -2645,16 +2652,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                       Note: The control does not break the loop in this step to continue with all the imported scripts to populate the cache.
           1. Asynchronously complete these steps with |response|.
 
-          If the algorithm asynchronously completes with null, then:
+          When the algorithm asynchronously completes, continue the rest of these steps, with |script| being the asynchronous completion value.
 
-              1. Invoke [=Reject Job Promise=] with |job| and `TypeError`.
-
-                  Note: This will do nothing if [=Reject Job Promise=] was previously invoked with "{{SecurityError}}" {{DOMException}}.
+      1. If |script| is null or [=Is Async Module=] with |script|'s [=script/record=], |script|'s [=script/base URL=], and « » is true, then:
```
hiroshige-g, line 2657: A random idea: How about checking whether `|evaluationPromise|.\[[PromiseState]] is "pending"` instead of `[=Is Async Module=] ... is true`?

This makes the logic a little simpler (using only `|evaluationPromise|.\[[PromiseState]]`, removing module graph traversal and module inspection), but this is not equivalent from the current condition (e.g. for modules with `await undefined`).

littledan: A goal of the design here was to not be "racy" in the sense that, both asynchronous behavior and errors are based on the module syntax, not whether it has already run or not, or whether it dynamically reached an `await`. This would make behavior more stable.

hiroshige-g: Thanks, it makes sense > based on syntax.

docs/index.bs
```diff
@@ -3513,6 +3530,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+    <h3 id="is-async-module-algorithm"><dfn>Is Async Module</dfn></h3>
+
+      : Input
+      :: |record|, a [=Module Record=]
+      :: |moduleMap|, a [=/module map=]
+      :: |base|, a [=/URL=]
+      :: |seen|, a [=/set=] of [=/URLs=]
+      : Output
+      :: a boolean
+
```
hiroshige-g, line 3544: > If |record| is not a [=Cyclic Module Record=]

Does this include the case of |record| == `null`?

jakearchibald:
@mfalken could you take a look at this too?

docs/index.bs
```diff
@@ -2645,16 +2652,15 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      1. If |script| is null or [=Is Async Module=] with |script|'s [=script/record=], |script|'s [=script/base URL=], and « » is true, then:
+          1. Invoke [=Reject Job Promise=] with |job| and `TypeError`.
 
-              1. If |newestWorker| is null, then [=map/remove=] [=scope to registration map=][|scopeURL|, [=URL serializer|serialized=]].
-              1. Invoke <a>Finish Job</a> with |job| and abort these steps.
+              Note: This will do nothing if [=Reject Job Promise=] was previously invoked with "{{SecurityError}}" {{DOMException}}.
 
-          Else, continue the rest of these steps after the algorithm's asynchronous completion, with |script| being the asynchronous completion value.
+          1. If |newestWorker| is null, then [=map/remove=] [=scope to registration map=][|scopeURL|, [=URL serializer|serialized=]].
+          1. Invoke <a>Finish Job</a> with |job| and abort these steps.
```
jakearchibald, line 2663:
```suggestion
          1. Invoke [=Finish Job=] with |job| and abort these steps.
```

docs/index.bs
```diff
@@ -2873,8 +2879,19 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Set |workerGlobalScope|'s [=ServiceWorkerGlobalScope/force bypass cache for import scripts flag=] if |forceBypassCache| is true.
           1. Create a new {{WorkerLocation}} object and associate it with |workerGlobalScope|.
           1. If |serviceWorker| is an <a>active worker</a>, and there are any <a>tasks</a> queued in |serviceWorker|'s <a>containing service worker registration</a>'s [=service worker registration/task queues=], <a lt="queue a task">queue</a> them to |serviceWorker|'s <a>event loop</a>'s [=/task queues=] in the same order using their original <a>task sources</a>.
-          1. Let |evaluationStatus| be the result of <a lt="run a classic script">running the classic script</a> |script| if |script| is a <a>classic script</a>, otherwise, the result of <a lt="run a module script">running the module script</a> |script| if |script| is a [=module script=].
-          1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Let |evaluationStatus| be null.
+          1. If |script| is a <a>classic script</a>, then:
```
jakearchibald, line 2883:
```suggestion
          1. If |script| is a [=classic script=], then:
```

docs/index.bs
```diff
@@ -2873,8 +2879,19 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+              1. Set |evaluationStatus| to the result of <a lt="run a classic script">running the classic script</a> |script|.
```
jakearchibald, line 2884:
```suggestion
              1. Set |evaluationStatus| to the result of [=run a classic script|running the classic script=] |script|.
```

docs/index.bs
```diff
@@ -2873,8 +2879,19 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+              1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Otherwise, if |script| is a [=module script=], then:
+              1. Let |evaluationPromise| be the result of <a lt="run a module script">running the module script</a> |script|, with report errors set to false.
```
jakearchibald, line 2887:
```suggestion
              1. Let |evaluationPromise| be the result of [=run a module script|running the module script=] |script|, with report errors set to false.
```

docs/index.bs
```diff
@@ -2873,8 +2879,19 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+              1. Assert: |evaluationPromise|.\[[PromiseState]] is not "pending".
+              1. If |evaluationPromise|.\[[PromiseState]] is "rejected":
+                  1. Set |startFailed| to true and abort these steps.
+
+                      Note: This is intended to maintain compatibilty with failing synchronous modules before the introduction of asynchronous modules.
+              1. Otherwise:
+                  1. Set |evaluationStatus| to [$NormalCompletion$](undefined).
```
jakearchibald, line 2894: `undefined`?

jakearchibald (changes requested):
Looks good aside from the `undefined`. Also, are there any tests?

docs/index.bs
Ms2ger: I don't understand this comment.

Ms2ger (commented):
I don't know off-hand what the testing situation is. Can check later.

docs/index.bs
jakearchibald: Ignore me, I misinterpreted this as a markdown link.

jakearchibald (approved):

Ms2ger:
A test was added at https://github.com/web-platform-tests/wpt/pull/26019; thanks @codehag for pointing it out to me. Let me know if you want me to look into adding more tests.

docs/index.bs
```diff
@@ -3513,6 +3530,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      :: |base|, a [=/URL=]
+      :: |seen|, a [=/set=] of [=/URLs=]
+      : Output
+      :: a boolean
+
+      1. If |record| is not a [=Cyclic Module Record=], then:
+          1. Return false.
+      1. If |record|.\[[Async]] is true, then:
+          1. Return true.
+      1. [=list/For each=] string |requested| of |record|.\[[RequestedModules]],
```
mfalken, line 3549: nit: use a colon instead of a comma at the end for consistency

docs/index.bs
```diff
@@ -3513,6 +3530,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+          1. Let |url| be the result of [=resolve a module specifier|resolving a module specifier=] given |base| and |requested|.
+          1. Assert: |url| is never failure, because [=resolve a module specifier|resolving a module specifier=] must have been previously successful with these same two arguments.
+          1. If |seen| does not [=set/contain=] |url|, then:
+              1. [=set/Append=] |url| to |seen|.
+              1. If [=Is Async Module=] for |moduleMap|[|url|]'s [=script/record=], |moduleMap|, |base|, and |seen| is true, then
```
mfalken, line 3554: nit: colon at end

mfalken (commented):
Sorry for the delay! I needed some time to understand this.

Can the commit description elaborate more on what the change is (the bug thread is very long). It sounds like this is implementing https://github.com/w3c/ServiceWorker/issues/1407#issuecomment-531673790, is that right?

docs/index.bs
```diff
@@ -3513,6 +3530,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      : Output
+      :: a boolean
+
+      1. If |record| is not a [=Cyclic Module Record=], then:
+          1. Return false.
+      1. If |record|.\[[Async]] is true, then:
+          1. Return true.
+      1. [=list/For each=] string |requested| of |record|.\[[RequestedModules]],
+          1. Let |url| be the result of [=resolve a module specifier|resolving a module specifier=] given |base| and |requested|.
+          1. Assert: |url| is never failure, because [=resolve a module specifier|resolving a module specifier=] must have been previously successful with these same two arguments.
```
syg, line 3551: During implementation in Chrome we realized that linking failures means this assert can trip. If the module graph couldn't be successfully linked, we can't assume the entire module graph is accessible this way.

jakearchibald: I'm not really familiar with the module loading stuff. Any idea what spec change is needed here?

codehag:
I think we addressed the open issues, should we rebase?

docs/index.bs
syg: I don't grok the module loading stuff enough myself to be confident with any suggested fix. I don't think resolving a module specifier itself can fail, but that a module graph could have failed to Link(). In which case a requested module's url may not map to any module record in the module map.

Is there an observable error timing currently for a "SW has a TLA" error vs a linking error? My in-flight [CL](https://chromium-review.googlesource.com/c/chromium/src/+/2903690) in Chrome chooses to only check if a module graph has any TLAs in it if it successfully linked. If it failed to link, it doesn't check if it's async at all and let the linking error propagate.

docs/index.bs
```diff
@@ -3513,6 +3528,32 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      : Output
+      :: a boolean
+
+      1. If |record| is not a [=Cyclic Module Record=], then:
+          1. Return false.
+      1. If |record|.\[[Async]] is true, then:
+          1. Return true.
+      1. [=list/For each=] string |requested| of |record|.\[[RequestedModules]]:
+          1. Let |url| be the result of [=resolve a module specifier|resolving a module specifier=] given |base| and |requested|.
+          1. If |url| is *failure*:
```
syg, line 3549: I don't know if this can actually fail (I'm inclined to believe no, it can't).

docs/index.bs
```diff
@@ -3513,6 +3528,32 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+              1. Return false.
+          1. If |seen| does not [=set/contain=] |url|, then:
+              1. [=set/Append=] |url| to |seen|.
+              1. If [=Is Async Module=] for |moduleMap|[|url|]'s [=script/record=], |moduleMap|, |base|, and |seen| is true, then:
```
syg, line 3553: The spec bug I was pointing out in https://github.com/w3c/ServiceWorker/pull/1444#discussion_r637051733 is here, moduleMap[url] might not have a module record in case of link failure.

Ms2ger: Tried to fix this.

codehag: I think we also want to bring the assertion back: https://github.com/Ms2ger/ServiceWorker/pull/2/commits/6290a2c783987f79eed665db40481815a0e6501c

docs/index.bs
```diff
@@ -2940,8 +2946,17 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-          1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Let |evaluationStatus| be null.
+          1. If |script| is a [=classic script=], then:
+              1. Set |evaluationStatus| to the result of [=run a classic script|running the classic script=] |script|.
+              1. If |evaluationStatus|.\[[Value]] is empty, this means the script was not evaluated. Set |startFailed| to true and abort these steps.
+          1. Otherwise, if |script| is a [=module script=], then:
+              1. Let |evaluationPromise| be the result of [=run a module script|running the module script=] |script|, with report errors set to false.
+              1. Assert: |evaluationPromise|.\[[PromiseState]] is not "pending".
+              1. If |evaluationPromise|.\[[PromiseState]] is "rejected":
+                  1. Set |evaluationStatus| to [$ThrowCompletion$](|evaluationPromise|.\[[PromiseResult]]).
```
mfalken, line 2957: ThrowCompletion() does not become a link in the Preview. Is it possible to fix it?

docs/index.bs
mfalken: While reviewing this I was not sure if this comment was resolved. If I understand correctly,  per the next comment thread, the assert failure was fixed in a later commit to this PR that added the following:

> If |moduleMap|[|url|] does not have a [=script/record=], then return false.

mfalken (approved):
I'm very sorry for the long delay. I am not an expert here, but this looks good to me. In particular, my main concern from the previous review about startFailure has been addressed. Thank you.

mfalken:
Actually I'll just merge this and we can fix the link later.

PR #1494, opened 2020-01-03
Rewriting abstract

### LINKED ISSUE #1493: Inaccurate Abstract
"This specification describes a method that enables applications to take advantage of persistent background processing, including hooks to enable bootstrapping of web applications while offline."

It mentions "background" a few more times, but I think in practice they're not so much for background processing as this might make it seem.

I think it'd be better to describe it as being able to handle specific type of events in parallel to the main application, including potentially when the main application is not running. And that last bit only has some agreement for push thus far in practice, which is still visible to the user in some manner as I understand it.

jakearchibald:
Fixes #1493

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/pull/1494.html" title="Last updated on Jan 9, 2020, 12:00 PM UTC (a235a64)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1494/7c23924...a235a64.html" title="Last updated on Jan 9, 2020, 12:00 PM UTC (a235a64)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -13,11 +13,9 @@ Repository: w3c/ServiceWorker
 Group: serviceworkers
 !Tests: <a href=https://github.com/web-platform-tests/wpt/tree/master/service-workers>web-platform-tests service-workers/</a> (<a href=https://github.com/web-platform-tests/wpt/labels/service-workers>ongoing work</a>)
 Status Text: This is a living document. Readers need to be aware that this specification may include unimplemented features, and details that may change. <a href="https://w3c.github.io/ServiceWorker/v1/">Service Workers 1</a> is a version that is advancing toward a W3C Recommendation.
-Abstract: The core of this specification is a [=web worker=] that wakes in order for an [=event=] to be [=dispatched=] on its <a interface lt="ServiceWorkerGlobalScope">global object</a>. This provides an event destination that can be used when other destinations would be inappropriate, or no other destination exists.
+Abstract: The core of this specification is a [=web worker=] that wakes in order for an [=event=] to be [=dispatched=] on its <a interface lt="ServiceWorkerGlobalScope">global object</a>. This provides an <a href="#extensibility">extensible</a> event destination that can be used when other destinations would be inappropriate, or no other destination exists.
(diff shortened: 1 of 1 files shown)


docs/index.bs
```diff
@@ -13,11 +13,13 @@ Repository: w3c/ServiceWorker
...
 Abstract:
-Abstract: The core of this system is an event-driven <a>Web Worker</a>, which responds to events dispatched from documents and other sources. A system for managing installation, versions, and upgrades is provided.
+Abstract: For example, to allow the developer to decide how a page should be fetched, an event needs to dispatch potentially before any other execution contexts exist for that origin. To react to a push message, or completion of a persistent download, the context that originally registered interest may no longer exist. In these cases, the service worker is the ideal event destination.
 Abstract:
-Abstract: The service worker is a generic entry point for event-driven background processing in the Web Platform that is <a href="#extensibility">extensible by other specifications</a>.
+Abstract: The service worker is a generic entry point for event-driven processing that is <a href="#extensibility">extensible by other specifications</a>.
```
annevk: I'd probably drop this bit.

jakearchibald:
@jungkees @slightlyoff @mattto any thoughts about this before I merge?

jungkees:
It seems to me that the first paragraph of the change seems a bit too specific in using terms. I think abstract should just explain what it is simply. In that regard, the original text reads easier.

To @annevk's point, I'll leave a comment in the issue.

mattto:
I'm OK with this PR. Happy to take another look if you incorporate @jungkees's suggestions too.

jakearchibald:
@jungkees 

> It seems to me that the first paragraph of the change seems a bit too specific in using terms. 

Hmm, I was trying to describe what the service worker is before things like the fetch event are added on top, and why it's used for particular event rather than just firing events in . a document.

Eg, I could remove some of the terms & make it:

The core of this specification is a worker that wakes to receive events. This provides an event destination that can be used when other destinations would be inappropriate, or no other destination exists.

Is that better?

Issue #1397, opened 2019-03-28
provide a way to execute work after browser has consumed a response


wanderview:
Recently I've been looking at some code that does something like this:

```javascript
addEventListener('fetch', evt => {
  evt.respondWith(async function() {
    let response = await fancyResponseLoader(evt.request);
    doCompletionWork(response);
    return response;
  }());
});
```

After it has a Response it does some amount of completion work.  This could book keeping, recording metrics, opportunistic caching, etc.

What I have seen in these cases is that the completion work ends up delaying the browser from processing the response.

It might be nice to provide some way for the code to execute something after the browser has processed the response.  A couple options:

  * Make `respondWith()` return a promise for once it has completed its work on the service worker thread.
  * Add a `FetchEvent.complete` promise for once it has completed its work on the SW thread.  This could also support fallback cases where respondWith() is not called at all.

With something like this we would rewrite the example above as:

```javascript
addEventListener('fetch', evt => {
  evt.respondWith(async function() {
    let response = await fancyResponseLoader(evt.request);
    evt.waitUntil(evt.complete.then(_ => doCompletionWork(response)));
    return response;
  }());
});
```

The work around for this issue is to try to use a microtask or task to get your completion work to execute after the browser is done.  This is pretty much guesswork, though, since I don't think we clearly define how browsers process the response on the SW thread.  For example, in chrome we end up internally queuing a microtask after the respondWith promise completes.  So if you want to run your completion work after you need two microtasks to get behind the browser.  And of course other browsers might be different.  Providing an explicit API would avoid this sort of confusion.

asutherland:
So the goal is just to avoid just interrupting the single fetch response consumption, rather than wanting something like https://w3c.github.io/requestidlecallback/ to allow multiple fetches' processing to be deferred until the ServiceWorker reaches an idle point?

(Note that although https://www.chromestatus.com/features/5572795866021888 indicates public support from Mozilla for that API, the link dates to Firefox OS times and there's nothing at https://github.com/mozilla/standards-positions/issues right now so I don't know what the current Mozilla/Firefox position might be, although in principle I think we're on-board with APIs to help sites improve time-to-interactive.)

wanderview:
Right, the completion work may be somewhat timing dependent so we want to avoid using a full task via `setTimeout(f,0)` or an idle callback.  We'd like to still perform the work as close to completion as possible, but *after* the browser has processed the Response.

asutherland:
[...]

PR #1496, opened 2020-01-10
Add FetchEvent.handled (#1397)

tingshao:
***
<a href="https://pr-preview.s3.amazonaws.com/tingshao/ServiceWorker/pull/1496.html" title="Last updated on Jan 22, 2020, 4:37 AM UTC (05328eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1496/910cdd8...tingshao:05328eb.html" title="Last updated on Jan 22, 2020, 4:37 AM UTC (05328eb)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -40,6 +40,9 @@ spec: infra;
         text: list;
         for: set; text: append
         for: list; text: append
+spec: webidl;
+    type: dfn;
+        text: resolve;
 </pre>
 
 <pre class="anchors">
@@ -1484,6 +1487,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         readonly attribute DOMString clientId;
         readonly attribute DOMString resultingClientId;
         readonly attribute DOMString replacesClientId;
+        readonly attribute Promise&lt;void&gt; handled;
 
         void respondWith(Promise&lt;Response&gt; r);
       };
@@ -1495,6 +1499,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         DOMString clientId = "";
         DOMString resultingClientId = "";
         DOMString replacesClientId = "";
+        Promise&lt;void&gt; handled;
       };
     </pre>
 
@@ -1534,6 +1539,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       <h4 id="fetch-event-replacesClientId">{{FetchEvent/replacesClientId|event.replacesClientId}}</h4>
 
       <dfn attribute for="FetchEvent"><code>replacesClientId</code></dfn> attribute *must* return the value it was initialized to. When an <a>event</a> is created the attribute *must* be initialized to the empty string.
+    </section>
+
+    <section>
+      <h4 id="fetch-event-handled">{{FetchEvent/handled|event.handled}}</h4>
+
+      <dfn attribute for="FetchEvent"><code>handled</code></dfn> attribute *must* return the value it was initialized to. When an <a>event</a> is created the attribute *must* be initialized to a pending [=promise=].
     </section>
 
     <section algorithm="fetch-event-respondwith">
@@ -2932,6 +2943,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |client| be |request|'s [=request/client=].
       1. Let |reservedClient| be |request|'s [=request/reserved client=].
       1. Let |preloadResponse| be a new [=promise=].
+      1. Let |workerRealm| be null.
+      1. Let |eventHandled| be null.
       1. Let |fetchInstance| be the instance of the [=/fetch=] algorithm representing the ongoing fetch.
       1. Assert: |request|'s [=request/destination=] is not "<code>serviceworker</code>".
       1. If |request|'s [=request/destination=] is either "<code>embed</code>" or "<code>object</code>", then:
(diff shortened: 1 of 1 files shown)


tingshao:
Does anyone know how to restart the ipr check? 
I got this error when the PR is created because this is my first PR for this repo and I didn't link my github account to w3c account. Then I created my w3c account (I'm from Intel which is a related organization) and linked to the github account. But I can't trigger the ipr check again even a force git push didn't work.
Thanks for help.

tingshao:
@mattto  Do you know how to solve this problem? thanks.

jakearchibald:
> Does anyone know how to restart the ipr check?

Done!

docs/index.bs
```diff
@@ -2932,6 +2940,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |client| be |request|'s [=request/client=].
       1. Let |reservedClient| be |request|'s [=request/reserved client=].
       1. Let |preloadResponse| be a new [=promise=].
+      1. Let |eventHandled| be a new [=promise=].
```
jakearchibald: We're a little inconsistent here because specs were still being figured out when we first wrote this, but the above should use [=a new promise=], which points to https://heycam.github.io/webidl/#a-new-promise.

You should also specify a realm for this, so this line should come just before line 20, when you have the active worker, and therefore "the [=relevant realm=] of the |activeWorker|'s [=service worker/global object=]".

docs/index.bs
```diff
@@ -3007,6 +3017,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           The |task| *must* use |activeWorker|'s <a>event loop</a> and the <a>handle fetch task source</a>.
 
       1. Wait for |task| to have executed or for |handleFetchFailed| to be true.
+      1. If |handleFetchFailed| is false and |eventCanceled| is false, then resolve |eventHandled| with undefined, otherwise reject it with "{{NetworkError}}" {{DOMException}}.
```
jakearchibald: Use [=resolve=] and [=reject=]. You don't need to say "with undefined", that's the default.

"reject it with a" rather than just "reject it with". And again I'd specify the realm.

I'm not sure the condition is right here. It's only an exception if |respondWithEntered| is false and |eventCanceled| is true.

It might be better to handle resolving before each return in the next few lines, then we aren't duplicating the logic and risking them getting out of sync.

tingshao:
> This is heading in the right direction. Are you planning to write web platform tests too?

Thanks for resolving the ipr issue and review. Yes, I plans to write the tests and will send the PR for review when ready.

docs/index.bs
```diff
@@ -2982,35 +2991,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+             The |task| *must* use |activeWorker|'s <a>event loop</a> and the <a>handle fetch task source</a>.
 
       1. Wait for |task| to have executed or for |handleFetchFailed| to be true.
       1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
-      1. If |respondWithEntered| is false, then return a [=network error=] if |eventCanceled| is true and null otherwise.
-      1. If |handleFetchFailed| is true, then return a [=network error=].
-      1. Return |response|.
+      1. If |respondWithEntered| is false, then:
+          1. If |eventCanceled| is true, then [=reject=] |eventHandled| with a "{{NetworkError}}" {{DOMException}} and return a [=network error=].
+          1. [=resolve=] |eventHandled| and return null.
```
jakearchibald:
```suggestion
          1. [=Resolve=] |eventHandled| and return null.
```

docs/index.bs
```diff
@@ -2982,35 +2991,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      1. If |handleFetchFailed| is true, then [=reject=] |eventHandled| with a "{{NetworkError}}" {{DOMException}} and return a [=network error=].
+      1. [=resolve=] |eventHandled| and return |response|.
```
jakearchibald:
```suggestion
      1. [=Resolve=] |eventHandled| and return |response|.
```

docs/index.bs
tingshao: Thanks. I changed to declare |eventHandled| to be null fist in top-level to make it able to be accessed in those last steps. Then set it to a new promise when |activeWorker| is available. Hope this could also cover your idea.

docs/index.bs
tingshao: Done

docs/index.bs
```diff
@@ -2982,35 +2991,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-              1. Else, set |response| to |e|'s [=FetchEvent/potential response=].
-          1. If |e|'s <a>canceled flag</a> is set, set |eventCanceled| to true.
-          1. If |fetchInstance| is [=fetch/terminated=], then [=queue a task=] to [=AbortSignal/signal abort=] on |requestObject|'s {{Request/signal}}.
-
-          If |task| is discarded, set |handleFetchFailed| to true.
-
-          The |task| *must* use |activeWorker|'s <a>event loop</a> and the <a>handle fetch task source</a>.
+      1. Else:
+          1. Let |workerRealm| be the [=relevant realm=] of the |activeWorker|'s [=service worker/global object=].
+          1. Set |eventHandled| to [=a new promise=] in |workerRealm|. In following steps, [=resolve=] or [=reject=] |eventHandled| in |workerRealm| only. If |eventHandled| is null, ignore the [=resolve=] or [=reject=] operations on it.
```
jakearchibald: > If |eventHandled| is null, ignore the [=resolve=] or [=reject=] operations on it.

I'm not sure we should hand-wave like this. Can we just perform a null check when we try to do stuff with this object?

> In following steps, [=resolve=] or [=reject=] |eventHandled| in |workerRealm| only.

I don't think this really works, since it doesn't matter which realm resolves/rejects the promise. The realm only matters for objects.

In this case, you'd just change the instances of

"[=reject=] |eventHandled| with a "{{NetworkError}}" {{DOMException}}"

to

"[=reject=] |eventHandled| with a "{{NetworkError}}" {{DOMException}} **in |workerRealm|**"

It's the error object that gets the realm.

docs/index.bs
tingshao: [=Resolve=] seems compiled to be linked to 'https://w3c.github.io/FileAPI/#blob-url-resolve' in the output. Shall we drop the link by using simple text like "Resolve |eventHandled| and ...".

jakearchibald:
The phrase should link to https://heycam.github.io/webidl/#resolve.

You can tell bikeshed the correct link using
https://github.com/WICG/background-fetch/blob/master/index.bs#L18

tingshao:
Thanks, the new patch resolved all comments, and please take a look.

docs/index.bs
```diff
@@ -2982,35 +2994,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-          1. If |e|'s <a>canceled flag</a> is set, set |eventCanceled| to true.
-          1. If |fetchInstance| is [=fetch/terminated=], then [=queue a task=] to [=AbortSignal/signal abort=] on |requestObject|'s {{Request/signal}}.
-
-          If |task| is discarded, set |handleFetchFailed| to true.
-
-          The |task| *must* use |activeWorker|'s <a>event loop</a> and the <a>handle fetch task source</a>.
+      1. Else:
+          1. Let |workerRealm| be the [=relevant realm=] of the |activeWorker|'s [=service worker/global object=].
+          1. Set |eventHandled| to [=a new promise=] in |workerRealm|.
+          1. [=queue a task=] |task| to run the following substeps:
```
jakearchibald:
```suggestion
          1. [=Queue a task=] |task| to run the following substeps:
```

docs/index.bs
```diff
@@ -2982,35 +2994,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+             The |task| *must* use |activeWorker|'s <a>event loop</a> and the <a>handle fetch task source</a>.
 
       1. Wait for |task| to have executed or for |handleFetchFailed| to be true.
       1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
-      1. If |respondWithEntered| is false, then return a [=network error=] if |eventCanceled| is true and null otherwise.
-      1. If |handleFetchFailed| is true, then return a [=network error=].
-      1. Return |response|.
+      1. If |respondWithEntered| is false, then:
+          1. If |eventCanceled| is true, then [=reject=] |eventHandled| with a "{{NetworkError}}" {{DOMException}} in |workerRealm| if |eventHandled| is not null and return a [=network error=].
```
jakearchibald:
```suggestion
          1. If |eventCanceled| is true, then:
              1. If |eventHandled| is not null, then [=reject=] |eventHandled| with a "{{NetworkError}}" {{DOMException}} in |workerRealm|.
              1. Return a [=network error=].
```

docs/index.bs
```diff
@@ -2982,35 +2994,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+          1. [=Resolve=] |eventHandled| if |eventHandled| is not null and return null.
```
jakearchibald:
```suggestion
          1. If |eventHandled| is not null, then [=resolve=] |eventHandled|.
          1. Return null.
```

docs/index.bs
```diff
@@ -2982,35 +2994,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      1. If |handleFetchFailed| is true, then [=reject=] |eventHandled| with a "{{NetworkError}}" {{DOMException}} in |workerRealm| if |eventHandled| is not null and return a [=network error=].
```
jakearchibald:
```suggestion
      1. If |handleFetchFailed| is true, then:
          1. If |eventHandled| is not null, then [=reject=] |eventHandled| with a "{{NetworkError}}" {{DOMException}} in |workerRealm|.
          1. Return a [=network error=].
```

docs/index.bs
```diff
@@ -2982,35 +2994,41 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      1. [=Resolve=] |eventHandled| if |eventHandled| is not null and return |response|.
```
jakearchibald:
```suggestion
      1. If |eventHandled| is not null, then [=resolve=] |eventHandled|.
      1. Return |response|.
```

docs/index.bs
tingshao: Done

docs/index.bs
tingshao: Done

docs/index.bs
tingshao: Done

docs/index.bs
tingshao: Done

docs/index.bs
tingshao: Done

tingshao:
> I've expanded some of the lines out as it was a little tricky to work out which "if" was applying to what.
> 
> One remaining issue: |workerRealm| is out of scope when it's used.

Sorry that, I didn't notice the scope issue of |workerRealm|.  Then I added a top level declaration for it.  And thanks for breaking the long sentences. All is updated in the new change.

tingshao:
@jakearchibald @youennf @asutherland
The tests are ready now, any more comments on this PR and the tests are welcome. thanks.

jakearchibald:
@tingshao Thank you! Can you create tickets for Chrome, Firefox, and Safari to implement this?

tingshao:
> @tingshao Thank you! Can you create tickets for Chrome, Firefox, and Safari to implement this?

Sure, thanks.

Issue #682, opened 2015-04-17
Clients.claim() does not deal with a client already controlled by another registration


nikhilm:
Consider a page located at /a/b/c/d.html

1. Register SW for /a
2. Load d.html -> d is controlled by /a
3. Register another SW for /a/b. Now d is still controlled by /a but the 'matching registration for scope' will be '/a/b'
4. Register another SW for '/a/b/c'. This SW calls claim() in onactivate.
    At this point, we should first run Handle Service Worker Client Unload steps with the client, then change the client's active worker. The former is not being done right now.

jungkees:
`.claim()` claims all the clients whose origin is the same origin to the service worker the method is running  on to be immediately controlled by it as long as the service worker is an active worker.

In the example, the registration for '/a', for '/a/b' and for '/a/b/c' are three different registrations. And at step 4, the service worker in the registration for '/a/b/c', as it's been an active worker already, immediately becomes the controller for d.html which was being controlled by the registration for '/a' by this time.

I don't see a reason Handle Service Worker Client Unload should be run here. Please let me know if I'm misunderstanding your question.

jakearchibald:
In @nikhilm's example, imagine the registration with scope `/a` has an active worker and a waiting worker. The waiting worker cannot become active until all its clients are gone. It only has one client, `/a/b/c/d.html`.

* New registration with scope `/a/b/c` is created
* SW for `/a/b/c` activates
* SW for `/a/b/c` calls `clients.claim()`
* `/a/b/c/d.html` is now controlled by the registration at scope `/a/b/c`

This means the registration with scope `/a` now has no clients, so its waiting worker should become active. As @nikhilm says, calling Handle Service Worker Client Unload before [3.1.3.1](https://slightlyoff.github.io/ServiceWorker/spec/service_worker/#clients-claim) should make this possible.

nikhilm:
:+1: @jakearchibald

jungkees:
Gotcha! Thanks!

PR #1506, opened 2020-02-26
Define Purge Service Worker Registrations

jungkees:
This change adds an algorithm, Purge Service Worker Registrations, that
immediately unregister - by calling into Clear Registration - service
worker registrations for a given origin. It takes an origin and a
boolean argument for unclaiming controlled clients. This is designed for
Clear-Site-Data's "storage" directive. job's immediate unregister flag
added in this change can be used for unregister({immediate: true}) or
alternative designs as a TODO.

Issue: https://github.com/w3c/ServiceWorker/issues/614.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2351,6 +2351,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
     A <a>job</a> has a <dfn id="dfn-job-list-of-equivalent-jobs">list of equivalent jobs</dfn> (a list of <a>jobs</a>). It is initially the empty list.
 
     A [=job=] has a <dfn id="dfn-job-force-bypass-cache-flag">force bypass cache flag</dfn>. It is initially unset.
+
+    A [=job=] has a <dfn id="dfn-immediate-unregister-flag">immediate unregister flag</dfn>. It is initially unset.
   </div>
 
 
@@ -3161,7 +3163,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       : Output
       :: none
 
-      1. If the [=environment settings object/origin=] of |job|'s [=job/scope url=] is not |job|'s [=job/client=]'s [=environment settings object/origin=], then:
+      1. If |job|'s [=job/client=] is not null, and the [=environment settings object/origin=] of |job|'s [=job/scope url=] is not |job|'s [=job/client=]'s [=environment settings object/origin=], then:
           1. Invoke [=Reject Job Promise=] with |job| and "{{SecurityError}}" {{DOMException}}.
           1. Invoke <a>Finish Job</a> with |job| and abort these steps.
       1. Let |registration| be the result of running <a>Get Registration</a> algorithm passing |job|'s [=job/scope url=] as the argument.
@@ -3170,7 +3172,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. Invoke <a>Finish Job</a> with |job| and abort these steps.
       1. [=map/Remove=] [=scope to registration map=][|job|'s [=job/scope url=]].
       1. Invoke <a>Resolve Job Promise</a> with |job| and true.
-      1. Invoke [=Try Clear Registration=] with |registration|.
+      1. If |job|'s [=immediate unregister flag=] is set, invoke [=Clear Registration=] with |registration|.
+      1. Else, invoke [=Try Clear Registration=] with |registration|.
 
           Note: If [=Try Clear Registration=] does not trigger [=Clear Registration=] here, [=Clear Registration=] is tried again when the last client [=using=] the registration is [=Handle Service Worker Client Unload|unloaded=] or the [=extend lifetime promises=] for the registration's service workers settle.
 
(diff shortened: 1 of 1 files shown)


jungkees:
I gave it a try for unblocking the kill-switch work for Clear-Site-Data: "storage". Please take a look.

/cc @asakusuma @n8schloss @asutherland @cdumez @youennf @slightlyoff

docs/index.bs
```diff
@@ -3230,6 +3233,36 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+              1. [=While=] |jobQueue| is not empty:
+                  1. Let |job| be the first item in |jobQueue|.
+                  1. Invoke [=Reject Job Promise=] with |job| and "{{AbortError}}" {{DOMException}}.
+                  1. [=queue/Dequeue=] from |jobQueue|.
+              1. Invoke [=Schedule Job=] with |job|.
+              1. Wait until |job|'s [=job promise=] settles.
+              1. If |unclaim| is true, then:
+                  1. For each [=/service worker client=] |client| [=using=] |registration|:
+                      1. Assert: |client|'s [=active service worker=] is not null.
+                      1. Invoke [=Handle Service Worker Client Unload=] with |client|.
```
asakusuma: What's the purpose of calling `Handle Service Worker Client Unload` here? I believe `Handle Service Worker Client Unload` will call `Try Unregister` and `Try Activate`, not sure either of those are necessary since we already did `unregister` and we shouldn't be trying to activate a new service worker.

docs/index.bs
```diff
@@ -3230,6 +3233,36 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      :: None
+
+      1. [=map/For each=] <var ignore=''>scope</var> → |registration| of [=scope to registration map=]:
+          1. Let |scopeURL| be |registration|'s [=service worker registration/scope url=].
+          1. If |scopeURL|'s [=/origin=] is |origin|, then:
+              1. Let |job| be the result of running [=Create Job=] with *unregister*, |scopeURL|, null, null, and null.
+              1. Set |job|'s [=immediate unregister flag=].
+              1. Let |jobQueue| be [=scope to job queue map=][|job|'s [=job/scope url=], [=URL serializer|serialized=]].
+              1. Assert: |jobQueue| is not null.
+              1. [=While=] |jobQueue| is not empty:
```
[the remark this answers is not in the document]
jungkees: I made the caller (Clear-Site-Data) purge the registrations with [the _unclaim_ enabled]((https://github.com/w3c/webappsec-clear-site-data/pull/63/files#diff-117d6498d2aa8019cc0abf5eeb87a9faR631)). I wanted to consider the possibility of any future change where a caller may want to purge the storage without unclaiming.

However, if our decision is making Clear-Site-Data: "strorage" unclaim by default, I can move the unclaiming steps into Unregister algorithm so it runs when job's _immediate unregister flag_ is set.

docs/index.bs
```diff
@@ -3230,6 +3233,36 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+                  1. Let |job| be the first item in |jobQueue|.
+                  1. Invoke [=Reject Job Promise=] with |job| and "{{AbortError}}" {{DOMException}}.
```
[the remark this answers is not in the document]
jungkees: I think ["AbortError" DOMException](https://heycam.github.io/webidl/#aborterror) is most relevant for the error context here. When the service operation sends out the Clear-Site-Data: "storage", it can expect the pending register/update/unregister jobs to be aborted and delete from the queue. Would it make sense?

docs/index.bs
jungkees: I think you're right. My concern was about taking care of the registration with different scope than the one that is claimed for in the case of claim(): https://github.com/w3c/ServiceWorker/issues/682. But as we basically purge the registration here entirely, we don't need to update the registration state. Can @asakusuma or anyone confirm that is right?

docs/index.bs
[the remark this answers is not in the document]
annevk: I think we should use "AbortError" for https://dom.spec.whatwg.org/#aborting-ongoing-activities generally. I'd use the generic error here type. (The message that goes with it is currently not standardized and up to implementations.)

docs/index.bs
```diff
@@ -3230,6 +3233,36 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+      : Input
+      :: |origin|, an [=/origin=]
+      :: |unclaim|, an optional boolean, false by default
+      : Output
+      :: None
+
+      1. [=map/For each=] <var ignore=''>scope</var> → |registration| of [=scope to registration map=]:
+          1. Let |scopeURL| be |registration|'s [=service worker registration/scope url=].
+          1. If |scopeURL|'s [=/origin=] is |origin|, then:
```
jakearchibald: Nit: I tend to invert early-exit conditions like this.

1. If |scopeURL|'s [=/origin=] is not |origin|, then [=continue=].

Then you don't need to nest the remaining steps.

docs/index.bs
```diff
@@ -3230,6 +3233,36 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+              1. Set |job|'s [=immediate unregister flag=].
+              1. Let |jobQueue| be [=scope to job queue map=][|job|'s [=job/scope url=], [=URL serializer|serialized=]].
+              1. Assert: |jobQueue| is not null.
+              1. [=While=] |jobQueue| is not empty:
+                  1. Let |job| be the first item in |jobQueue|.
+                  1. Invoke [=Reject Job Promise=] with |job| and "{{AbortError}}" {{DOMException}}.
+                  1. [=queue/Dequeue=] from |jobQueue|.
+              1. Invoke [=Schedule Job=] with |job|.
+              1. Wait until |job|'s [=job promise=] settles.
+              1. If |unclaim| is true, then:
```
jakearchibald: If unclaim is false, wouldn't that leave pages being controlled by a service worker in a "redundant" state? That doesn't seem right. If we're ripping the service worker out, surely we have to unclaim too?

docs/index.bs
jakearchibald: I agree we shouldn't be calling "Handle Service Worker Client Unload" here (the client isn't unloading). The registration state is already purged in the unregister job.

docs/index.bs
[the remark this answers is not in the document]
jungkees: Changed it to TypeError.

docs/index.bs
jungkees: Removed the step calling into Handle Service Worker Client Unload.

docs/index.bs
jungkees: Addressed. That's nicer indeed.

docs/index.bs
jungkees: I thought deleting registration and maintaining controller as two separate primitives, but agree to your point that it wouldn't be useful to keep the controller which is in "redundant" state. I moved unclaim steps into Unregister that unconditionally runs when the immediate unregister flag is set.

jungkees:
>  it feels like we need to do something about the currently running job.

Agreed. I'll follow up on this as a TODO.

annevk:
One thing that has happened since February is that we're getting much closer to being able to integrate the underlying semantics of Clear-Site-Data into the storage standard. This is being discussed in https://github.com/whatwg/storage/issues/88 and issues referenced from there.

https://storage.spec.whatwg.org/ also defines a set of primitives that service workers need to use to store their data. We're currently working on doing the same for localStorage/sessionStorage/IDB.

The work you are doing here really ties into all that. It might be worth tackling as a follow-up, but I thought I'd raise it now to create some more awareness and alert you to the possibility that the particular way these things integrate might change a bit.

jungkees:
@annevk, thanks for the heads up. Sounds great. Looking forward to plugging into the primitives Storage comes up with.

wanderview:
What is the status of this PR?  Is it close to merging?  I think Microsoft already implemented and shipped the clear-site-data purging in chromium.

jungkees:
Things we should address to merge this PR:
- Covering currently running jobs more accurately: https://github.com/w3c/ServiceWorker/pull/1506#issuecomment-599823612
- Check out if we should use any primitive updated in the Storage standard: https://github.com/w3c/ServiceWorker/pull/1506#issuecomment-643420767

I'm afraid I can't find cycles now. If anyone can help with this, I'd appreciate that.

asakusuma:
@jungkees @jakearchibald would it be possible to merge this PR or add a placeholder so that we can start cross linking from other specs and proposals, like storage/buckets?

jakearchibald:
I think we should remove some of the not-quite-correct normative steps, but happy to land a placeholder algorithm.

PR #1516, opened 2020-06-02
Introduce Cross-Origin Embedder Policy

### LINKED ISSUE #1490: cache.match() and COEP
What should we do if `cache.match()` is called in a context with `Cross-Origin-Embedder-Policy: require-corp` and the Response to be returned does not have a `Cross-Origin-Resource-Policy` header?

I would like to advocate that we reject the `match()` since it seems possible there could be information stored in the headers that should not be exposed to spectre attacks.  Also, it seems you can have a CORS response without a CORP header that would fail the COEP check and we would not want to expose the body in that case.

@mikewest @annevk @yutakahirano @makotoshimazu

yutakahirano:
This is part of https://github.com/whatwg/html/pull/5454.

 - Define embedder policy in environment settings object for service
   workers.
 - Add the CORP check in #dom-cache-matchall.

Closes #1490 and whatwg/fetch#985.

***
<a href="https://pr-preview.s3.amazonaws.com/yutakahirano/ServiceWorker/pull/1516.html" title="Last updated on Jun 10, 2020, 1:16 PM UTC (65484fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1516/0aceaac...yutakahirano:65484fc.html" title="Last updated on Jun 10, 2020, 1:16 PM UTC (65484fc)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -164,6 +164,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A <a>script resource</a> has an associated <dfn export for="script resource" id="dfn-referrer-policy">referrer policy</dfn> (a [=/referrer policy=]). It is initially the empty string.
 
+    A [=/service worker=] has an associated <dfn>embedder policy</dfn> (an [=/embedder policy=]). It is initially set to a new [=/embedder policy=].
+
     A [=/service worker=] has an associated <dfn export id="dfn-script-resource-map">script resource map</dfn> which is an <a>ordered map</a> where the keys are [=/URLs=] and the values are [=/responses=].
 
     A [=/service worker=] has an associated  <dfn export id="dfn-set-of-used-scripts">set of used scripts</dfn> (a [=ordered set|set=]) whose [=list/item=] is a [=/URL=]. It is initially a new [=ordered set|set=].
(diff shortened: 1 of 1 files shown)


yutakahirano:
@annevk @domenic PTAL.

Some links don't work because whatwg/html#5454 has not been landed yet.

docs/index.bs
```diff
@@ -1861,7 +1863,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                     1. Add a copy of |requestResponse|'s response to |responses|.
             1. Else:
                 1. Let |requestResponses| be the result of running [=Query Cache=] with |r| and |options|.
+                1. Let |request for CORP check| be a copy of |r|.
+                1. Set |request for CORP check|'s [=request/mode=] to "<code>no-cors</code>".
+                1. Set |request for CORP check|'s [=request/origin=] to |promise|'s [=relevant settings object=]'s [=environment settings object/origin=].
+                1. Set |request for CORP check|'s [=request/client=] to |promise|'s [=relevant settings object=].
                 1. [=list/For each=] |requestResponse| of |requestResponses|:
+                    1. If |requestResponse|'s [=response/type=] is "<code>opaque</code>" and [=cross-origin resource policy check=] with |request for CORP check| and |requestResponse| returns <b>blocked</b>, then reject |promise| with a `TypeError` and abort these steps.
```
annevk, line 1871: Can we provide a better abstraction for this caller whereby you give an origin and a response and get an answer?

It's also not clear to me that the current algorithm does the right thing here as it's not using the URL from the response to compare with the request origin, but rather the URL of the request.

yutakahirano: Done, I believe.

docs/index.bs
```diff
@@ -3467,6 +3476,24 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      :: |settings object|, an [=/environment settings object=]
+      : Output
+      :: a boolean
+
+      1. If |response|'s [=response/type=] is not "<code>opaque</code>", then return true.
+      1. Let |request| be a copy of |passed request|.
+      1. Set |request|'s [=request/mode=] to "<code>no-cors</code>".
+      1. Set |request|'s [=request/origin=] to |settings object|'s [=environment settings object/origin=].
+      1. Set |request|'s [=request/client=] to |settings object|.
+      1. Return true if [=cross-origin resource policy check=] with |request| and |response| returns <b>allowed</b>, and false otherwise.
```
annevk, line 3494: As far as I can tell you moved this algorithm, but didn't change it. Why do we need to change request's origin and such? Why does request's mode need to be overridden?

yutakahirano: Done, I believe.

yutakahirano:
Can we land this as well?

jakearchibald:
Do we have tests for the cache API change?

yutakahirano:
Thank you!

> Do we have tests for the cache API change?

Yes, as [html/cross-origin-embedder-policy](https://github.com/web-platform-tests/wpt/tree/master/html/cross-origin-embedder-policy)/\*cache-storage\*.https.html.

liuchengwei555 (commented):
Apple

PR #1533, opened 2020-08-27
Use new Streams algorithms

domenic:
Follows https://github.com/whatwg/streams/pull/1073.

Do not merge until that PR is merged. But, feel free to review.

See also https://github.com/tabatkins/bikeshed/issues/1758.

***
<a href="https://pr-preview.s3.amazonaws.com/domenic/ServiceWorker/pull/1533.html" title="Last updated on Sep 15, 2020, 2:16 PM UTC (a483802)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1533/e4c2953...domenic:a483802.html" title="Last updated on Sep 15, 2020, 2:16 PM UTC (a483802)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1589,7 +1589,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                 1. Let |potentialResponse| be a copy of |response|'s associated [=Response/response=], except for its [=response/body=].
                 1. If |response|'s [=response/body=] is non-null, run these substeps:
                     1. Let |reader| be the result of [=ReadableStream/getting a reader=] from |response|'s [=response/body=]'s [=stream=].
-                    1. Let |pullAlgorithm| be an action that runs these subsubsteps:
+                    1. Let |pullAlgorithm| be an action that runs these steps:
                         1. Let |readRequest| be a new [=read request=] with the following [=struct/items=]:
                             : [=read request/chunk steps=], given |chunk|
                             ::
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1600,6 +1600,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                             ::
                                 1. Set |end-of-body| to true.
                             : [=read request/error steps=]
+                            ::
                                 1. [=ReadableStream/error=] |newStream| with a {{TypeError}}.
                         1. [=ReadableStreamDefaultReader/Read a chunk=] from |reader| given |readRequest|.
                     1. Let |cancelAlgorithm| be an action that [=ReadableStreamDefaultReader/cancels=] |reader|.
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -34,8 +34,6 @@ spec: fetch;
     type: dfn
         for: /; text: fetch
         text: empty
-    type: interface
-        text: ReadableStream
 spec: infra;
     type: dfn;
         text: list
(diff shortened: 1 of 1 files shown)


ricea:
Drive-by lgtm.

docs/index.bs
```diff
@@ -1590,25 +1588,33 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                 1. Let |done| be false.
                 1. Let |potentialResponse| be a copy of |response|'s associated [=Response/response=], except for its [=response/body=].
                 1. If |response|'s [=response/body=] is non-null, run these substeps:
-                    1. Let |reader| be the result of [=get a reader|getting a reader=] from |response|'s [=response/body=]'s [=stream=].
+                    1. Let |reader| be the result of [=ReadableStream/getting a reader=] from |response|'s [=response/body=]'s [=stream=].
+                    1. Let |pullAlgorithm| be an action that runs these subsubsteps:
```
jakearchibald, line 1592:
```suggestion
                    1. Let |pullAlgorithm| be an action that runs these steps:
```

I'll remove all other uses of substeps in another PR.

docs/index.bs
```diff
@@ -1590,25 +1588,33 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+                            : [=read request/chunk steps=], given |chunk|
+                            ::
+                                1. Assert: |chunk| is a {{Uint8Array}}.
+                                1. Append the bytes represented by |chunk| to |bytes|.
+                                1. Perform ! [=DetachArrayBuffer=](|chunk|.\[[ViewedArrayBuffer]]).
+                            : [=read request/close steps=]
+                            ::
+                                1. Set |end-of-body| to true.
+                            : [=read request/error steps=]
+                                1. [=ReadableStream/error=] |newStream| with a {{TypeError}}.
```
jakearchibald, line 1603: The markup seems inconsistent here. With the error steps, the steps aren't in a definition item. Given that they're both a single step, do they need to be in an ordered list at all? Should it be:

```suggestion
                            : [=read request/close steps=]
                            :: Set |end-of-body| to true.
                            : [=read request/error steps=]
                            :: [=ReadableStream/error=] |newStream| with a {{TypeError}}.
```

Also, is it typical to use lower case for "close steps" and "error" even though they're at the beginning of a sentence?

docs/index.bs
```diff
@@ -1590,25 +1588,33 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-                        1. When |promise| is fulfilled with an object whose `done` property is true, set |end-of-body| to true.
-                        1. When |promise| is fulfilled with a value that matches with neither of the above patterns, or |promise| is rejected, [=ReadableStream/error=] |newStream| with a `TypeError`.
-                    1. Let |cancel| be an action that [=ReadableStream/cancels=] |response|'s [=response/body=]'s [=stream=] with |reader|.
-                    1. Let |newStream| be the result of [=ReadableStream/construct a ReadableStream object=] with |highWaterMark|, |sizeAlgorithm|, |pull|, and |cancel| in |targetRealm|.
+                    1. Let |newStream| be the result of [=ReadableStream/creating=] a {{ReadableStream}} with <a for=ReadableStream/create><var ignore>pullAlgorithm</var></a> set to |pullAlgorithm|, <a for=ReadableStream/create><var ignore>cancelAlgorithm</var></a> set to |cancelAlgorithm|, <a for=ReadableStream/create><var ignore>highWaterMark</var></a> set to |highWaterMark|, and <a for=ReadableStream/create><var ignore>sizeAlgorithm</var></a> set to |sizeAlgorithm|, in |targetRealm|.
```
jakearchibald, line 1608: Huh, was this the agreed markup for this? It seems weird to create a var that you have to ignore. And of course it doesn't work with bikeshed's shorthand.

jakearchibald (changes requested):

docs/index.bs
domenic: Yeah, see https://github.com/tabatkins/bikeshed/issues/1758

docs/index.bs
domenic: Thanks for catching the missing `::`!

I find it clearer when steps are numbered, but we could remove it if you'd prefer; the reader can understand.

I don't think of `<dd>`s as "sentences", personally, but instead terms. So I don't capitalize them unless they're proper nouns.

jakearchibald: Fair enough. Add in the `::` then it's good to go.

docs/index.bs
jakearchibald: I'm ok with this, but in general I don't see the need for the vars here.

When we "Set document's foo to bar", we don't wrap foo in a `<var>`, so I don't see the need to do it here.

domenic: See the discussion in https://github.com/whatwg/infra/issues/320

docs/index.bs
domenic: Forgot to push; should be good now!

jakearchibald (approved):

PR #1545, opened 2020-10-05
Define settings object's cross-origin isolated capability for SW

yutakahirano:
Related to https://github.com/whatwg/html/issues/5435 and
https://github.com/whatwg/html/pull/5752.

***
<a href="https://pr-preview.s3.amazonaws.com/yutakahirano/ServiceWorker/pull/1545.html" title="Last updated on Feb 4, 2021, 7:57 PM UTC (fdbaf4f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1545/044cc99...yutakahirano:fdbaf4f.html" title="Last updated on Feb 4, 2021, 7:57 PM UTC (fdbaf4f)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2895,7 +2895,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
               : The [=environment settings object/embedder policy=]
               :: Return |workerGlobalScope|'s [=WorkerGlobalScope/embedder policy=].
               : The [=environment settings object/cross-origin isolated capability=]
-              :: Return whether |workerGlobalScope|'s [=WorkerGlobalScope/embedder policy=]'s [=embedder policy/value] is "`[embedder policy value/require-corp]`".
+              :: Return whether |workerGlobalScope|'s [=WorkerGlobalScope/embedder policy=]'s [=embedder policy/value=] is "`[=embedder policy value/require-corp=]`".
 
           1. Set |settingsObject|'s [=environment/id=] to a new unique opaque string, [=creation URL=] to |serviceWorker|'s [=service worker/script url=], [=environment/top-level creation URL=] to null, [=environment/top-level origin=] to an [=implementation-defined=] value, [=environment/target browsing context=] to null, and [=active service worker=] to null.
           1. Set |workerGlobalScope|'s [=WorkerGlobalScope/url=] to |serviceWorker|'s [=service worker/script url=].
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2876,7 +2876,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |startFailed| be false.
       1. Let |agent| be the result of [=obtain a service worker agent|obtaining a service worker agent=], and run the following steps in that context:
           1. Let |agentCluster| be the [=agent cluster=] which contains |agent|.
-          1. If |workerGlobalScope|'s [=WorkerGlobalScope/embedder policy=] is "<code>[=embedder policy value/require-corp=]</code>", then set |agentCluster|'s [=agent cluster/cross-origin isolation mode=] to "<code>[=cross-origin isolation mode/logical=]</code>" or "<code>[=cross-origin isolation mode/concrete=]</code>". The one chosen is implementation-defined.
+          1. If |workerGlobalScope|'s [=WorkerGlobalScope/embedder policy=] is "<code>[=embedder policy value/require-corp=]</code>", then set |agentCluster|'s [=agent-cluster-cross-origin-isolation|cross-origin isolation mode=] to "<code>[=cross-origin isolation mode/logical=]</code>" or "<code>[=cross-origin isolation mode/concrete=]</code>". The one chosen is implementation-defined.
           1. Let |realmExecutionContext| be the result of [=create a new JavaScript realm|creating a new JavaScript realm=] given |agent| and the following customizations:
               * For the global object, create a new {{ServiceWorkerGlobalScope}} object. Let |workerGlobalScope| be the created object.
           1. Set |serviceWorker|'s [=service worker/global object=] to |workerGlobalScope|.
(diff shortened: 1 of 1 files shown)


yutakahirano:
This is ready for review.

cc: @domenic

domenic:
Interesting. /cc @annevk. I would have expected this to be done more similar to how shared workers are done.

In particular, is it OK to leave the service worker agent cluster's "cross-origin isolated" as false? That will impact its ability to postMessage SharedArrayBuffers to itself, for example, as well as whether the SharedArrayBuffer global exists. (Or maybe those are broken and should use the capability??)

annevk:
When I reread https://github.com/whatwg/html/issues/5435#issuecomment-609659086 and https://github.com/whatwg/html/issues/5435#issuecomment-640517912 I wonder if I misread the latter. In particular 5 there does not depend on both whereas I suggest above that it would. Given how we defined `[CrossOriginIsolated]` in IDL it would make sense to me to align `SharedArrayBuffer` with that. What do you think @yutakahirano?

Now then there is the question of shared/service workers. I think they should have a shared model. And I suppose the main question there is whether user agents can guarantee to run those in their own process, even on Android. That is, can a COEP shared/service worker always get a process separate from the document it is associated with (which could be a third-party document that does not have the capability)? If we can make that guarantee it seems okay for them to always have the capability. I'd rather not end up in a situation where they sometimes have the capability though, depending on system resources.

Does that help and make sense?

yutakahirano:
@domenic Oh sorry I missed that. I think it's good to add a boolean argument to https://html.spec.whatwg.org/multipage/webappapis.html#obtain-a-service-worker-agent, and give the appropriate value in https://w3c.github.io/ServiceWorker/#run-service-worker-algorithm. What do you think?

@annevk 

>  I suppose the main question there is whether user agents can guarantee to run those in their own process, even on Android. That is, can a COEP shared/service worker always get a process separate from the document it is associated with (which could be a third-party document that does not have the capability)?

I'm not sure if I understand. Chromium implementation is going to allocate a process for (origin or site, agent cluster's cross-origin isolated). We don't take the capability into account.

annevk:
@yutakahirano so the scenario is that A embeds B and B has a shared/service worker Bsw. All have the appropriate COOP+COEP headers. But A doesn't delegate the capability to B. Now in resource-constrained environments the model allows for A and B to be in the same process and I think the idea with the capability is (please correct me if I'm wrong) that B not having access to certain features means it's harder to attack A.

Now, if Bsw were to share that process, it could attack A.

yutakahirano:
I think I brought up the point at https://github.com/whatwg/html/pull/5752#issuecomment-662302867. At that time the conclusion was don't care.

annevk:
Okay, I do think we should care about it as otherwise the cross-origin isolated capability is security theater and we would be better of not having the additional complexity. Perhaps I'm missing some subtlety in the motivation behind it?

yutakahirano:
Is this something we want to talk about at SW WG in TPAC?
@mikewest @camillelamy are you interested in this topic?

domenic:
> I think it's good to add a boolean argument to https://html.spec.whatwg.org/multipage/webappapis.html#obtain-a-service-worker-agent, and give the appropriate value in https://w3c.github.io/ServiceWorker/#run-service-worker-algorithm. What do you think?

That seems right to me.

Regarding the capability issue, the attack in https://github.com/w3c/ServiceWorker/pull/1545#issuecomment-704188323 is an interesting point. I hadn't fully considered the process model implications here. /cc @arturjanc

yutakahirano:
After seeing whatwg/html#6060, I'm starting thinking it makes sense to introduce a platform dependency to cross-origin-isolated capability. It represents whether web developers can use SAB and co. On some platforms COOP+COEP is sufficient to ensure the security guarantees but on other platforms it's more difficult.

The cross-origin isolated capability on shared and service workers is false, when the user agent may put a cross-origin environment into the process.

annevk:
It's a bit mucky though as there will also be platforms where it doesn't matter if A delegates to B or not as they never share a process. But if it didn't delegate B would not and Bsw would have it. Maybe that's okay, but it doesn't seem satisfying nor very clear to web developers.

mfalken:
(belated) Notes from [Service Worker WG meeting on Nov 20](https://docs.google.com/document/d/1ybS1q2HCPh3bNNOkjGpAPFug19A2BsIxYEi-i6lrB1w/edit?usp=sharing):

* The majority of time was spent trying to understand the issue.
* A clear description of the problem/attack would help the WG understand the issue.

yutakahirano:
Introducing implementation-dependency has already been done in whatwg/html#6098 (Search for "The one chosen is implementation-defined" in https://html.spec.whatwg.org/multipage/workers.html#run-a-worker).

Updated the change.

yutakahirano:
Reopening as suggested at https://github.com/whatwg/html/pull/6325#issuecomment-901060045.

yutakahirano:
cc: @ArthurSonzogni @nhiroki 

I updated the PR. How does this look like?

docs/index.bs
```diff
@@ -2901,7 +2906,7 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
       1. Let |script| be |serviceWorker|'s [=service worker/script resource=].
       1. Assert: |script| is not null.
       1. Let |startFailed| be false.
-      1. Let |agent| be the result of [=obtain a service worker agent|obtaining a service worker agent=], and run the following steps in that context:
+      1. Let |agent| be the result of [=obtain a service worker agent|obtaining a service worker agent=] with |serviceWorker|'s [=service worker/embedder policy=], and run the following steps in that context:
```
yutakahirano, line 2909: (Let's assume the procedure sets agent's agent cluster's cross-origin isolation mode. I'll fix the HTML side.)

yutakahirano:
@annevk would you take a look again?

docs/index.bs
```diff
@@ -2679,11 +2682,13 @@ spec: rfc7231; urlPrefix: https://tools.ietf.org/html/rfc7231
               1. Asynchronously complete these steps with a <a>network error</a>.
           1. Let |url| be |request|'s [=request/url=].
           1. Set |updatedResourceMap|[|url|] to |response|.
+          1. Set |embedderPolicy| to the result of [=obtaining an embedder policy=] from |response| and |request|'s |client|.
           1. If |response|'s [=response/cache state=] is not "`local`", set |registration|'s [=last update check time=] to the current time.
           1. Set |hasUpdatedResources| to true if any of the following are true:
             * |newestWorker| is null.
             * |newestWorker|'s [=service worker/script url=] is not |url| or |newestWorker|'s [=service worker/type=] is not |job|'s [=worker type=].
             * |newestWorker|'s [=script resource map=][|url|]'s [=response/body=] is not byte-for-byte identical with |response|'s [=response/body=].
+            * |newestWorker|'s [=service worker/embedder policy=] does not equal to |embedderPolicy|.
```
annevk, line 2691: Since an embedder policy is essentially a struct these days, I think we need to define the comparison operation. Or do all values need to be equal here, including the reporting values?

annevk (commented):
Thanks for working on this @yutakahirano! I think this looks reasonable, modulo the comparison operation. It would be nice if we could share more logic with workers, but I suppose that requires a bigger rewrite on the HTML side.

docs/index.bs
yutakahirano: I don't have a strong opinion. I think it's reasonable to update the script if coep's value changes, but I'm not so sure about other properties (reporting value, endpoint, reporting endpoint).

@ArthurSonzogni @nhiroki @jakearchibald do you have any opinions?

ArthurSonzogni: (Not a SW specialist)
I guess this govern whether or a not the new script should be used to replace the old SW at some point. As a developer, I think this is desirable to make it happen when any of those 4 COEP attributes are changed.
Do you foresee any strong drawback to this?

PR #1563, opened 2021-02-08
Have "Handle Fetch" return error when request's streaming body is unusable

yutakahirano:
***
<a href="https://pr-preview.s3.amazonaws.com/yutakahirano/ServiceWorker/pull/1563.html" title="Last updated on Feb 18, 2021, 4:09 AM UTC (40e9e10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1563/044cc99...yutakahirano:40e9e10.html" title="Last updated on Feb 18, 2021, 4:09 AM UTC (40e9e10)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3039,6 +3039,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                   1. Wait until |e|'s [=FetchEvent/wait to respond flag=] is unset.
                   1. If |e|'s [=FetchEvent/respond-with error flag=] is set, set |handleFetchFailed| to true.
                   1. Else, set |response| to |e|'s [=FetchEvent/potential response=].
+              1. If |response| is null and |requestBody| is not null, then:
+                  1. If |requestBody|'s [=Body/source=] is null, then:
+                      1. If |requestBody|'s [=Body/stream=] is [=Body/disturbed=] or [=Body/locked=], then return a [=network error=] and null.
+                      1. Set |requestBody| to the result of [=ReadableStream/create a proxy|creating a proxy=] for |requestBody|'s [=Body/stream=].
+                  1. Else:
+                      1. Set |requestBody| to the first component of the result of [=bodyinit/safely-extract|safely extracting=] |requestBody|'s [=Body/source=].
               1. If |e|'s <a>canceled flag</a> is set, set |eventCanceled| to true.
               1. If |fetchInstance| is [=fetch/terminated=], then [=queue a task=] to [=AbortSignal/signal abort=] on |requestObject|'s {{Request/signal}}.
 
@@ -3048,12 +3054,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       1. Wait for |task| to have executed or for |handleFetchFailed| to be true.
       1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
-      1. If |response| is null and |requestBody| is not null, then:
-          1. If |requestBody|'s [=Body/source=] is null, then:
-              1. If |requestBody|'s [=Body/stream=] is [=Body/disturbed=] or [=Body/locked=], then return a [=network error=] and null.
-              1. Set |requestBody| to the result of [=ReadableStream/create a proxy|creating a proxy=] for |requestBody|'s [=Body/stream=].
-          1. Else:
-              1. Set |requestBody| to the first component of the result of [=bodyinit/safely-extract|safely extracting=] |requestBody|'s [=Body/source=].
       1. If |respondWithEntered| is false, then:
           1. If |eventCanceled| is true, then:
               1. If |eventHandled| is not null, then [=reject=] |eventHandled| with a "{{NetworkError}}" {{DOMException}} in |workerRealm|.
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2949,8 +2949,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       : Input
       :: |request|, a [=/request=]
       : Output
-      :: |response|, null or a [=/response=]
-      :: |requestBody|, null or a [=/body=]
+      :: |response|, a [=/response=]
 
       1. Let |handleFetchFailed| be false.
       1. Let |respondWithEntered| be false.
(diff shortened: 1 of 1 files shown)


yutakahirano:
@annevk @yoichio

This is related to whatwg/fetch#1144.

docs/index.bs
```diff
@@ -2950,6 +2950,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       :: |request|, a [=/request=]
       : Output
       :: |response|, a [=/response=]
```
annevk, line 2952:
```suggestion
      :: |response|, null or a [=/response=]
```

docs/index.bs
```diff
@@ -2950,6 +2950,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      :: |requestBody|, a [=/body=]
```
annevk, line 2953:
```suggestion
      :: |requestBody|, null or a [=/body=]
```

docs/index.bs
```diff
@@ -3001,21 +3006,23 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+          1. Return null and |requestBody|.
       1. If |activeWorker|'s <a>state</a> is "`activating`", wait for |activeWorker|'s <a>state</a> to become "`activated`".
       1. If the result of running the [=Run Service Worker=] algorithm with |activeWorker| is *failure*, then set |handleFetchFailed| to true.
       1. Else:
           1. Set |workerRealm| to the [=relevant realm=] of the |activeWorker|'s [=service worker/global object=].
           1. Set |eventHandled| to [=a new promise=] in |workerRealm|.
           1. [=Queue a task=] |task| to run the following substeps:
               1. Let |e| be the result of <a>creating an event</a> with {{FetchEvent}}.
+              1. If |requestBody| is not null, then:
+                  1. Set |requestBody|'s [=Body/stream=] to the result of [=ReadableStream/create a proxy|creating a proxy=] for |requestBody|'s [=Body/stream=].
```
annevk, line 3025: Shouldn't this happen much earlier? We need to essentially transfer the request body back. If we just return it there are no side effects for the service worker side and there need to be when it's backed by a stream, right?

yutakahirano: Which point do you think is the best?

Do you mean [transferring stream](https://github.com/whatwg/streams/issues/276) by "transfer"? I think what's needed is a bit different from that transferring as 

 - We don't keep chunk boundary.
 - Chunks other than Uint8Arrays are rejected.

I think we need to define a means to create such a Body and/or ReadableStream in the fetch spec, but I don't want it to block streaming upload - it is an existing problem for Response.body, too.

annevk: It became a problem when we added streams. I think now would be a good time to solve it.

And what I mean is that if body is a stream with a null source, service workers needs to transfer ownership back to fetch, such that invoking `req.text()` will start failing afterwards. Presumably this applies to all cases where we return null.

yutakahirano: Ah, thank you. What do you think about the latest one then?

docs/index.bs
```diff
@@ -2962,26 +2963,30 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |workerRealm| be null.
       1. Let |eventHandled| be null.
       1. Let |fetchInstance| be the instance of the [=/fetch=] algorithm representing the ongoing fetch.
+      1. Let |requestBody| be |request|'s [=request/body=].
       1. Assert: |request|'s [=request/destination=] is not "<code>serviceworker</code>".
       1. If |request|'s [=request/destination=] is either "<code>embed</code>" or "<code>object</code>", then:
-          1. Return null.
+          1. Return null and |requestBody|.
```
annevk, line 2969: How is this case not impacted? Is that because there's not even a fetch event for these? Maybe we should return null and null there for clarity to indicate nothing could have changed?

docs/index.bs
```diff
@@ -2962,26 +2963,30 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-          1. If |registration| is null or |registration|'s <a>active worker</a> is null, return null.
+          1. If |registration| is null or |registration|'s <a>active worker</a> is null, return null and |requestBody|.
           1. If |request|'s [=request/destination=] is not {{RequestDestination/"report"}}, set |reservedClient|'s <a>active service worker</a> to |registration|'s <a>active worker</a>.
           1. If |request| is a [=navigation request=], |registration|'s [=navigation preload enabled flag=] is set, |request|'s [=request/method=] is \`<code>GET</code>\`, and |registration|'s [=active worker=]'s <a>set of event types to handle</a> [=set/contains=] <code>fetch</code>, then:
 
               Note: If the above is true except |registration|'s [=active worker=]'s <a>set of event types to handle</a> **does not** contain <code>fetch</code>, then the user agent may wish to show a console warning, as the developer's intent isn't clear.
 
-              1. Let |preloadRequest| be the result of [=request/cloning=] the request |request|.
+              1. Let |preloadRequest| be a copy of |request|, except for its [=request/body=].
```
annevk, line 2986: I don't see why we cannot clone here. Body will still have its source afterwards.

docs/index.bs
```diff
@@ -3032,6 +3039,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                   1. Wait until |e|'s [=FetchEvent/wait to respond flag=] is unset.
                   1. If |e|'s [=FetchEvent/respond-with error flag=] is set, set |handleFetchFailed| to true.
                   1. Else, set |response| to |e|'s [=FetchEvent/potential response=].
+              1. If |response| is null and |requestBody| is not null, then:
+                  1. If |requestBody|'s [=Body/source=] is null, then:
+                      1. If |requestBody|'s [=Body/stream=] is [=Body/disturbed=] or [=Body/locked=], then return a [=network error=] and null.
```
annevk, line 3044: You want =Body/unusable= here.

docs/index.bs
```diff
@@ -3032,6 +3039,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                   1. Wait until |e|'s [=FetchEvent/wait to respond flag=] is unset.
                   1. If |e|'s [=FetchEvent/respond-with error flag=] is set, set |handleFetchFailed| to true.
                   1. Else, set |response| to |e|'s [=FetchEvent/potential response=].
+              1. If |response| is null and |requestBody| is not null, then:
+                  1. If |requestBody|'s [=Body/source=] is null, then:
```
annevk, line 3043: This is unfortunate, but indeed something we're stuck with. Not sure if that's worth pointing out somehow.

annevk (commented):
I think this mostly looks good and I think I also misunderstood some things initially. Left suggestions for improvements.

yutakahirano:
Fixed the PR per the discussion at whatwg/fetch#1144.

docs/index.bs
```diff
@@ -3032,6 +3032,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                   1. Wait until |e|'s [=FetchEvent/wait to respond flag=] is unset.
                   1. If |e|'s [=FetchEvent/respond-with error flag=] is set, set |handleFetchFailed| to true.
                   1. Else, set |response| to |e|'s [=FetchEvent/potential response=].
+              1. If |response| is null, |request|'s [=request/body=] is not null, |request|'s [=request/body=]'s [=Body/source=] is null, and |request|'s [=request/body=] is [=Body/unusable=], set |handleFetchFailed| to true.
```
annevk, line 3035: The second check here is redundant with checking whether it is unusable. In order for a body to be unusable it has to be non-null.

yutakahirano: Makes sense, but I'd keep this because the next change added a cancel step.

annevk (approved):
Probably also needs review from a service worker maintainer.

yutakahirano:
@jakearchibald , can you take a look? This is part of whatwg/fetch#1144.

docs/index.bs
```diff
@@ -3032,6 +3032,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
                   1. Wait until |e|'s [=FetchEvent/wait to respond flag=] is unset.
                   1. If |e|'s [=FetchEvent/respond-with error flag=] is set, set |handleFetchFailed| to true.
                   1. Else, set |response| to |e|'s [=FetchEvent/potential response=].
+              1. If |response| is null, |request|'s [=request/body=] is not null, and |request|'s [=request/body=]'s [=Body/source=] is null, then:
+	          1. If |request|'s [=request/body=] is [=Body/unusable=], set |handleFetchFailed| to true.
+		  2. Else, [=ReadableStream/cancel=] |request|'s [=request/body=] with undefined.
```
MattiasBuelens, line 3037: The steps are numbered incorrectly, see screenshot:
image

I think it's because the indentation is inconsistent (tabs mixed with spaces). Also, all numbered steps should be prefixed with `1.`, we shouldn't have a `2.` in there.
```suggestion
              1. If |response| is null, |request|'s [=request/body=] is not null, and |request|'s [=request/body=]'s [=Body/source=] is null, then:
                  1. If |request|'s [=request/body=] is [=Body/unusable=], set |handleFetchFailed| to true.
                  1. Else, [=ReadableStream/cancel=] |request|'s [=request/body=] with undefined.
```

MattiasBuelens (changes requested):

docs/index.bs
yutakahirano: Thank you!

jakearchibald (approved):
LGTM! Glad this turned out so simple.

jakearchibald:
@yutakahirano are you happy for this to be merged?

yutakahirano:
Yeah, thank you!

PR #1575, opened 2021-03-28
Maintain timing for worker start/ready, for navigation requests.

noamr:
The timing when worker is started or ready are web-accessible via
the navigation timing API (https://w3c.github.io/navigation-timing/).

`workerStart` is specified to be the time when the service worker is run.

In Chromium and Gecko, `worker ready time` (exposed as `fetchStart`)
is set to the right before the time when the Fetch event can be
dispatched for the newly created worker.

In preparation for https://github.com/w3c/navigation-timing/issues/136

***
<a href="https://pr-preview.s3.amazonaws.com/noamr/ServiceWorker/pull/1575.html" title="Last updated on May 13, 2021, 11:19 AM UTC (21510d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1575/e005153...noamr:21510d9.html" title="Last updated on May 13, 2021, 11:19 AM UTC (21510d9)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -197,7 +197,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         * Detects abnormal operation: such as infinite loops and tasks exceeding imposed time limits (if any) while handling the events.
     </section>
 
-   <section>
+    <section>
       <h4 id="service-worker-events">Events</h4>
 
       The Service Workers specification defines <dfn export id="dfn-service-worker-events">service worker events</dfn> (each of which is an [=event=]) that include (see the <a href="#execution-context-events">list</a>):
@@ -2985,7 +2985,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       : Input
       :: |request|, a [=/request=]
-      :: |crossOriginIsolatedCapability|, a boolean
+      :: |useHighResPerformanceTimers|, a boolean
       : Output
       :: |response|, a [=/response=]
 
@@ -3048,7 +3048,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. If the result of running the [=Should Skip Event=] algorithm with "fetch" and |activeWorker| is true, then:
           1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
           1. Return null.
-      1. Let |timingInfo|'s [=service worker timing info/start time=] be the [=coarsened shared current time=] given |crossOriginIsolatedCapability|.
+      1. Let |timingInfo|'s [=service worker timing info/start time=] be the [=coarsened shared current time=] given |useHighResPerformanceTimers|.
       1. If |activeWorker|'s <a>state</a> is "`activating`", wait for |activeWorker|'s <a>state</a> to become "`activated`".
       1. If the result of running the [=Run Service Worker=] algorithm with |activeWorker| is *failure*, then set |handleFetchFailed| to true.
       1. Else:
@@ -3065,7 +3065,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
               1. If |request| is a <a>non-subresource request</a> and |request|'s [=request/destination=] is not {{RequestDestination/"report"}}, initialize |e|'s {{FetchEvent/resultingClientId}} attribute to |reservedClient|'s [=environment/id=], and to the empty string otherwise.
               1. If |request| is a <a>navigation request</a>, initialize |e|'s {{FetchEvent/replacesClientId}} attribute to |request|'s [=request/replaces client id=], and to the empty string otherwise.
               1. Initialize |e|’s {{FetchEvent/handled}} to |eventHandled|.
-              1. Let |timingInfo|'s [=service worker timing info/fetch event dispatch time=] to the [=coarsened shared current time=] given |crossOriginIsolatedCapability|.
+              1. Let |timingInfo|'s [=service worker timing info/fetch event dispatch time=] to the [=coarsened shared current time=] given |useHighResPerformanceTimers|.
               1. <a>Dispatch</a> |e| at |activeWorker|'s [=service worker/global object=].
               1. Invoke [=Update Service Worker Extended Events Set=] with |activeWorker| and |e|.
               1. If |e|'s [=FetchEvent/respond-with entered flag=] is set, set |respondWithEntered| to true.
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3052,7 +3052,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
           1. Return null.
       1. If |activeWorker|'s <a>state</a> is "`activating`", wait for |activeWorker|'s <a>state</a> to become "`activated`".
-      1. Let |workerStartTime| be the  [=coarsened shared current time=] given <var>crossOriginIsolatedCapability</var>.
+      1. Let |workerStartTime| be the [=coarsened shared current time=] given |crossOriginIsolatedCapability|.
       1. If the result of running the [=Run Service Worker=] algorithm with |activeWorker| is *failure*, then set |handleFetchFailed| to true.
       1. Else:
           1. Set |workerRealm| to the [=relevant realm=] of the |activeWorker|'s [=service worker/global object=].
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3048,6 +3048,8 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. If the result of running the [=Should Skip Event=] algorithm with "fetch" and |activeWorker| is true, then:
           1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
           1. Return null.
+      1. If |useHighResPerformanceTimers| is true, then set |useHighResPerformanceTimers| to |activeWorker|'s  
+         [=service worker/global object=]'s [=WorkerGlobalScope/cross-origin isolated capability=].
       1. Let |timingInfo|'s [=service worker timing info/start time=] be the [=coarsened shared current time=] given |useHighResPerformanceTimers|.
       1. If |activeWorker|'s <a>state</a> is "`activating`", wait for |activeWorker|'s <a>state</a> to become "`activated`".
       1. If the result of running the [=Run Service Worker=] algorithm with |activeWorker| is *failure*, then set |handleFetchFailed| to true.
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3076,7 +3076,6 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
               1. If |response| is null, |request|'s [=request/body=] is not null, and |request|'s [=request/body=]'s [=Body/source=] is null, then:
                   1. If |request|'s [=request/body=] is [=Body/unusable=], set |handleFetchFailed| to true.
                   1. Else, [=ReadableStream/cancel=] |request|'s [=request/body=] with undefined.
-              1. If |response| is not null, then set |response|'s [=response/service worker timing=] to |timingInfo|.
               1. If |e|'s <a>canceled flag</a> is set, set |eventCanceled| to true.
               1. If |fetchInstance| is [=fetch/terminated=], then [=queue a task=] to [=AbortSignal/signal abort=] on |requestObject|'s {{Request/signal}}.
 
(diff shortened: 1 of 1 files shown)


noamr:
@jakearchibald @wanderview r?

jakearchibald:
Sorry for the delay, I've been on leave. I'll try to get to this this week, or early next week.

wanderview:
Sorry, I missed I was tagged here.  I generally defer to the editors to do spec reviews.

docs/index.bs
```diff
@@ -2971,6 +2975,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. Let |preloadResponse| be a new [=promise=].
       1. Let |workerRealm| be null.
       1. Let |eventHandled| be null.
+      1. Let |crossOriginIsolatedCapability| be |request|'s [=request/client=]'s [=environment settings object/cross-origin isolated capability=].
```
jakearchibald, line 2978: Has someone familiar with isolation looked at this? Two timings are taken, once from "in parallel", and another from the service worker thread. Neither of these are the request's client, so does the request's client's cross-origin isolated capability apply here?

docs/index.bs
```diff
@@ -3020,6 +3025,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
           1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
           1. Return null.
       1. If |activeWorker|'s <a>state</a> is "`activating`", wait for |activeWorker|'s <a>state</a> to become "`activated`".
+      1. Let |workerStartTime| be the  [=coarsened shared current time=] given <var>crossOriginIsolatedCapability</var>.
```
jakearchibald, line 3028:
```suggestion
      1. Let |workerStartTime| be the [=coarsened shared current time=] given |crossOriginIsolatedCapability|.
```
Nit

jakearchibald, line 3028: The previous step (`If |activeWorker|'s <a>state</a> is "activating"`) probably should be included in this measurement, since we've decided we need a service worker, but it might not be ready. If the site is seeing long delays due to activation delays, that feels important.

docs/index.bs
```diff
@@ -3035,6 +3041,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
               1. If |request| is a <a>non-subresource request</a> and |request|'s [=request/destination=] is not {{RequestDestination/"report"}}, initialize |e|'s {{FetchEvent/resultingClientId}} attribute to |reservedClient|'s [=environment/id=], and to the empty string otherwise.
               1. If |request| is a <a>navigation request</a>, initialize |e|'s {{FetchEvent/replacesClientId}} attribute to |request|'s [=request/replaces client id=], and to the empty string otherwise.
               1. Initialize |e|’s {{FetchEvent/handled}} to |eventHandled|.
+              1. If |request| is a <a>navigation request</a>, then:
```
jakearchibald, line 3044:
```suggestion
              1. If |request| is a [=navigation request=], then:
```

I realise we're not consistent with the syntax, but we're trying to use shorthand as we make changes.

jakearchibald (commented):
Thanks for the PR, and sorry for the delay reviewing. @mfalken and I took a look.

It isn't clear to me how these values will be read. I'm slightly worried that there's one service worker, therefore one "start time" and one "ready time", but many requests. It feels like "start time" and "ready time" might be overwritten by another request in a race.

Maybe these values should instead be associated with the response, since the timing is from the perspective of the request/response. But I'm just guessing, because I'm not sure how these values will be used.

For instance, if the service worker was started a year ago, and never closed for some reason, should "worker start" be the time a year ago, or should it be the time a given request decided it needed a service worker?

wanderview:
> For instance, if the service worker was started a year ago, and never closed for some reason, should "worker start" be the time a year ago, or should it be the time a given request decided it needed a service worker?

Yea, I thought it would be the time the request decided it needed a service worker if the worker was already running.

docs/index.bs
noamr: It's how things are done in FETCH. Not sure if it applies to service workers. in the same way. @annevk ?

noamr:
> Thanks for the PR, and sorry for the delay reviewing. @mfalken and I took a look.
> 
> It isn't clear to me how these values will be read. I'm slightly worried that there's one service worker, therefore one "start time" and one "ready time", but many requests. It feels like "start time" and "ready time" might be overwritten by another request in a race.

OK, I will change it to mean the "first" start/ready time. These values are only meant for Navigation Timing, I'll prepare a Navigation Timing PR that reads them to make it clearer. (they would have inter-dependencies so this one would have to go in first).

> 
> Maybe these values should instead be associated with the response, since the timing is from the perspective of the request/response. But I'm just guessing, because I'm not sure how these values will be used.
> 
> For instance, if the service worker was started a year ago, and never closed for some reason, should "worker start" be the time a year ago, or should it be the time a given request decided it needed a service worker?

The service worker's start time would be a year ago, but Navigation Timing would expose the later time between that time and the time FETCH has handed control over to the service worker.

noamr:
I've refactored the PR, I agree that it makes much more sense to associate the info with the response rather than with the worker.
https://github.com/w3c/navigation-timing/pull/143 is the navigation timing PR that shows how this info is going to be used. To complete the picture, we would need to modify two additional specs:

- FETCH would have to send SW-FETCH the `crossOriginIsolatedCapability` as the second argument.
- HTML would have to send the response's `service worker timing` when creating the navigation timing entry.

There are probably more ways to accomplish the same results, I went with the way that made the most sense to me but I'm open to refactor given an alternative approach.

docs/index.bs
```diff
@@ -193,15 +193,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
           * <dfn export id="dfn-functional-events">Functional events</dfn>: {{fetch!!event}} and the [=events=] defined by other specifications that <a href="#extensibility">extend</a> the Service Workers specification. (See the <a href="#execution-context-events">list</a>.)
           * {{message!!event}} and {{messageerror!!event}}.
     </section>
-  </section>
+    <section>
+      <h4 id="service-worker-timing">Timing</h4>
+
+      Service workers mark certain points in times, that are later exposed by the {{PerformanceNavigationTiming|navigation timing}} API.
+
+      A [=/response=] has an associated null or [=service worker timing info=] <dfn export for=response>service worker timing</dfn>, initially set to null.
```
annevk, line 209: I would prefer if we defined this in Fetch and SW just does the marking. Mainly to keep response defined in a single place.

Is it correct that this information gets lost if a network error is returned instead at some point?

noamr: > I would prefer if we defined this in Fetch and SW just does the marking. Mainly to keep response defined in a single place.

OK. I was trying to make it so that FETCH doesn't have to be modified every time we add a new service-worker metric. Should the struct live in FETCH as well?

> Is it correct that this information gets lost if a network error is returned instead at some point?

This is only exposed through Navigation Timing, if there's a network response the document won't be loaded and these values would be meaningless. In other words, correct.

annevk: I think it's okay to keep the struct here to make that easier.

noamr: @jakearchibald I want to make sure we're OK with how this whole process works, and then I'll publish:

1. A SW PR that defines the new struct
2. A FETCH PR that associates the struct with a response and adds arguments to the SW FETCH call
3. The marking part of the SW PR

docs/index.bs
annevk: This is interesting as there might be a mismatch between the service worker agent and the requesting agent. I suspect we have to consult both and only if they are both true you would get the high-resolution time. Does that make sense?

That might be a change from what is implemented today as that probably only considers the requesting agent.

(Fetch uses the requesting agent exclusively as it doesn't create any agents of its own.)

docs/index.bs
```diff
@@ -193,15 +193,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
       The Service Workers specification defines <dfn export id="dfn-service-worker-events">service worker events</dfn> (each of which is an [=event=]) that include (see the <a href="#execution-context-events">list</a>):
           * <dfn export id="dfn-lifecycle-events">Lifecycle events</dfn>: {{install!!event}} and {{activate!!event}}.
           * <dfn export id="dfn-functional-events">Functional events</dfn>: {{fetch!!event}} and the [=events=] defined by other specifications that <a href="#extensibility">extend</a> the Service Workers specification. (See the <a href="#execution-context-events">list</a>.)
           * {{message!!event}} and {{messageerror!!event}}.
     </section>
-  </section>
+    <section>
+      <h4 id="service-worker-timing">Timing</h4>
+
+      Service workers mark certain points in times, that are later exposed by the {{PerformanceNavigationTiming|navigation timing}} API.
```
jakearchibald, line 207:
```suggestion
      Service workers mark certain points in time that are later exposed by the {{PerformanceNavigationTiming|navigation timing}} API.
```

docs/index.bs
jakearchibald, line 209: For consistency:

```suggestion
      A [=/response=] has an associated <dfn export for=response>service worker timing</dfn> (a [=service worker timing info=] or null). It is initially set to null.
```

docs/index.bs
```diff
@@ -193,15 +193,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+      A [=/response=] has an associated null or [=service worker timing info=] <dfn export for=response>service worker timing</dfn>, initially set to null.
+
+      <p>A <dfn export>service worker timing info</dfn> is a <a for=/>struct</a>. It has the
+      following <a for=struct>items</a>:
```
jakearchibald, line 212:
```suggestion
      A <dfn export>service worker timing info</dfn> is a [=/struct=]. It has the following [=struct/items=]:
```

docs/index.bs
```diff
@@ -193,15 +193,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+
+      <dl>
+      <dt><dfn export for="service worker timing info">start time</dfn> (default 0)
+      <dt><dfn export for="service worker timing info">fetch event dispatch time</dfn> (default 0)
+      <dd>A {{DOMHighResTimeStamp}}.
+      </dl>
```
jakearchibald, line 218:
```suggestion
      <section dfn-for="service worker timing info">
        : <dfn export>start time</dfn>
        :: A {{DOMHighResTimeStamp}}, initially 0.
        : <dfn export>fetch event dispatch time</dfn>
        :: A {{DOMHighResTimeStamp}}, initially 0.
      </section>
```

I haven't fully tested the above parses properly, but you see what I'm going for.

docs/index.bs
```diff
@@ -2958,6 +2974,7 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
       : Input
       :: |request|, a [=/request=]
+      :: |crossOriginIsolatedCapability|, a boolean
```
jakearchibald, line 2977: It isn't clear to me which client this isolation refers to. Maybe it should just be `useHighResPerformanceTimers` or something which makes the scope of its usage clear?

docs/index.bs
```diff
@@ -193,15 +193,31 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
         * Detects abnormal operation: such as infinite loops and tasks exceeding imposed time limits (if any) while handling the events.
     </section>
 
-    <section>
+   <section>
```
jakearchibald, line 196: Unintentional change?

jakearchibald (changes requested):
Minor changes, but generally LGTM

docs/index.bs
```diff
@@ -201,7 +201,21 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+      A <dfn export>service worker timing info</dfn> is a [=/struct=]. It has the following [=struct/items=]:
+
+      <section dfn-for="service worker timing info">
+        : <dfn export>start time</dfn>
+        :: A {{DOMHighResTimeStamp}}, initially 0.
+        : <dfn export>fetch event dispatch time</dfn>
+        :: A {{DOMHighResTimeStamp}}, initially 0.
+      </section>
+    </section>
+   </section>
```
jakearchibald, line 218:
```suggestion
  </section>
```

jakearchibald:
Give me a nudge when this is ready to go

noamr:
> Give me a nudge when this is ready to go

Thanks, will do! first I need to wait for the refs to apply and post a FETCH PR.

noamr:
> Give me a nudge when this is ready to go

Nudge @jakearchibald :)

noamr:
> What about [#1575 (comment)](https://github.com/w3c/ServiceWorker/pull/1575#discussion_r612409033)?

Fixed in new revision.

docs/index.bs
```diff
@@ -3046,6 +3048,9 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
       1. If the result of running the [=Should Skip Event=] algorithm with "fetch" and |activeWorker| is true, then:
           1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
           1. Return null.
+      1. If |useHighResPerformanceTimers| is true, then set |useHighResPerformanceTimers| to |activeWorker|'s  
+         [=service worker/global object=]'s [=WorkerGlobalScope/cross-origin isolated capability=].
```
jakearchibald, line 3052:
```suggestion
      1. If |useHighResPerformanceTimers| is true, then set |useHighResPerformanceTimers| to |activeWorker|'s [=service worker/global object=]'s [=WorkerGlobalScope/cross-origin isolated capability=].
```

jakearchibald (approved):

jakearchibald:
Thank you @noamr!

PR #1590, opened 2021-04-29
Service worker struct should be exported without a 'for' attribute

noamr:
***
<a href="https://pr-preview.s3.amazonaws.com/noamr/ServiceWorker/pull/1590.html" title="Last updated on Apr 29, 2021, 9:48 AM UTC (5f55eaf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1590/b807be1...noamr:5f55eaf.html" title="Last updated on Apr 29, 2021, 9:48 AM UTC (5f55eaf)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -210,12 +210,12 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
 
     A <dfn export>service worker timing info</dfn> is a [=/struct=]. It has the following [=struct/items=]:
 
-      <section dfn-for="service worker timing info">
-        : <dfn export>start time</dfn>
-        :: A {{DOMHighResTimeStamp}}, initially 0.
-        : <dfn export>fetch event dispatch time</dfn>
-        :: A {{DOMHighResTimeStamp}}, initially 0.
-      </section>
+    <section dfn-for="service worker timing info">
+      : <dfn export>start time</dfn>
+      :: A {{DOMHighResTimeStamp}}, initially 0.
+      : <dfn export>fetch event dispatch time</dfn>
+      :: A {{DOMHighResTimeStamp}}, initially 0.
+    </section>
   </section>
 
   <section dfn-for="service worker registration">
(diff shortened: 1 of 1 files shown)


noamr:
@jakearchibald r?
@ylafon I think I need another IPR form thingy.

jakearchibald:
Ah, this is because it's in `<section dfn-for="service worker">`. It probably shouldn't be a subsection of that.

Can you move it out a level, so it's after section 2.1, rather than within section 2.1? If you do that, you don't need the `for=""`.

noamr:
> Ah, this is because it's in `<section dfn-for="service worker">`. It probably shouldn't be a subsection of that.
> 
> Can you move it out a level, so it's after section 2.1, rather than within section 2.1? If you do that, you don't need the `for=""`.

Done

docs/index.bs
```diff
@@ -201,20 +201,21 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
+    Service workers mark certain points in time that are later exposed by the {{PerformanceNavigationTiming|navigation timing}} API.
+
+    A <dfn export>service worker timing info</dfn> is a [=/struct=]. It has the following [=struct/items=]:
 
       <section dfn-for="service worker timing info">
         : <dfn export>start time</dfn>
         :: A {{DOMHighResTimeStamp}}, initially 0.
         : <dfn export>fetch event dispatch time</dfn>
         :: A {{DOMHighResTimeStamp}}, initially 0.
       </section>
```
jakearchibald, line 218:
```suggestion
    <section dfn-for="service worker timing info">
      : <dfn export>start time</dfn>
      :: A {{DOMHighResTimeStamp}}, initially 0.
      : <dfn export>fetch event dispatch time</dfn>
      :: A {{DOMHighResTimeStamp}}, initially 0.
    </section>
```

docs/index.bs
```diff
@@ -201,20 +201,21 @@ spec: webappsec-referrer-policy; urlPrefix: https://w3c.github.io/webappsec-refe
...
-    <section>
-      <h4 id="service-worker-timing">Timing</h4>
+  </section>
 
-      Service workers mark certain points in time that are later exposed by the {{PerformanceNavigationTiming|navigation timing}} API.
+  <section>
+    <h3 id="service-worker-timing">Serview Worker Timing</h3>
 
-      A <dfn export>service worker timing info</dfn> is a [=/struct=]. It has the following [=struct/items=]:
+    Service workers mark certain points in time that are later exposed by the {{PerformanceNavigationTiming|navigation timing}} API.
```
jakearchibald, line 209:
```suggestion
    Service workers mark certain points in time that are later exposed by the [=PerformanceNavigationTiming!!idl|navigation timing=] API.
```

I haven't tested if this works, but it should mean that "navigation timing" isn't displayed as if it's a method/object.

docs/index.bs
noamr: I think this doesn't pass build.

jakearchibald (approved):

Issue #1594, opened 2021-05-20
How does ServiceWorker interact with non-"fully active" documents?


rakina:
Currently the ServiceWorker specification does not specify how it interacts with non-"[fully active](https://html.spec.whatwg.org/multipage/browsers.html#fully-active)" (bfcached) documents/window clients.
* Iterating over all clients does not filter out window clients with non-fully active documents. Example: [Clients.get()](https://w3c.github.io/ServiceWorker/#clients-get), [Clients.matchAll()](https://w3c.github.io/ServiceWorker/#clients-matchall), [Clients.claim()](https://w3c.github.io/ServiceWorker/#clients-claim)
* Actions on window clients whose document is no longer fully active can still happen. Example: [Client.postMessage()](https://w3c.github.io/ServiceWorker/#client-postmessage)

We recommend following the [WIP guide](https://github.com/w3ctag/design-principles/pull/317#pullrequestreview-663440698) we've made on handling non-fully active documents, by adding "fully active" checks so that we won't include clients with non-fully active documents, and if necessary trigger discarding of the document (e.g. for Clients.claim()). This is already how it behaves in Chrome's implementation of BFCache, and @hiroshige-g will also publish some WPTs soon to show how it behaves in various browsers.

cc @domenic @fergal @altimin

wanderview:
> Clients.claim(): When called while a page is BFCached, the page is evicted from BFCache (Chrome) or the page is ignored/not claimed (Firefox/Safari).

I think clients.claim() not evicting on firefox is a known bug.  Although maybe I am thinking of this [one](https://bugzilla.mozilla.org/show_bug.cgi?id=1418489).

> Other: on Firefox, after a page is restored from BFCache, the page seems uncontrolled by service workers (even if it was controlled before BFCached). Is this a known issue?

That seems surprising.  @asutherland?

asutherland:
Thank you for bringing this to my attention!  I have filed https://bugzilla.mozilla.org/show_bug.cgi?id=1735875 for us to investigate further.  It's very cool to see the RemoteContext mechanism at working, especially since other mechanisms like BroadcastChannel could interfere with BFCache!

rakina:
FYI we're aiming to land the WPTs at https://github.com/web-platform-tests/wpt/pull/31082 soon, PTAL if you have time @asutherland @smaug---- :)

PR #1597, opened 2021-06-10
Handle non-fully active clients.

mfalken:
For #1594


mfalken:
There are tests for this in progress at https://chromium-review.googlesource.com/c/chromium/src/+/2798554, along with many other bfcache tests. So maybe this can be landed?

I'm not sure why the PR Preview is failing to build.

PR #1618, opened 2021-11-10
Use fetch algorithms when getting VARY header values. (Fixes #1609)

### LINKED ISSUE #1609: Cache API Vary header processing should be clarified
> Let fieldValues be the list containing the elements corresponding to the field-values of the Vary header for the value of the header with name `Vary`.

This needs to be rewritten on top of https://fetch.spec.whatwg.org/#concept-header-list-get (or perhaps the algorithm below "get, decode, and split") as this sentence makes little sense.

wanderview:


wanderview:
@jakearchibald PTAL at this short pull request to use the proper algorithm from fetch spec.

yoshisatoyanagisawa:
Do you still want to move forward the review?
At a brief look at the PR, it looks good, though.

PR #1619, opened 2021-11-10
Make cache.put() drain the internal response body. (Fixes #1617)

### LINKED ISSUE #1617: cache API put should read internal response body of filtered responses
Step 11.1 of cache put:

https://w3c.github.io/ServiceWorker/#cache-put

Tries to get the body stream of the response in order to drain it.  This does not work properly if the response is an opaque filtered response since the body is hidden.  The algorithm should instead get the body of the filtered internal response.

wanderview:


wanderview:
@jakearchibald PTAL.  This is one is very small as well.  Thanks.

yoshisatoyanagisawa:
Do you still want to move forward the review?
At a brief look at the PR, it looks good, though.

Issue #1671, opened 2023-02-13
Skip service worker no-op fetch handler


yoshisatoyanagisawa:
Hi Service Worker WG!

Let me ask your feedback on our proposal:

Some sites have a no-op (no operation) fetch handler (e.g. `onfetch = () => {}`). Since having the fetch handler was one of the requirements to be [a progressive web app (PWA)](https://developer.mozilla.org/docs/Web/Progressive_web_apps), we assume that they did that to make their site recognized as PWA. However, it only brings overheads to start a service worker and execute a no-op handler without bringing any feature benefits like caching or offline capabilities because the code does nothing. To make the navigation to such pages faster, we would like to omit the service worker start and the handler execution from the navigation critical path if a user agent identifies that the service worker fetch handler is no-op.

This is mostly a behind-the-scenes performance improvement. However, it can subtly change the behavior in edge cases, such as dynamically-updated fetch handlers.

- [Explainer](https://github.com/yoshisatoyanagisawa/service-worker-skip-no-op-fetch-handler/blob/main/README.md)

Currently, the explainer is in my private repository, but I am happy to move the explainer into this repository if the group is interested in centralizing discussion

jakearchibald:
I'm happy with this proposal. There's an observable change in behaviour for cases where the fetch listener changes over time (eg with `Math.random()` or `Date.now()`), but these cases are somewhat unpredictable anyway, because it depends on the service worker terminating and starting up to pick up the new function.

PR #1672, opened 2023-02-15
Skip executing fetch handlers if it is no-op.

yoshisatoyanagisawa:
This is the specification change for the proposal (https://github.com/w3c/ServiceWorker/issues/1671).

I got comments and/or contribution from following folks when I was drafting this specification update:
+@jakearchibald 
+@hiroshige-g

Let me ask feedbacks from the WG.
Thank you.

***
<a href="https://pr-preview.s3.amazonaws.com/yoshisatoyanagisawa/ServiceWorker/pull/1672.html" title="Last updated on Mar 17, 2023, 9:11 AM UTC (0d63641)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1672/25f60ac...yoshisatoyanagisawa:0d63641.html" title="Last updated on Mar 17, 2023, 9:11 AM UTC (0d63641)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -23,26 +23,34 @@ Markup Shorthands: css no, markdown yes
 
 <pre class="link-defaults">
 spec: html;
-    type: dfn; text: task queues; for: /
     type: element; text: link
-    type: dfn; text: event loop; for: /
+    type: dfn
+        text: task queues; for: /
+        text: event loop; for: /
+        text: script
 
 spec: dom;
     type: interface; text: Document
+    type: dfn; text: event
 
 spec: fetch;
     type: dfn
         for: /; text: fetch
         text: empty
+        text: fetch controller
+
 spec: infra;
     type: dfn;
         text: list
         text: queue
         for: set; text: append
         for: list; text: append
+
 spec: webidl;
     type: dfn;
         text: resolve;
+
+spec:csp-next; type:dfn; text:enforced
 </pre>
 
 <pre class="anchors">
@@ -146,7 +154,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
   Developers using the <a href="https://developer.mozilla.org/docs/Web/API/Window/applicationCache">HTML5 Application Cache</a> have also <a href="http://alistapart.com/article/application-cache-is-a-douchebag">reported that several attributes</a> of the design contribute to <a href="http://alistapart.com/article/application-cache-is-a-douchebag#section6">unrecoverable errors</a>. A key design principle of the [=/service worker=] is that errors should *always* be recoverable. Many details of the update process of [=/service workers=] are designed to avoid these hazards.
 
-  [=/Service workers=] are started and kept alive by their relationship to events, not documents. This design borrows heavily from developer and vendor experience with <a>Shared Workers</a> and <a href="https://developer.chrome.com/extensions/background_pages">Chrome Background Pages</a>. A key lesson from these systems is the necessity to time-limit the execution of background processing contexts, both to conserve resources and to ensure that background context loss and restart is top-of-mind for developers. As a result, [=/service workers=] bear more than a passing resemblance to <a href="https://developer.chrome.com/extensions/event_pages">Chrome Event Pages</a>, the successor to Background Pages. [=/Service workers=] may be started by user agents *without an attached document* and may be killed by the user agent at nearly any time. Conceptually, [=/service workers=] can be thought of as Shared Workers that can start, process events, and die without ever handling messages from documents. Developers are advised to keep in mind that [=/service workers=] may be started and killed many times a second.
+  [=/Service workers=] are started and kept alive by their relationship to events, not documents. This design borrows heavily from developer and vendor experience with [[HTML#shared-workers-and-the-sharedworker-interface|shared workers]] and <a href="https://developer.chrome.com/extensions/background_pages">Chrome Background Pages</a>. A key lesson from these systems is the necessity to time-limit the execution of background processing contexts, both to conserve resources and to ensure that background context loss and restart is top-of-mind for developers. As a result, [=/service workers=] bear more than a passing resemblance to <a href="https://developer.chrome.com/extensions/event_pages">Chrome Event Pages</a>, the successor to Background Pages. [=/Service workers=] may be started by user agents *without an attached document* and may be killed by the user agent at nearly any time. Conceptually, [=/service workers=] can be thought of as Shared Workers that can start, process events, and die without ever handling messages from documents. Developers are advised to keep in mind that [=/service workers=] may be started and killed many times a second.
 
   [=/Service workers=] are generic, event-driven, time-limited script contexts that run at an origin. These properties make them natural endpoints for a range of runtime services that may outlive the context of a particular document, e.g. handling push notifications, background data synchronization, responding to resource requests from other origins, or receiving centralized updates to expensive-to-calculate data (e.g., geolocation or gyroscope).
 </section>
@@ -208,9 +216,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       <h4 id="service-worker-events">Events</h4>
 
       The Service Workers specification defines <dfn export id="dfn-service-worker-events">service worker events</dfn> (each of which is an [=event=]) that include (see the <a href="#execution-context-events">list</a>):
-          * <dfn export id="dfn-lifecycle-events">Lifecycle events</dfn>: {{install!!event}} and {{activate!!event}}.
-          * <dfn export id="dfn-functional-events">Functional events</dfn>: {{fetch!!event}} and the [=events=] defined by other specifications that <a href="#extensibility">extend</a> the Service Workers specification. (See the <a href="#execution-context-events">list</a>.)
-          * {{message!!event}} and {{messageerror!!event}}.
+          * <dfn export id="dfn-lifecycle-events">Lifecycle events</dfn>: {{ServiceWorkerGlobalScope/install!!event}} and {{ServiceWorkerGlobalScope/activate!!event}}.
+          * <dfn export id="dfn-functional-events">Functional events</dfn>: {{ServiceWorkerGlobalScope/fetch!!event}} and the [=events=] defined by other specifications that <a href="#extensibility">extend</a> the Service Workers specification. (See the <a href="#execution-context-events">list</a>.)
+          * {{ServiceWorkerGlobalScope/message!!event}} and {{ServiceWorkerGlobalScope/messageerror!!event}}.
     </section>
   </section>
 
@@ -258,7 +266,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     A [=/service worker registration=] has an associated <dfn export>navigation preload header value</dfn>, which is a [=byte sequence=]. It is initially set to \`<code>true</code>\`.
 
-    A [=/service worker registration=] is said to be <dfn export id="dfn-service-worker-registration-unregistered">unregistered</dfn> if [=registration map=][this [=service worker registration=]'s ([=service worker registration/storage key=], [=URL serializer|serialized=] [=service worker registration/scope url=])] is not this [=/service worker registration=].
+    A [=/service worker registration=] is said to be <dfn export id="dfn-service-worker-registration-unregistered">unregistered</dfn> if [=registration map=][this [=/service worker registration=]'s ([=service worker registration/storage key=], [=URL serializer|serialized=] [=service worker registration/scope url=])] is not this [=/service worker registration=].
 
     <section>
       <h4 id="service-worker-registration-lifetime">Lifetime</h4>
@@ -316,7 +324,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       Otherwise, if the created [=document=]'s [=/origin=] is an [=opaque origin=] or not the [=same origin|same=] as its creator [=document=]'s [=/origin=], the [=window client=]'s [=active service worker=] is set to null.
       Otherwise, it is set to the creator [=document=]'s [=/service worker client=]'s [=active service worker=].
 
-      Note: For an initial [=navigate|navigation=] with [=replacement enabled=], the initial [=window client=] that was [created](https://html.spec.whatwg.org/#set-up-a-window-environment-settings-object) when the [=/browsing context=] was [created](https://html.spec.whatwg.org/#creating-a-new-browsing-context) is reused, but the [=active service worker=] is determined by the same behavior as above.
+      Note: For an initial replacement [=navigate|navigation=], the initial [=window client=] that was [created](https://html.spec.whatwg.org/#set-up-a-window-environment-settings-object) when the [=/browsing context=] was [created](https://html.spec.whatwg.org/#creating-a-new-browsing-context) is reused, but the [=active service worker=] is determined by the same behavior as above.
 
       Note: <a element-attr for=iframe lt=sandbox>Sandboxed</a> <{iframe}>s without the sandboxing directives, `allow-same-origin` and `allow-scripts`, result in having the [=active service worker=] value of null as their [=/origin=] is an [=opaque origin=].
     </section>
@@ -408,7 +416,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     A {{ServiceWorker}} object represents a [=/service worker=]. Each {{ServiceWorker}} object is associated with a [=/service worker=]. Multiple separate objects implementing the {{ServiceWorker}} interface across documents and workers can all be associated with the same [=/service worker=] simultaneously.
 
-    A {{ServiceWorker}} object has an associated {{ServiceWorkerState}} object which is itself associated with [=/service worker=]'s <a>state</a>.
+    A {{ServiceWorker}} object has an associated {{ServiceWorkerState}} object which is itself associated with [=/service worker=]'s [=service worker/state=].
 
     <section>
       <h4 id="service-worker-creation">Getting {{ServiceWorker}} instances</h4>
@@ -487,11 +495,11 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
                 1. Let |destination| be the {{ServiceWorkerGlobalScope}} object associated with |serviceWorker|.
                 1.  Let |deserializeRecord| be <a abstract-op>StructuredDeserializeWithTransfer</a>(|serializeWithTransferResult|, |destination|'s [=global object/Realm=]).
 
-                    If this throws an exception, let |e| be the result of [=creating an event=] named {{messageerror!!event}}, using {{ExtendableMessageEvent}}, with the {{ExtendableMessageEvent/origin}} attribute initialized to |origin| and the {{ExtendableMessageEvent/source}} attribute initialized to |source|.
+                    If this throws an exception, let |e| be the result of [=creating an event=] named {{ServiceWorkerGlobalScope/messageerror!!event}}, using {{ExtendableMessageEvent}}, with the {{ExtendableMessageEvent/origin}} attribute initialized to |origin| and the {{ExtendableMessageEvent/source}} attribute initialized to |source|.
                 1. Else:
                     1. Let |messageClone| be |deserializeRecord|.\[[Deserialized]].
                     1. Let |newPorts| be a new [=frozen array type|frozen array=] consisting of all {{MessagePort}} objects in |deserializeRecord|.\[[TransferredValues]], if any, maintaining their relative order.
-                    1. Let |e| be the result of [=creating an event=] named {{message!!event}}, using {{ExtendableMessageEvent}}, with the {{ExtendableMessageEvent/origin}} attribute initialized to |origin|, the {{ExtendableMessageEvent/source}} attribute initialized to |source|, the {{ExtendableMessageEvent/data}} attribute initialized to |messageClone|, and the {{ExtendableMessageEvent/ports}} attribute initialized to |newPorts|.
+                    1. Let |e| be the result of [=creating an event=] named {{ServiceWorkerGlobalScope/message!!event}}, using {{ExtendableMessageEvent}}, with the {{ExtendableMessageEvent/origin}} attribute initialized to |origin|, the {{ExtendableMessageEvent/source}} attribute initialized to |source|, the {{ExtendableMessageEvent/data}} attribute initialized to |messageClone|, and the {{ExtendableMessageEvent/ports}} attribute initialized to |newPorts|.
                 1. [=Dispatch=] |e| at |destination|.
                 1. Invoke [=Update Service Worker Extended Events Set=] with |serviceWorker| and |e|.
     </section>
@@ -625,7 +633,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         1. Let |registration| be the [=ServiceWorkerRegistration/service worker registration=].
         1. Let |newestWorker| be the result of running <a>Get Newest Worker</a> algorithm passing |registration| as its argument.
         1. If |newestWorker| is null, return [=a promise rejected with=] an "{{InvalidStateError}}" {{DOMException}} and abort these steps.
-        1. If [=this=]'s [=relevant global object=] |globalObject| is a {{ServiceWorkerGlobalScope}} object, and |globalObject|'s associated [=ServiceWorkerGlobalScope/service worker=]'s <a>state</a> is "`installing`", return [=a promise rejected with=] an "{{InvalidStateError}}" {{DOMException}} and abort these steps.
+        1. If [=this=]'s [=relevant global object=] |globalObject| is a {{ServiceWorkerGlobalScope}} object, and |globalObject|'s associated [=ServiceWorkerGlobalScope/service worker=]'s [=service worker/state=] is "`installing`", return [=a promise rejected with=] an "{{InvalidStateError}}" {{DOMException}} and abort these steps.
         1. Let |promise| be a <a>promise</a>.
         1. Let |job| be the result of running <a>Create Job</a> with *update*, |registration|'s [=service worker registration/storage key=], |registration|'s [=service worker registration/scope url=], |newestWorker|'s [=service worker/script url=], |promise|, and [=this=]'s <a>relevant settings object</a>.
         1. Set |job|'s <a>worker type</a> to |newestWorker|'s [=service worker/type=].
@@ -724,7 +732,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     A {{ServiceWorkerContainer}} object has an associated <dfn for="ServiceWorkerContainer">ready promise</dfn> (a [=promise=] or null). It is initially null.
 
-    A {{ServiceWorkerContainer}} object has a <a>task source</a> called the <dfn export id="dfn-client-message-queue" for="ServiceWorkerContainer">client message queue</dfn>, initially empty. A [=ServiceWorkerContainer/client message queue=] can be enabled or disabled, and is initially disabled. When a {{ServiceWorkerContainer}} object's [=ServiceWorkerContainer/client message queue=] is enabled, the <a>event loop</a> *must* use it as one of its <a>task sources</a>. When the {{ServiceWorkerContainer}} object's <a>relevant global object</a> is a {{Window}} object, all <a>tasks</a> <a lt="queue a task">queued</a> on its [=ServiceWorkerContainer/client message queue=] *must* be associated with its <a>relevant settings object</a>'s <a>responsible document</a>.
+    A {{ServiceWorkerContainer}} object has a <a>task source</a> called the <dfn export id="dfn-client-message-queue" for="ServiceWorkerContainer">client message queue</dfn>, initially empty. A [=ServiceWorkerContainer/client message queue=] can be enabled or disabled, and is initially disabled. When a {{ServiceWorkerContainer}} object's [=ServiceWorkerContainer/client message queue=] is enabled, the <a>event loop</a> *must* use it as one of its <a>task sources</a>. When the {{ServiceWorkerContainer}} object's <a>relevant global object</a> is a {{Window}} object, all <a>tasks</a> <a lt="queue a task">queued</a> on its [=ServiceWorkerContainer/client message queue=] *must* be associated with its <a>relevant settings object</a>'s [=associated document=].
 
     <section algorithm="navigator-service-worker-controller">
       <h4 id="navigator-service-worker-controller">{{ServiceWorkerContainer/controller}}</h4>
@@ -800,7 +808,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         1. Let |promise| be [=a new promise=].
         1. Run the following steps [=in parallel=]:
             1. Let |registrations| be a new [=list=].
-            1. [=map/For each=] (|storage key|, |scope|) → |registration| of [=registration map=]:
+            1. [=list/For each=] (|storage key|, <var ignore>scope</var>) → |registration| of [=registration map=]:
                 1. If |storage key| [=storage key/equals=] |client storage key|, then [=append=] |registration| to |registrations|.
             1. [=Queue a task=] on |promise|'s [=relevant settings object=]'s [=responsible event loop=], using the [=DOM manipulation task source=], to run the following steps:
                 1. Let |registrationObjects| be a new [=list=].
@@ -1103,7 +1111,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
       The following are the <a>event handlers</a> (and their corresponding <a>event handler event types</a>) that *must* be supported, as <a>event handler IDL attributes</a>, by all objects implementing the {{ServiceWorkerGlobalScope}} interface:
 
-      <table class="data">
+      <table class="data" dfn-for="ServiceWorkerGlobalScope">
         <thead>
           <tr>
             <th><a>event handler</a></th>
@@ -1112,24 +1120,24 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         </thead>
         <tbody>
           <tr>
-            <td><dfn attribute for="ServiceWorkerGlobalScope"><code>oninstall</code></dfn></td>
-            <td>{{install!!event}}</td>
+            <td><dfn attribute><code>oninstall</code></dfn></td>
+            <td>{{ServiceWorkerGlobalScope/install!!event}}</td>
           </tr>
           <tr>
-            <td><dfn attribute for="ServiceWorkerGlobalScope"><code>onactivate</code></dfn></td>
-            <td>{{activate!!event}}</td>
+            <td><dfn attribute><code>onactivate</code></dfn></td>
+            <td>{{ServiceWorkerGlobalScope/activate!!event}}</td>
           </tr>
           <tr>
-            <td><dfn attribute for="ServiceWorkerGlobalScope"><code>onfetch</code></dfn></td>
-            <td>{{fetch!!event}}</td>
+            <td><dfn attribute><code>onfetch</code></dfn></td>
+            <td>{{ServiceWorkerGlobalScope/fetch!!event}}</td>
           </tr>
           <tr>
-            <td><dfn attribute for="ServiceWorkerGlobalScope"><code>onmessage</code></dfn></td>
-            <td>{{message!!event}}</td>
+            <td><dfn attribute><code>onmessage</code></dfn></td>
+            <td>{{ServiceWorkerGlobalScope/message!!event}}</td>
           </tr>
           <tr>
-            <td><dfn attribute for="ServiceWorkerGlobalScope"><code>onmessageerror</code></dfn></td>
-            <td>{{messageerror!!event}}</td>
+            <td><dfn attribute><code>onmessageerror</code></dfn></td>
+            <td>{{ServiceWorkerGlobalScope/messageerror!!event}}</td>
           </tr>
         </tbody>
       </table>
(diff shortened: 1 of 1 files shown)


docs/index.bs
```diff
@@ -2970,16 +2972,34 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               1. For each |eventType| of |settingsObject|'s [=environment settings object/global object=]'s associated list of <a>event listeners</a>' event types:
                   1. [=set/Append=] |eventType| to |workerGlobalScope|'s associated [=ServiceWorkerGlobalScope/service worker=]'s <a>set of event types to handle</a>.
 
-                  Note: If the global object's associated list of event listeners does not have any event listener added at this moment, the service worker's set of event types to handle remains an empty set. The user agents are encouraged to show a warning that the event listeners must be added on the very first evaluation of the worker script.
+                  Note: If the global object's associated list of event listeners does not have any event listener added at this moment, the service worker's set of event types to handle remains an empty set.
 
               1. Set |script|'s <a>has ever been evaluated flag</a>.
+              1. If the [=No-Op Handler identification=] algorithm with |serviceWorker| returns true, the [=has had non-no-op fetch listeners flag=] is set to false.
```
jakearchibald, line 2978: This seems backwards. I don't think we ever want to unset this flag. Or have I misunderstood the design?


```suggestion
              1. If [=No-Op Handler identification=] given |serviceWorker| is false, then set |serviceWorker|'s [=has had non-no-op fetch listeners flag=].
```

When referring to a property like `[=has had non-no-op fetch listeners flag=]`, be clear about which object you're referring to.

docs/index.bs
```diff
@@ -2970,16 +2972,34 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
 
+  <section algorithm>
+    <h3 id="no-op-handler-identification-algorithm"><dfn>No-Op Handler identification</dfn></h3>
+
+      : Input
+      :: |serviceWorker|, a [=/service worker=]
+      : Output
+      :: a boolean
+
+      1. If |service worker|'s set of event types to handle does not contain "fetch", then returns false.
```
jakearchibald, line 2994: `|service worker|` should be `|serviceWorker|` (as defined in the input)

References need to be linked properly here.

docs/index.bs
```diff
@@ -2970,16 +2972,34 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      1. If the callback's function bodies (https://tc39.es/ecma262/#prod-FunctionBody) of all fetch event listener in |service worker| are empty, then the user agent *may* return true.
```
jakearchibald, line 2995: Things like "callback" need to be referenced properly. Eg, see https://dom.spec.whatwg.org/#event-listener-callback - you'll need to handle cases where the 'callback' is actually an `EventListener` object.

`|service worker|` should be `|serviceWorker|` (as defined in the input)

Rather than just paste the link of https://tc39.es/ecma262/#prod-FunctionBody, create a proper reference. See the `<pre class="anchors">` section at the top of the document.

"of all fetch event listener" needs to be defined correctly in terms of https://dom.spec.whatwg.org/#concept-event-listener.

"empty" needs to be defined more robustly.

docs/index.bs
```diff
@@ -2970,16 +2972,34 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+      Note: it detects something like `onfetch = () => {}`. Some sites have a fetch event listener with empty body to make them recognized as progressive web application (PWA).
```
jakearchibald, line 2997: This comment should be expanded upon, but I can help with that.

docs/index.bs
```diff
@@ -2970,16 +2972,34 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+      1. return false.
```
jakearchibald, line 2999:
```suggestion
      1. Return false.
```

docs/index.bs
```diff
@@ -3041,7 +3061,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
           1. Set |registration| to the result of running <a>Match Service Worker Registration</a> given |storage key| and |request|'s [=request/url=].
           1. If |registration| is null or |registration|'s <a>active worker</a> is null, return null.
           1. If |request|'s [=request/destination=] is not {{RequestDestination/"report"}}, set |reservedClient|'s <a>active service worker</a> to |registration|'s <a>active worker</a>.
-          1. If |request| is a [=navigation request=], |registration|'s [=navigation preload enabled flag=] is set, |request|'s [=request/method=] is \`<code>GET</code>\`, and |registration|'s [=active worker=]'s <a>set of event types to handle</a> [=set/contains=] <code>fetch</code>, then:
+          1. If |request| is a [=navigation request=], |registration|'s [=navigation preload enabled flag=] is set, |request|'s [=request/method=] is \`<code>GET</code>\`, |registration|'s [=active worker=]'s <a>set of event types to handle</a> [=set/contains=] <code>fetch</code>, and the result of running [=Skippable Fetch Handler=] algorithm with [=active worker=] is false then:
```
jakearchibald, line 3064:
```suggestion
          1. If |request| is a [=navigation request=], |registration|'s [=navigation preload enabled flag=] is set, |request|'s [=request/method=] is \`<code>GET</code>\`, |registration|'s [=active worker=]'s <a>set of event types to handle</a> [=set/contains=] <code>fetch</code>, and the result of running [=Skippable Fetch Handler=] algorithm with |registration|'s [=active worker=] is false then:
```

docs/index.bs
```diff
@@ -3077,6 +3097,12 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       1. If the result of running the [=Should Skip Event=] algorithm with "fetch" and |activeWorker| is true, then:
           1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
           1. Return null.
+      1. If the result of running [=Skippable Fetch Handler=] algorithm with |activeWorker| is true, then:
```
jakearchibald, line 3100:
```suggestion
      1. If the result of running the [=Skippable Fetch Handler=] algorithm with |activeWorker| is true, then:
```

docs/index.bs
```diff
@@ -3077,6 +3097,12 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+          1. run the following steps [=in parallel=]:
```
jakearchibald, line 3101:
```suggestion
          1. [=In parallel=]:
```

docs/index.bs
```diff
@@ -3077,6 +3097,12 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+              1. If |activeWorker|'s state is "activating", wait for |activeWorker|'s state to become "activated".
```
jakearchibald, line 3102:
```suggestion
              1. If |activeWorker|'s [=service worker/state=] is "activating", then wait for |activeWorker|'s [=service worker/state=] to become "activated".
```

Make sure terms you use are referenced

docs/index.bs
```diff
@@ -191,6 +191,8 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     A [=/service worker=] has an associated <dfn>start status</dfn> which can be null or a [=Completion=]. It is initially null.
 
+    A [=/service worker=] has an associated <dfn>has had non-no-op fetch listeners flag</dfn>. It is initially true.
```
jakearchibald, line 194: Maybe rename this to reflect the uncertain nature?

```suggestion
    A [=/service worker=] has an associated <dfn>may have had non-empty fetch listener flag</dfn>. It is initially unset.
```

docs/index.bs
```diff
@@ -3132,6 +3158,17 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
   </section>
 
+  <section algorithm>
+    <h3 id="skippable-fetch-handler-algorithm"><dfn>Skippable Fetch Handler</dfn></h3>
+      : Input
+      :: |serviceWorker|, a [=/service worker=]
+      : Output
+      :: a boolean
+
+      1. If |serviceworker| has |has non-no-op fetch listeners flag|, and the value is false, then the user agent *may* return true.
```
jakearchibald, line 3168: What's the reason for "may" here, given that "No-Op Handler identification" already uses "may"?

jakearchibald (changes requested):

docs/index.bs

docs/index.bs

docs/index.bs

docs/index.bs

docs/index.bs
yoshisatoyanagisawa: Let me update the flag name.  The current flag name requires to have the default value because we should not think fetch handlers are no-op by default.

docs/index.bs

docs/index.bs
yoshisatoyanagisawa: I am not so confident to write references, but I tried to address this.  please take a look.

docs/index.bs

docs/index.bs
yoshisatoyanagisawa: I have renamed the flag without negation.  Please let me know what you think.

docs/index.bs
yoshisatoyanagisawa: I just tried to follow what "Should Skip Event" does, but it does not need to be *may*.  Let me omit.

docs/index.bs
yoshisatoyanagisawa: "event listeners" seems to be already defined.  Do you want to replace all of them with DOM standard's event listener?

annevk:
I suspect that event handlers might be tricky here as while an event handler can be considered empty, the event listener it ends up adding won't be (as it wraps the function). Perhaps this can be addressed by poking at HTML's event handler map. Gets a little hairy though.

Might be good to get some advice from @domenic and possibly some JS folks.

docs/index.bs
yoshisatoyanagisawa: To avoid the conflict of the "event listeners" name, I avoided to use it while referring callback.

jakearchibald:
@yoshisatoyanagisawa it's been a while since I worked on the service worker spec, and it's built up some errors over time as other specs have changed, and bikeshed has updated. I'm going to fix up all those errors to make it easier for you to make this change.

jakearchibald:
@ yoshisatoyanagisawa I've got the errors down to 14 (I need changes in other specs to land before I can make it 0). Hopefully the conflicts aren't too much trouble.

I'm going to take a closer look at this PR tomorrow.

yoshisatoyanagisawa:
@jakearchibald thank you.  I have resolved a merge conflict, and fixed bikeshed errors caused within my change.  Please take another look.

docs/index.bs
```diff
@@ -2973,16 +2983,44 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      : Input
+      :: |workerGlobalScope|, a [=service worker/global object=].
+      : Output
+      :: a boolean
+
+      1. If |workerGlobalScope|'s <a>set of event types to handle</a> does not [=set/contain=] <code>fetch</code>, then returns false.
+      1. set |eventListenerList| to [=map/value=] whose [=map/key=] is <code>fetch</code> in |workerGlobalScope|'s <a>set of event types to handle</a>.
+      1. [=list/For each=] |eventListener| of |eventListenerList|:
+          1. If |eventListener|'s [=callback=] is not null:
+              1. Set |callback| to an ECMAScript object coverted from |eventListener|'s [=callback=], which is [=idl-callback-interface=].
```
domenic, line 3012:
```suggestion
              1. Set |callback| to the result of [=converting|convert to an ECMAScript value=] |eventListener|'s [=callback=] to an ECMAScript value.
```

docs/index.bs
```diff
@@ -2973,16 +2983,44 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+              1. If |callback| is a [=function=], and [=function body=] is not empty (i.e. either [=statement=] or [=declaration=] exist), then return false.
```
domenic, line 3013:
```suggestion
              1. If [$IsCallable$](|callback|), and |callback|'s [=function body=] is not empty (i.e. either a [=statement=] or [=declaration=] exist), then return false.
```

docs/index.bs
```diff
@@ -2973,16 +2983,44 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+              1. If |callback| does not have <code>handleEvent</code> property, then return false.
+              1. If |callback|'s <code>handleEvent</code> property is not a [=function=], then return false.
+              1. If a [=function body=] of |callback|'s <code>handleEvent</code> property is not empty (i.e. either [=statement=] or [=declaration=] exist), return false.
```
domenic, line 3016:
```suggestion
              1. Return false.
```

domenic (commented):
Some suggested tweaks to be more precise, plus a suggestion that you bail out in the object case instead of potentially triggering side effects from the `handleEvent` getter.

docs/index.bs

docs/index.bs
yoshisatoyanagisawa: I followed the suggestion, and converted the statement to two statements.

docs/index.bs
yoshisatoyanagisawa: Considering https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/bindings/core/v8/js_event_listener.cc;l=14;drc=609b81c1b05359d844df8786169ba20f230152b4 behavior, it might be fair to say that this would also see handleEvent.  I tried to update this with  https://webidl.spec.whatwg.org/#call-a-user-objects-operation 10 upon your offline suggestion.

docs/index.bs
```diff
@@ -62,6 +64,8 @@ spec: ecma-262; urlPrefix: http://tc39.github.io/ecma262/
     type: abstract-op
         text: NormalCompletion; url: sec-normalcompletion
         text: ThrowCompletion; url: sec-throwcompletion
+        text: IsCallable; url: sec-iscallable
+        text: Get; url: sec-get-o-p
```
domenic, line 68: These should not be necessary these days. But, maybe they are; if removing them breaks then ignore my advice :)

docs/index.bs
```diff
@@ -2973,16 +2983,46 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+  <section algorithm>
+    <h3 id="empty-handler-identification-algorithm"><dfn>Empty Handler identification</dfn></h3>
+
+      : Input
+      :: |workerGlobalScope|, a [=service worker/global object=].
+      : Output
+      :: a boolean
+
+      1. If |workerGlobalScope|'s <a>set of event types to handle</a> does not [=set/contain=] <code>fetch</code>, then returns false.
+      1. set |eventListenerList| to [=map/value=] whose [=map/key=] is <code>fetch</code> in |workerGlobalScope|'s <a>set of event types to handle</a>.
```
domenic, line 3009: The inner loop below this line now looks great.

This step doesn't seem correct though. |eventListenerList| should be https://dom.spec.whatwg.org/#eventtarget-event-listener-list . And then you need to filter out the ones where https://dom.spec.whatwg.org/#event-listener-type is not "fetch".

docs/index.bs
yoshisatoyanagisawa: They seem to be needed. 
LINE ~3011: No 'abstract-op' refs found for 'IsCallable'.
[$IsCallable$]
LINE ~3012: No 'abstract-op' refs found for 'Get'.
[$Get$]
LINE ~3015: No 'abstract-op' refs found for 'IsCallable'.
[$IsCallable$]

docs/index.bs
yoshisatoyanagisawa: I misunderstood the mental model of "set of event types to handle".  I thought is is something like std::map, but it is actually set.  I have updated the explanation like so.

docs/index.bs
```diff
@@ -2973,16 +2984,48 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+    <h3 id="empty-handler-identification-algorithm"><dfn>Empty Handler identification</dfn></h3>
+
+      : Input
+      :: |workerGlobalScope|, a [=service worker/global object=].
+      : Output
+      :: a boolean
+
+      1. If |workerGlobalScope|'s <a>set of event types to handle</a> does not [=set/contain=] <code>fetch</code>, then returns false.
+      1. Let |eventListenerList| be an empty [=list=].
+      1. [=list/For each=] |eventListener| of |workerGlobalScope|'s <a>set of event types to handle</a>:
```
jakearchibald, line 3011: This isn't right. `<a>set of event types to handle</a>` is a set of strings, not listeners.

In cases like this, where I want to get something like "all event listener callbacks of a particular type", I try to think of existing things that do that already that I could take inspiration from. In this case, I'd look at `target.dispatchEvent`, since it has to run all of the callbacks of a particular type.

So, I'd start at https://dom.spec.whatwg.org/#dom-eventtarget-dispatchevent, and follow the algorithm, and see how they gather up the callbacks. We probably need to do a similar thing here.

Following the trail takes me to https://dom.spec.whatwg.org/#concept-event-listener-invoke, which shows me how to get the listeners for a particular event target.

Then it goes to https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke, which shows me how to filter by event type.

Then it goes to https://webidl.spec.whatwg.org/#call-a-user-objects-operation which shows me how to get the callback from the two different forms of listener.

docs/index.bs
```diff
@@ -3130,6 +3179,17 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
   </section>
 
+  <section algorithm>
+    <h3 id="skippable-fetch-handler-algorithm"><dfn>Skippable Fetch Handler</dfn></h3>
+      : Input
+      :: |serviceWorker|, a [=/service worker=]
+      : Output
+      :: a boolean
+
+      1. If |serviceWorker| has [=all fetch listeners are empty flag=] and set, then the user agent *may* return true.
```
jakearchibald, line 3189: I think the "may" condition should sit in "Empty Handler identification", since that's the part of the process that's optional.

jakearchibald (changes requested):

docs/index.bs
yoshisatoyanagisawa: Please feel free to reopen if this still matters.

docs/index.bs
yoshisatoyanagisawa: Please feel free to reopen if this still matters.

docs/index.bs
```diff
@@ -2973,16 +2984,48 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               1. For each |eventType| of |settingsObject|'s [=environment settings object/global object=]'s associated list of <a>event listeners</a>' event types:
                   1. [=set/Append=] |eventType| to |workerGlobalScope|'s associated [=ServiceWorkerGlobalScope/service worker=]'s <a>set of event types to handle</a>.
 
-                  Note: If the global object's associated list of event listeners does not have any event listener added at this moment, the service worker's set of event types to handle remains an empty set. The user agents are encouraged to show a warning that the event listeners must be added on the very first evaluation of the worker script.
+                  Note: If the global object's associated list of event listeners does not have any event listener added at this moment, the service worker's <a>set of event types to handle</a> remains an empty set.
```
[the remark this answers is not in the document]

docs/index.bs
```diff
@@ -2973,16 +2984,48 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
           1. Run the <a>responsible event loop</a> specified by |settingsObject| until it is destroyed.
           1. [=map/Clear=] |workerGlobalScope|'s [=map of active timers=].
       1. Wait for |serviceWorker| to be [=running=], or for |startFailed| to be true.
       1. If |startFailed| is true, then return *failure*.
       1. Return |serviceWorker|'s [=start status=].
   </section>
 
+  <section algorithm>
+    <h3 id="empty-handler-identification-algorithm"><dfn>Empty Handler identification</dfn></h3>
```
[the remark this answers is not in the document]
yoshisatoyanagisawa: added "fetch".

docs/index.bs
yoshisatoyanagisawa: Sure.  I have moved the part to "Empty Fetch Handler Identification" algorithm.

docs/index.bs
```diff
@@ -2973,16 +2984,48 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               1. For each |eventType| of |settingsObject|'s [=environment settings object/global object=]'s associated list of <a>event listeners</a>' event types:
                   1. [=set/Append=] |eventType| to |workerGlobalScope|'s associated [=ServiceWorkerGlobalScope/service worker=]'s <a>set of event types to handle</a>.
 
-                  Note: If the global object's associated list of event listeners does not have any event listener added at this moment, the service worker's set of event types to handle remains an empty set. The user agents are encouraged to show a warning that the event listeners must be added on the very first evaluation of the worker script.
+                  Note: If the global object's associated list of event listeners does not have any event listener added at this moment, the service worker's <a>set of event types to handle</a> remains an empty set.
 
               1. Set |script|'s <a>has ever been evaluated flag</a>.
+              1. If the [=Empty Handler identification=] algorithm with |workerGlobalScope| returns true, the [=all fetch listeners are empty flag=] is set.
```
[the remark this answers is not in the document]
yoshisatoyanagisawa: I think you are talking about [Approaches to deal with the handler updates after the initialization](https://github.com/yoshisatoyanagisawa/service-worker-skip-no-op-fetch-handler#approaches-to-deal-with-the-handler-updates-after-the-initialization) in the explainer.  This implements [Option 2](https://github.com/yoshisatoyanagisawa/service-worker-skip-no-op-fetch-handler#option-2-updates-are-checked-every-time-when-the-service-worker-starts).  However, this just set "all fetch listeners are empty flag", and never unset.  Upon the algorithm, once the flag is set, it would be kept and does not behave like Option 2 (i.e. the fetch handler function body is evaluated every time service worker starts).  Let me fix.

docs/index.bs
```diff
@@ -2973,16 +2984,48 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
 
+  <section algorithm>
+    <h3 id="empty-handler-identification-algorithm"><dfn>Empty Handler identification</dfn></h3>
+
+      : Input
+      :: |workerGlobalScope|, a [=service worker/global object=].
+      : Output
+      :: a boolean
+
+      1. If |workerGlobalScope|'s <a>set of event types to handle</a> does not [=set/contain=] <code>fetch</code>, then returns false.
```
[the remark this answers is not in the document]
yoshisatoyanagisawa: The behavior is intended, but people may feel the flag name does not match with the behavior.
I intend to say the fetch handler exists and all of them are empty, but it sounds long.

docs/index.bs
```diff
@@ -2973,16 +2984,49 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+          1. If |eventListener|'s [=callback=] is not null:
+              1. Set |callback| to the result of [=converting|convert to an ECMAScript value=] |eventListener|'s [=callback=] to an ECMAScript value.
+              1. If [$IsCallable$](|callback|) is false:
+                  1. Let |getResult| be [=Completion=]([$Get$](|callback|), <code>handleEvent</code>).
+                  1. If |getResult| is [=abrupt completion=], then return false.
+                  1. Set |callback| to |getResult|.
+              1. If [$IsCallable$](|callback|) is false, then return false.
+              1. If |callback|'s [=function body=] is not empty (i.e. either a [=statement=] or [=declaration=] exist), then return false.
+
+              Note: it detects something like `onfetch = () => {}`. Some sites have a fetch event listener with empty body to make them recognized as progressive web application (PWA).
```
annevk, line 3024: @domenic I think this still doesn't work for `onfetch` as HTML wraps that function in its own function when creating the callback, right? See https://github.com/WebKit/standards-positions/issues/129#issuecomment-1431008116.

domenic: Yes. We started discussing how to fix that (offline) but then got sidetracked into making the surrounding parts of the spec more precise.

To detech `onfetch = () => {}` I think the following would work, nested immediately under "If _eventListener_'s callback is not null":

> 1. If _eventListener_ is an `EventListener` of the type created by [activate an event handler](https://html.spec.whatwg.org/#activate-an-event-handler), then set |callback| to the result of [converting] _eventTarget_'s [event handler map](https://html.spec.whatwg.org/#event-handler-map)["`fetch`"]'s [value](https://html.spec.whatwg.org/#event-handler-value) to an ECMAScript value.
> 2. Otherwise, set |callback| to ...

(you could also factor out the [converting] into a separate step.)

docs/index.bs
yoshisatoyanagisawa: Hmm, I got stuck on finding who sets currentTarget (https://dom.spec.whatwg.org/#dom-event-currenttarget) in Dispatching Events (https://dom.spec.whatwg.org/#concept-event-dispatch).
If that is cleared, we can get event listeners from currentTarget like invoke's step 6 (https://dom.spec.whatwg.org/#concept-event-listener-invoke), and the current algorithm can be used.
I tired to update anyway.

docs/index.bs
yoshisatoyanagisawa: Can I ask where eventTarget in Step 1 come from?

yoshisatoyanagisawa: I am not fully understand the comment, but I tried to update the steps as suggested.
https://github.com/w3c/ServiceWorker/pull/1672/commits/27f0e006edf2af02c1c11a9421879aa11f467c51

yoshisatoyanagisawa:
Can I ask some more comments on how the empty fetch handler identification algorithm should be?  I am also happy to listen any other comments if there is.

docs/index.bs
```diff
@@ -2973,16 +2984,52 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+      1. If |workerGlobalScope|'s [=set of event types to handle=] does not [=set/contain=] <code>fetch</code>, then returns false.
+      1. Let |onFetchHandler| be |workerGlobalScope|'s <code>onfetch</code> attribute.
+      1. If |onFetchHandler|'s <a spec="html">listener</a> is null, then return false.
+      1. Let |eventListenerList| be an empty [=list=].
+      1. [=list/For each=] |eventListener| of [onFetchHandler|'s <a spec="html">listener</a>:
+          1. If |eventListener|'s <a spec="dom">type</a> is <code>fetch</code>, then [=list/append=] |eventListener| to |eventListenerList|.
+      1. [=list/For each=] |eventListener| of |eventListenerList|:
+          1. If |eventListener|'s [=callback=] is not null:
+              1. If |eventListener| is an EventListener of the type created by <a spec="html">activate an event handler</a>, then set |callback| to the result of [=converting|convert to an ECMAScript value=] |workerGlobalScope|'s <a spec="html">event handler map</a>["fetch"]'s value to an ECMAScript value.
```
jakearchibald, line 3018: @domenic ignoring the formatting issues, are you ok with this kind of hand wave? Alternative language could be: "Get the [event handler](https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers) with a [listener](https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-listener) of |eventListener|'s [=callback=], or null if none exists."

Then we can pull the value from that object.

docs/index.bs
```diff
@@ -3130,6 +3183,17 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+  <section algorithm>
+    <h3 id="skippable-fetch-handler-algorithm"><dfn>Skippable Fetch Handler</dfn></h3>
+      : Input
+      :: |serviceWorker|, a [=/service worker=]
+      : Output
+      :: a boolean
+
+      1. If |serviceWorker| has [=all fetch listeners are empty flag=] and set, then the user agent return true.
+      1. Return false.
+  </section>
```
jakearchibald, line 3195: I still don't think we need this. It's just returning an existing boolean value.

jakearchibald (changes requested):

docs/index.bs
```diff
@@ -2973,16 +2984,52 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
                   1. [=set/Append=] |eventType| to |workerGlobalScope|'s associated [=ServiceWorkerGlobalScope/service worker=]'s <a>set of event types to handle</a>.
 
-                  Note: If the global object's associated list of event listeners does not have any event listener added at this moment, the service worker's set of event types to handle remains an empty set. The user agents are encouraged to show a warning that the event listeners must be added on the very first evaluation of the worker script.
+                  Note: If the global object's associated list of event listeners does not have any event listener added at this moment, the service worker's [=set of event types to handle=] remains an empty set.
 
               1. Set |script|'s <a>has ever been evaluated flag</a>.
+              1. Unset the [=all fetch listeners are empty flag=].
+              1. If the [=Empty Fetch Handler Identification=] algorithm with |workerGlobalScope| returns true, the [=all fetch listeners are empty flag=] may be set.
```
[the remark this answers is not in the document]
domenic: I like your suggested wording!

docs/index.bs
domenic: I suggested this type of handwave, because I didn't realize we could do better. But I think you're right. I think it should be something like this:

- Outside the loop, let |eventHandler| be |workerGlobalScope|'s <a spec="html">event handler map</a>["fetch"].

- Inside the loop, check if |eventListener| equals |eventHandler|'s [listener](https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-listener). If so, set |callback| to the result of converting |eventHandler|'s [value](https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-value) to an ECMAScript value.

docs/index.bs
```diff
@@ -2973,16 +2984,52 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
           1. Run the <a>responsible event loop</a> specified by |settingsObject| until it is destroyed.
           1. [=map/Clear=] |workerGlobalScope|'s [=map of active timers=].
       1. Wait for |serviceWorker| to be [=running=], or for |startFailed| to be true.
       1. If |startFailed| is true, then return *failure*.
       1. Return |serviceWorker|'s [=start status=].
   </section>
 
+  <section algorithm>
+    <h3 id="all-fetch-handlers-are-empty-algorithm"><dfn>All Fetch Handlers Are Empty</dfn></h3>
```
jakearchibald, line 3001: Ugh, my bad, we should be using "listeners" rather than "handlers". I should have caught this in the last review, sorry.

docs/index.bs
```diff
@@ -2973,16 +2985,51 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
 
+  <section algorithm>
+    <h3 id="all-fetch-listeners-are-empty-algorithm"><dfn>All Fetch Listeners Are Empty</dfn></h3>
+
+      : Input
+      :: |workerGlobalScope|, a [=service worker/global object=].
+      : Output
+      :: a boolean
+
+      1. If |workerGlobalScope|'s [=set of event types to handle=] does not [=set/contain=] <code>fetch</code>, then returns true.
```
jakearchibald, line 3009:
```suggestion
      1. If |workerGlobalScope|'s [=set of event types to handle=] does not [=set/contain=] <code>fetch</code>, then return true.
```

docs/index.bs
```diff
@@ -2973,16 +2985,51 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      1. Let |eventHandler| be |workerGlobalScope|'s <a spec="html">event handler map</a>["fetch"]'s value.
```
jakearchibald, line 3010:
```suggestion
      1. Let |eventHandler| be |workerGlobalScope|'s <a spec="html">event handler map</a>["onfetch"]'s value.
```

As per https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-name, handler names always start "`on`".

docs/index.bs
```diff
@@ -2973,16 +2985,51 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      1. Let |eventListenerList| be an empty [=list=].
+      1. [=list/For each=] |eventListener| of |workerGlobalScope|'s [=event listener list=]:
+          1. If |eventListener|'s <a spec="dom">type</a> is <code>fetch</code>, then [=list/append=] |eventListener| to |eventListenerList|.
```
jakearchibald, line 3013:
```suggestion
          1. If |eventListener|'s <a spec="dom">type</a> is "<code>fetch</code>", then [=list/append=] |eventListener| to |eventListenerList|.
```

docs/index.bs
```diff
@@ -2973,16 +2985,51 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      1. [=list/For each=] |eventListener| of |eventListenerList|:
+          1. If |eventListener|'s [=callback=] is not null:
```
jakearchibald, line 3015:
```suggestion
          1. If |eventListener|'s [=event listener/callback=] is null, then [=continue=].
```

Then un-indent the subsequent lines. This is mostly personal preference. I prefer early-return/continue rather than nesting everything in an `if`.

I swapped `[=callback=]` for `[=event listener/callback=]`. It isn't an issue now, but it feels like this could become ambiguous in future.

docs/index.bs
```diff
@@ -2973,16 +2985,51 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+              1. If |eventListener| equals |eventHandler|'s <a spec="html">listener</a>, then set |callback| to the result of [=converting|convert to an ECMAScript value=] |eventHandler|'s value to an ECMAScript value.
```
jakearchibald, line 3016: "|eventHandler|" may be null here.

Also, "value" in "|eventHandler|'s value" needs to be referenced properly.

docs/index.bs
```diff
@@ -2973,16 +2985,51 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+              1. Otherwise, set |callback| to the result of [=converting|convert to an ECMAScript value=] |eventListener|'s [=callback=] to an ECMAScript value.
+              1. If [$IsCallable$](|callback|) is false:
+                  1. Let |getResult| be [=Completion=]([$Get$](|callback|), <code>handleEvent</code>).
+                  1. If |getResult| is [=abrupt completion=], then return false.
+                  1. Set |callback| to |getResult|.
+              1. If [$IsCallable$](|callback|) is false, then return false.
+              1. If |callback|'s [=function body=] is not empty (i.e. either a [=statement=] or [=declaration=] exist), then return false.
+
+              Note: it detects something like `onfetch = () => {}`. Some sites have a fetch event listener with empty body to make them recognized as progressive web application (PWA).
```
jakearchibald, line 3025:
```suggestion
              Note: This detects "<code>fetch</code>" listeners like `() => {}`. Some sites have a fetch event listener with empty body to make them recognized by Chromium as a progressive web application (PWA).
```

docs/index.bs
```diff
@@ -2973,16 +2985,51 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+      1. Return true.
+
+      Note: the user agents are encouraged to show warnings that the event listeners are no-op, and their executions can be skipped.
```
jakearchibald, line 3029:
```suggestion
      Note: User agents are encouraged to show a warning indicating that empty "<code>fetch</code>" listeners are unnecessary, and may have a negative performance impact.
```

jakearchibald (changes requested):
This is really close to ready!

The PR uses `<a spec="">…</a>` links to link to concepts that haven't been exported by other specs. Instead, PRs should be made against those specs to export those terms. Example: https://github.com/whatwg/html/pull/8924

yoshisatoyanagisawa:
I hope to touch other specs for exporting terms after this PR.

yoshisatoyanagisawa:
I believe I have addressed all existing comments.  Are there anything to change?

docs/index.bs
```diff
@@ -2973,16 +2985,51 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      :: a boolean
+
+      1. If |workerGlobalScope|'s [=set of event types to handle=] does not [=set/contain=] <code>fetch</code>, then return true.
+      1. Let |eventHandler| be |workerGlobalScope|'s <a spec="html">event handler map</a>["onfetch"]'s value.
+      1. Let |eventListenerList| be an empty [=list=].
+      1. [=list/For each=] |eventListener| of |workerGlobalScope|'s [=event listener list=]:
+          1. If |eventListener|'s <a spec="dom">type</a> is "<code>fetch</code>", then [=list/append=] |eventListener| to |eventListenerList|.
+      1. [=list/For each=] |eventListener| of |eventListenerList|:
+          1. If |eventListener|'s [=event listener/callback=] is null, then [=continue=].
+          1. If |eventHandler| is not null, and |eventListener| equals |eventHandler|'s <a spec="html">listener</a>, then set |callback| to the result of [=converting|convert to an ECMAScript value=] |eventHandler|'s <a spec="html">listener</a> value to an ECMAScript value.
```
jakearchibald, line 3016: Lines like this are still pointing to unexported terms in other specs. Can you get these exported in other specs first? I don't want to land the spec linking to unexported terms.

docs/index.bs
```diff
@@ -3075,6 +3122,12 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       1. If the result of running the [=Should Skip Event=] algorithm with "fetch" and |activeWorker| is true, then:
           1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
           1. Return null.
+      1. If |activeWorker|'s [=all fetch listeners are empty flag=] is set:
+          1. [=In parallel=]:
+              1. If |activeWorker|'s [=service worker/state=] is "activating", then wait for |activeWorker|'s [=service worker/state=] to become "activated".
+              1. Run the [=Run Service Worker=] algorithm with |activeWorker|.
+              1. If |shouldSoftUpdate| is true, then run the [=Soft Update algorithm=] with |registration|.
```
jakearchibald, line 3129:
```suggestion
              1. If |shouldSoftUpdate| is true, then run the [=Soft Update=] algorithm with |registration|.
```

The link text here was wrong (and was causing a bikeshed error.

jakearchibald (changes requested):
One small error, and the exporting from other specs left to do.

yoshisatoyanagisawa:
Will you take another look at the PR?  I believe I have addressed your comments.
I suppose an error like "LINE ~3010: No 'dfn' refs found for 'event handler map' that are marked for export." is because the time lag between the PR is merged and available, but I am not sure.  I hope I can also have advice on that.

Thank you.

jakearchibald:
Sighhh, something git-related went wrong here. Let me repair…

jakearchibald:
Continuing in https://github.com/w3c/ServiceWorker/pull/1674

PR #1676, opened 2023-04-03
Avoid executing the code in the All Fetch Listeners Are Empty algorithm.

yoshisatoyanagisawa:
This is for stop calling Get for detecting an empty fetch listener.  Benefit of the change is that it won't execute the Javascript, and drawback is that it just ignores an empty fetch listener defined as a handleEvent property.  Since it is rare to configure the fetch event listener with the handleEvent property, not detecting empty function body for it should be negligible coverage reduction.

+@domenic

***
<a href="https://pr-preview.s3.amazonaws.com/yoshisatoyanagisawa/ServiceWorker/pull/1676.html" title="Last updated on Apr 3, 2023, 10:09 AM UTC (2ea42ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1676/4b38437...yoshisatoyanagisawa:2ea42ac.html" title="Last updated on Apr 3, 2023, 10:09 AM UTC (2ea42ac)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3009,7 +3009,6 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
           1. Let |callback| be null.
           1. If |eventHandler| is not null and |eventListenerCallback| equals |eventHandler|'s [=event handler/listener=]'s [=event listener/callback=], then set |callback| to the result of [=convert to an ECMAScript value|converting to an ECMAScript value=] |eventHandler|'s [=event handler/value=].
           1. Otherwise, set |callback| to the result of [=convert to an ECMAScript value|converting to an ECMAScript value=] |eventListenerCallback|.
-          1. If [$IsCallable$](|callback|) is false, then return false.
           1. If [$IsCallable$](|callback|) is false, or |callback|'s [=function body=] is not empty (i.e. either a [=statement=] or [=declaration=] exist), then return false.
 
           Note: This detects "<code>fetch</code>" listeners like `() => {}`. Some sites have a fetch event listener with empty body to make them recognized by Chromium as a progressive web application (PWA).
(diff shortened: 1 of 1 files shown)


jakearchibald:
@annevk @youennf @asutherland This is a change to the "optimise for empty fetch listeners" feature (https://github.com/w3c/ServiceWorker/issues/1671).

Previously, we were looking at both regular listeners, and `{ handleEvent }` listeners. However, running getters during the check adds a bunch of complexity. With this change, the optimisation doesn't apply if there's a non-callable listener object (as in, `{ handleEvent }`).

yoshisatoyanagisawa:
> This looks good, but can you add a note that makes it clear that we're deliberately returning false for handleEvent?

Sure.  I have added the note.
I am not confident that the explanation makes sense to you.

yoshisatoyanagisawa:
> false

docs/index.bs
```diff
@@ -3009,7 +3009,11 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
           1. Let |callback| be null.
           1. If |eventHandler| is not null and |eventListenerCallback| equals |eventHandler|'s [=event handler/listener=]'s [=event listener/callback=], then set |callback| to the result of [=convert to an ECMAScript value|converting to an ECMAScript value=] |eventHandler|'s [=event handler/value=].
           1. Otherwise, set |callback| to the result of [=convert to an ECMAScript value|converting to an ECMAScript value=] |eventListenerCallback|.
-          1. If [$IsCallable$](|callback|) is false, or |callback|'s [=function body=] is not empty (i.e. either a [=statement=] or [=declaration=] exist), then return false.
+          1. If [$IsCallable$](|callback|) is false, then return false.
+
+          Note: This deliberately ignores the result of [=Completion=]([$Get$](|callback|), <code>handleEvent</code>) because running getters during the check adds a bunch of complexity.
```
jakearchibald, line 3014:
```suggestion
              Note: [=Event listener/Callback=] objects that use {{handleEvent}} are assumed to be non-empty. This avoids calling the {{handleEvent}} getters, which could modify the event listeners during this check.
```

yoshisatoyanagisawa:
Hi, are there any updates on this?
This narrows down the area, which is treated as an empty fetch handler to exclude something rarely used for the PWA purpose.

jakearchibald (approved):
LGTM

PR #1686, opened 2023-07-10
Add a static routing API support.

yoshisatoyanagisawa:
A specification update to support the static routing API proposed in https://github.com/WICG/proposals/issues/102

***
<a href="https://pr-preview.s3.amazonaws.com/yoshisatoyanagisawa/ServiceWorker/pull/1686.html" title="Last updated on Aug 15, 2023, 9:15 AM UTC (cbd47b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1686/fa114cd...yoshisatoyanagisawa:cbd47b4.html" title="Last updated on Aug 15, 2023, 9:15 AM UTC (cbd47b4)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -2945,8 +2945,6 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               :: Return |realmExecutionContext|.
               : The [=environment settings object/module map=]
               :: Return |workerGlobalScope|'s [=WorkerGlobalScope/module map=].
-              : The <a>API URL character encoding</a>
-              :: Return [=UTF-8=].
               : The <a>API base URL</a>
               :: Return |serviceWorker|'s [=service worker/script url=].
               : The [=environment settings object/origin=]


docs/index.bs
```diff
@@ -1545,6 +1552,65 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+      <section>
+        <h5 id="verify-router-rule-algorithm"><dfn>VerifyRouterRule</dfn></h5>
+          : Input
+          :: |rule|, RouterRule
+          : Output
+          :: a boolean
+
+          1. If |rule|'s |condition| is empty, and abort these steps.
+          1. If |rule|'s |condition|'s |urlPattern| is empty, return false..
```
domenic, line 1601: What do you mean by "is empty"?

If you want to require that the IDL member be present, use `required` in the IDL and delete this step.

If you want to require that the value is not the empty string, then check "is the empty string" or `'s [=string/length=] is 0`.

docs/index.bs
```diff
@@ -1545,6 +1552,65 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+          1. If |urlPattern| contains a regular expression, return false.
```
domenic, line 1602: We need to define this more rigorously somehow. How does it work in the code?

domenic, line 1602: The |urlPattern| variable is not declared anywhere.

docs/index.bs
```diff
@@ -1545,6 +1552,65 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+             Note: Since running a user-defined regular expression has a security concern, it is prohibited.
+
+          1. If |urlPattern| cannot be [=urlpattern/parsed=], return false.
```
domenic, line 1606: This is not appropriately rigorous. Given a string (e.g. `|rule|["{{RouterRule/condition}}"]["{{RouterCondition/urlPattern}}"]`) we can then feed it to the [parse a pattern string](https://wicg.github.io/urlpattern/#parse-a-pattern-string) algorithm, or maybe the [parse a constructor string](https://wicg.github.io/urlpattern/#parse-a-constructor-string) algorithm. Which one do you want?

You then will get a return value from that. What kind of return value do you consider "cannot be parsed"?

Finally, you probably will want to store the result of parsing, so that you can execute the pattern later.

docs/index.bs
```diff
@@ -1545,6 +1552,65 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+          1. If |rule|'s |source| is empty, return false.
+          1. If |rule|'s |source| is not "network", return false.
```
domenic, line 1608: Both of these conditions are impossible because of the Web IDL layer's `required` and `enum` features.

docs/index.bs
```diff
@@ -1545,6 +1552,65 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+        1. Let |routerRules| be a list of {{RouterRule}} dictionaries.
+        1. If |rules| is a {{RouterRule}} dictionary:
+            1. Run [=VerifyRouterRule=] algorithm with |rules|.
+            1. Append |rules| to |routerRules|.
+        1. If |rules| is a sequence of {{RouterRule}} dictionaries.
+            1.  for each |rule| in |rules|:
+                1. If running [=VerifyRouterRule=] algorithm with |rule| returns false, <a>throw</a> a <code>TypeError</code>.
+                1. Append |rule| to |routerRules|.
+        1. Set |routerRules| to [=/service worker=]'s [=static router rules object=].
```
domenic, line 1591: I think this is backward.

docs/index.bs
```diff
@@ -3177,6 +3245,19 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+  <section algorithm>
+    <h3 id="get-router-source-algorithm"><dfn>GetRouterSource</dfn></h3>
+      : Input
+      :: |rules|, a list of {{RouterRule}}
+      :: |request|, a [=/request=]
+      : Output
+      :: {{RouterSourceEnum}} or null
+
+      1. For each |rule| in |rules|:
+          1. If |request|'s [=request/URL=] [=urlpattern/matches=] |rule|'s {{urlPattern}}, then returns |rule|'s {{RouterRule/source}}.
```
domenic, line 3257: rule's urlPattern (better written as `|rule|["{{RouterRule/condition}}"]["{{RouterCondition/urlPattern}}"]`) is a string. https://wicg.github.io/urlpattern/#match takes a `URLPattern` object, not a string. This is related to my issue above about needing to store the parsing result somewhere.

(Probably, you will need a separate hierarchy of parsed router rule [structs](https://infra.spec.whatwg.org/#structs), separate from the raw `RouterRule` dictionaries.)

domenic (changes requested):

docs/index.bs
```diff
@@ -1545,6 +1552,65 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+        1. If |rules| is a {{RouterRule}} dictionary:
+            1. Run [=VerifyRouterRule=] algorithm with |rules|.
+            1. Append |rules| to |routerRules|.
+        1. If |rules| is a sequence of {{RouterRule}} dictionaries.
+            1.  for each |rule| in |rules|:
+                1. If running [=VerifyRouterRule=] algorithm with |rule| returns false, <a>throw</a> a <code>TypeError</code>.
+                1. Append |rule| to |routerRules|.
+        1. Set |routerRules| to [=/service worker=]'s [=static router rules object=].
+
+      <section>
```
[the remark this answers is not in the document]
yoshisatoyanagisawa: Moved.

docs/index.bs
```diff
@@ -1545,6 +1552,65 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+        <h5 id="verify-router-rule-algorithm"><dfn>VerifyRouterRule</dfn></h5>
+          : Input
+          :: |rule|, RouterRule
```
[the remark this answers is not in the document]

docs/index.bs
```diff
@@ -1545,6 +1552,65 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+          : Output
+          :: a boolean
+
+          1. If |rule|'s |condition| is empty, and abort these steps.
```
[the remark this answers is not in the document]
yoshisatoyanagisawa: I have removed this line.

docs/index.bs
yoshisatoyanagisawa: Considering the future expansion of the spec, let me leave this optional now and say "is the empty string" instead.

docs/index.bs
yoshisatoyanagisawa: It is actually like: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/service_worker/service_worker_router_type_converter.cc;l=41

I have updated the document to align with what the code actually do.

yoshisatoyanagisawa: I have written the declaration.

docs/index.bs
yoshisatoyanagisawa: fixed.

docs/index.bs
yoshisatoyanagisawa: It might be the "parse a constructor string" algorithm in the future, but it should currently be the "parse a pattern string" algorithm, and only supports pathname.  I clarified that.
I clarified what "cannot be parsed" mean here.  TypeError by parser is "cannot be parsed".
As URLPattern spec do, let me make urlPattern has an associated component.  I am not sure what is an appropriate way to write it down.

docs/index.bs
yoshisatoyanagisawa: sure.
removed.

docs/index.bs
yoshisatoyanagisawa: I tried to clarify that part.  I made RouterCondition/urlPattern to have an associated component.

yoshisatoyanagisawa:
I have updated the spec upon comments.
Please take another look at it.

docs/index.bs
```diff
@@ -1545,6 +1554,49 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      };
+
+      dictionary RouterCondition {
+        USVString urlPattern;
+      };
+
+      enum RouterSourceEnum { "network" };
+    </pre>
+
+    Each {{RouterCondition/urlPattern}} object has an associated <dfn>hostname component</dfn>, a [=urlpattern/component=], which is initially unset.
```
domenic, line 1578: This doesn't really work. You can't associate a component with a string inside a dictionary.

I think the way to do this is to create a parallel hierarchy of [structs](https://infra.spec.whatwg.org/#structs) mirroring your dictionaries, that represent the parsed result. E.g. a "router rule" struct with:

- source (a `RouterSourceEnum`)
- condition (a router condition struct)

and then a "router condition" struct with

- urlPattern, a component

Then the static router rules object should be a list of parsed "router rules", not a list of `RouterRule`s.

docs/index.bs
```diff
@@ -1545,6 +1554,49 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+    <section>
+      <h4 id="register-router-method">{{InstallEvent/registerRouter(rules)|event.registerRouter(rules)}}</h4>
+
+      {{InstallEvent/registerRouter(rules)}} registers this service worker the rules to offload simple tasks that the fetch handler does.
+
+      <dfn method for="InstallEvent"><code>registerRouter(|rules|)</code></dfn> method *must* run these steps:
+
+        1. Let |routerRules| be a list of {{RouterRule}} dictionaries.
+        1. If |rules| is a {{RouterRule}} dictionary:
```
domenic, line 1588: It seems simpler to have one step be

1. If |rules| is a {{RouterRule}} dictionary, set |rules| to « |rules| ».

(this is using the notation from https://infra.spec.whatwg.org/#lists).

docs/index.bs
```diff
@@ -3177,6 +3231,41 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+      : Input
+      :: |rule|, a {{RouterRule}}
+      : Output
+      :: a boolean
+
+      1. If |rule|["{{RouterRule/condition}}"]["{{RouterCondition/urlPattern}}"] is the empty string, return false.
+      1. Let |urlPattern| be a result of [=urlpattern/parsing=] |rule|["{{RouterRule/condition}}"]["{{RouterCondition/urlPattern}}"]. If a <code>TypeError</code> is thrown, return false.
+      1. [=list/For each=] |part| of |urlPattern|:
+          1. If |part|'s [=urlpattern-part/type=] is "regexp", return false.
```
domenic, line 3245:
```suggestion
          1. If |part|'s [=urlpattern-part/type=] is "<code>regexp</code>", return false.
```

docs/index.bs
```diff
@@ -3177,6 +3231,41 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+              Note: Since running a user-defined regular expression has a security concern, it is prohibited.
+
+      1. Set |urlPattern| to |rule|["{{RouterRule/condition}}"]["{{RouterCondition/urlPattern}}"]'s associated [=hostname component=].
```
domenic, line 3249: You need to get the hostname component from somewhere. Right now you have a list of parts (|urlPattern|) and |rule|["{{RouterRule/condition}}"]["{{RouterCondition/urlPattern}}"], which is just a string. I don't know how to get a hostname component from either of those.

docs/index.bs
```diff
@@ -3177,6 +3231,41 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      1. Return true.
+  </section>
```
domenic, line 3251: Probably we should return false if the pattern has any nontrivial components besides hostname, right? Otherwise people might try to use patterns like `/:foo` and not realize that will do nothing.

(I don't know how to do that precisely since, per the above, I don't know how to get from a list of parts and a string to the relevant components.)

docs/index.bs

docs/index.bs

docs/index.bs
yoshisatoyanagisawa: I tried to follow what https://wicg.github.io/urlpattern/#urlpattern-pathname-component does, but I am doing that well or not.
The components should be an internal structure and it should not be directly accessed by web facing APIs.
Are there any examples having a hidden data structure?

docs/index.bs
yoshisatoyanagisawa: The order of variables looks wrong.  Let me fix it.

docs/index.bs
yoshisatoyanagisawa: If it is outside of acceptable patterns, I believe TypeError would be raised by https://wicg.github.io/urlpattern/#parse-a-pattern-string.  If not, since there should be components, and it might be fine to use it for pattern matching.

By the way, if you are talking about supporting URLPattern "field"? other than the pathname, I recently lean on using URLPatternInit instead of USVString.  I mean that if developers want to do something complicated, they should use URLPatternInit.  USVString is only usable for pathnames.  What do you think?

yoshisatoyanagisawa: Upon your suggestion offline, I have updated the specification to support a full set of URLPattern instead of only pathname.  I suppose the comment became obsolete now.

docs/index.bs
yoshisatoyanagisawa: Anyway, now it is URLInput inside a directory, and the comment get outdated.  Please feel free to leave a new comment or reopen.

yoshisatoyanagisawa:
This has been accidentally removed.  I decided to work on the spec update in my personal repository, and will eventually send a PR after work inside the repository finishes.

https://github.com/yoshisatoyanagisawa/ServiceWorker/pull/1 will be the initial step.

PR #1711, opened 2024-03-19
Implement the "not" condition support

yoshisatoyanagisawa:
This is a specification update on supporting the "not" condition, which is also written in
https://github.com/WICG/service-worker-static-routing-api/blob/main/final-form.md.

The "not" condition is used for reversing the condition inside for ease of writing a matching rule for a reversed condition.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1578,6 +1578,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         RunningStatus runningStatus;
 
         sequence&lt;RouterCondition&gt; _or;
+        RouterCondition _not;
       };
 
       typedef (RouterSourceDict or RouterSourceEnum) RouterSource;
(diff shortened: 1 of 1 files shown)


yoshisatoyanagisawa:
@domenic @sisidovski Will you take a look at the spec change?

docs/index.bs
```diff
@@ -1578,6 +1578,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         RunningStatus runningStatus;
 
         sequence&lt;RouterCondition&gt; _or;
+        RouterCondition _not;
```
domenic, line 1581: I think this one does not need escaping, so just `not` is OK?

domenic (approved):
LGTM with one fix

docs/index.bs

yoshisatoyanagisawa (commented):
Thanks for the review.

yoshisatoyanagisawa:
@mkruisselbrink Will you review this?

sisidovski (approved):
LGTM

yoshisatoyanagisawa:
@mkruisselbrink Do you have any concerns on this PR?  Please feel free to point them out.

mkruisselbrink (approved):
lgtm, sorry for the delay in reviewing

Issue #1746, opened 2025-01-22
What should happen in case of large number of added routes?


youennf:
If the service worker is adding too many routes, two things may happen:
1. testing the routes will take a long time, delaying loads
2. storing the routes may fill up the user's storage

I guess that setting a limit on the number of routes might be nice here.
For 1, maybe a TypeError or InvalidStateError be good.
For 2, maybe a QuotaError would be good.

Looking at Chrome's implementation, https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/common/service_worker/service_worker_router_rule.h;l=33;drc=27d34700b83f381c62e3a348de2e6dfdc08364b8;bpv=1;bpt=1, it seems that there is a max number of 256.

Chrome is also imposing a max depth for conditions of 10, which is aligned with the goal to limit the time to test the routes.

The spec should probably give guidelines for implementors, if not adding some checks in the algorithms.

sisidovski:
We've discussed this topic in https://github.com/WICG/service-worker-static-routing-api/issues/5 and not concluded yet. I agree that we should mention the length/depth limit in the spec. Can we just say specific numbers in the spec? Or should we avoid mentioning specific numbers and give some flexibilities to implementers?

Yes, Chrome sets those numbers as length/depth limit, but I'm a bit unsure if those numbers are really reasonable.

youennf:
> Or should we avoid mentioning specific numbers and give some flexibilities to implementers?

That seems reasonable to me.

yoshisatoyanagisawa:
Note that https://github.com/w3c/ServiceWorker/pull/1714 is trying to cover the case.  It suggests existence of a rule size limit and a depth limit, but it does not tell details.
In the review @mkruisselbrink suggested to have a lower limit, and upper limit is up to the browser vendors.

There was WPTs for checking rule size and depth in the public WPT repo, but I moved it to the internal because I believe the limit is Chromium specific.
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/wpt_internal/service-worker/static-router/static-router-invalid-depth-width.https.html

Maybe no WPTs for ensuring the lower limit yet.

asutherland:
Specifying and testing a lower limit seems like a nice approach since there is indeed a webcompat concern.

domenic:
[...]

PR #1714, opened 2024-04-30
Editorial: Add notice on the _or condition depth.

yoshisatoyanagisawa:
In https://github.com/WICG/service-worker-static-routing-api/issues/5 and https://github.com/WICG/service-worker-static-routing-api/issues/6, it was suggested to limit the number of conditions in the or condition syntax, and depth of nested and/or conditions.

This PR adds notice on such limitations.

***
<a href="https://pr-preview.s3.amazonaws.com/yoshisatoyanagisawa/ServiceWorker/pull/1714.html" title="Last updated on Apr 30, 2024, 6:04 AM UTC (ab80a3c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1714/8d64a5c...yoshisatoyanagisawa:ab80a3c.html" title="Last updated on Apr 30, 2024, 6:04 AM UTC (ab80a3c)">Diff</a>

PR #1752, opened 2025-01-30
Add limits to the number of router rule registration

sisidovski:
In https://github.com/w3c/ServiceWorker/issues/1746, we discussed if the spec should give guidelines about the limit of added routes, and how the user agent should behave if the added routes exceeds the limit.

We have an agreement that we're willing to pick some specific limits. This PR proposes 1024 as the total number of router conditions, and 10 as the max depth of router condition nesting level. If it exceeds those limits, the user agent throws a TypeError.

For the counting conditions part, let me clarify how we calculate the total count as this a bit tricky.

Assuming that router conditions can be nested by using `or` and `not` syntaxes. We count every sub router conditions inside the parent `or` or `not` condition, including the parent condition itself. 

Here are some examples.

```js
self.addEventListener('install', event => {
  // The condition count of this addRoutes() is 1.
  //
  // The max nesting level is 1.
  event.addRoutes({
    condition: {
      urlPattern: { pathname: "/form/*" },
    },
    source: "network"
  });

  // The condition count of this addRoutes() is 1.
  // Even though there are multiple fields in one router condition, we treat is as 1.
  //
  // The max nesting level is 1.
  event.addRoutes({
    condition: {
      urlPattern: { pathname: "/form/*" },
      requestMethod: "post",
      requestMode: "navigation",
      runningStatus: "running"
    },
    source: "network"
  });

  // The condition count of this addRoutes() is 2.
  //
  // The max nesting level is 1.
  event.addRoutes([
    {
      condition: { requestMethod: "post" },
      source: "network"
    },
    {
      condition: { requestMethod: "get" },
      source: "network"
    }
  ]);

  // The condition count of this addRoutes() is 4.
  // `or` condition itself is inluded in the total count.
  //
  // The max nesting level is 2.
  event.addRoutes([
    {
      condition: {
        or: [
          { urlPattern: { pathname: "/form/*" } },
          { requestMethod: "post" },
          { requestMode: "navigate" },
        ]
      },
      source: "network"
    }
  ]);

  // The condition count of this addRoutes() is 7.
  // 3 `or` + 1 `not` + 1 `requestMethod` + 1 `requestMode` + 1 `runningStatus` = 7
  //
  // The max nesting level is 5.
  event.addRoutes([
    {
      condition: {
        or: [
          {
            or: [{
              or: [
                { requestMethod: 'get' },
                {
                  not: {
                    requestMode: 'navigate'
                  }
                }
              ]
            }]
          },
          { runningStatus: 'running' }
        ]
      },
      source: "network"
    }
  ]);

  // In total, 15 router conditions are registered in the install event.
  // Max nesting level is 5.
});
```

https://github.com/w3c/ServiceWorker/pull/1714 will be migrated into this PR.

***
<a href="https://pr-preview.s3.amazonaws.com/sisidovski/ServiceWorker/pull/1752.html" title="Last updated on Feb 6, 2025, 5:43 AM UTC (645a400)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1752/b6ac20c...sisidovski:645a400.html" title="Last updated on Feb 6, 2025, 5:43 AM UTC (645a400)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -212,6 +212,10 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     A [=/service worker=] has an associated <dfn>list of router rules</dfn> (a [=list=] of {{RouterRule}}s). It is initially an empty [=list=].
 
+    A [=/service worker=] has an associated <dfn>router rule count</dfn> (the number of registered {{RouterRule}}s). It is initially set to zero.
+
+    Note: [=router rule count=] is defined as the limit of the total number of registered router rules. This prevents the user agent from spending much resources by evaluating too many router rules. The limit is 1024.
+
     A [=/service worker=] is said to be <dfn>running</dfn> if its [=event loop=] is running.
 
     <section>
@@ -3401,9 +3405,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               Note: For ease of understanding the router rule, the "or" condition is mutually exclusive with other conditions.
 
           1. Let |orConditions| be |condition|["{{RouterCondition/_or}}"].
-
+         
               Note: To limit the resource usage and a condition evaluation time, |orConditions|'s [=list/size=] can be limited.
-
+ 
           1. For each |orCondition| of |orConditions|:
               1. If running the [=Verify Router Condition=] algorithm with |orCondition| and |serviceWorker| returns false, return false.
           1. Set |hasCondition| to true.
@@ -3414,6 +3418,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
           1. If running the [=Verify Router Condition=] algorithm with |condition|["{{RouterCondition/not}}"] and |serviceWorker| returns false, return false.
           1. Set |hasCondition| to true.
+      1. If |hasCondition| is true, then:
+          1. Increament |serviceWorker|'s [=service worker/router rule count=] by one.
+          1. If |serviceWorker|'s [=service worker/router rule count=] exceeds 1024, which is the limit of registered router rules, return false.
       1. Return |hasCondition|.
   </section>
 
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1600,6 +1600,8 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       };
     </pre>
 
+    Note: {{RouterCondition/_or}} and {{RouterCondition/not}} might have the other {{RouterCondition/_or}} or {{RouterCondition/not}} inside. To avoid spending much resources by the nested condition or performance penalty on evaluation, depth of such nested conditions can be limited.
+
     <section>
       <h4 id="register-router-method">{{InstallEvent/addRoutes(rules)|event.addRoutes(rules)}}</h4>
 
@@ -3399,6 +3401,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               Note: For ease of understanding the router rule, the "or" condition is mutually exclusive with other conditions.
 
           1. Let |orConditions| be |condition|["{{RouterCondition/_or}}"].
+
+              Note: To limit the resource usage and a condition evaluation time, |orConditions|'s [=list/size=] can be limited.
+
           1. For each |orCondition| of |orConditions|:
               1. If running the [=Verify Router Condition=] algorithm with |orCondition| and |serviceWorker| returns false, return false.
           1. Set |hasCondition| to true.
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3494,7 +3494,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       1. [=list/For each=] |rule| of |routerRules|:
           1. Set |result| to be the result of running [=Count Router Inner Conditions=] with |rule|["{{RouterRule/condition}}"], |result|, and 10.
           1. If |result|'s [=count router condition result/quota exceeded=] is true, return false.
-      1. return true.
+      1. Return true.
   </section> 
 
   <section algorithm>
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -212,10 +212,6 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     A [=/service worker=] has an associated <dfn>list of router rules</dfn> (a [=list=] of {{RouterRule}}s). It is initially an empty [=list=].
 
-    A [=/service worker=] has an associated <dfn>router rule count</dfn> (the number of registered {{RouterRule}}s). It is initially set to zero.
-
-    Note: [=router rule count=] is defined as the limit of the total number of registered router rules. This prevents the user agent from spending much resources by evaluating too many router rules. The limit is 1024.
-
     A [=/service worker=] is said to be <dfn>running</dfn> if its [=event loop=] is running.
 
     <section>
@@ -1604,7 +1600,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       };
     </pre>
 
-    Note: {{RouterCondition/_or}} and {{RouterCondition/not}} might have the other {{RouterCondition/_or}} or {{RouterCondition/not}} inside. To avoid spending much resources by the nested condition or performance penalty on evaluation, depth of such nested conditions can be limited.
+    A <dfn export id="dfn-count-router-condition-result">count router condition result</dfn> is a [=struct=] that consists of:
+      * A <dfn export id="dfn-count-router-condition-result-total-count" for="count router condition result">total count</dfn> (a number).
+      * A <dfn export id="dfn-dfn-count-router-condition-result-depth" for="count router condition result">depth</dfn> (a number).
 
     <section>
       <h4 id="register-router-method">{{InstallEvent/addRoutes(rules)|event.addRoutes(rules)}}</h4>
@@ -1620,6 +1618,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
             1. If running the [=Verify Router Condition=] algorithm with |rule|["{{RouterRule/condition}}"] and |serviceWorker| returns false, return [=a promise rejected with=] a {{TypeError}}.
             1. Append |rule| to |routerRules|.
         1. If |routerRules| [=list/contains=] a {{RouterRule}} whose {{RouterRule/source}} is either of "{{RouterSourceEnum/fetch-event}}" or "{{RouterSourceEnum/race-network-and-fetch-handler}}", and |serviceWorker|'s [=set of event types to handle=] does not [=set/contain=] {{ServiceWorkerGlobalScope/fetch!!event}}, return [=a promise rejected with=] a {{TypeError}}.
+        1. If running the [=Check Router Registration Limit=] with |routerRules| returns false, return [=a promise rejected with=] a {{TypeError}}.
         1. Set |serviceWorker|'s [=service worker/list of router rules=] to |routerRules|.
         1. Return [=a promise resolved with=] undefined.
 
@@ -3405,9 +3404,6 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               Note: For ease of understanding the router rule, the "or" condition is mutually exclusive with other conditions.
 
           1. Let |orConditions| be |condition|["{{RouterCondition/_or}}"].
-         
-              Note: To limit the resource usage and a condition evaluation time, |orConditions|'s [=list/size=] can be limited.
- 
           1. For each |orCondition| of |orConditions|:
               1. If running the [=Verify Router Condition=] algorithm with |orCondition| and |serviceWorker| returns false, return false.
           1. Set |hasCondition| to true.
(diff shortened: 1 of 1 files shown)


docs/index.bs
```diff
@@ -212,6 +212,10 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     A [=/service worker=] has an associated <dfn>list of router rules</dfn> (a [=list=] of {{RouterRule}}s). It is initially an empty [=list=].
 
+    A [=/service worker=] has an associated <dfn>router rule count</dfn> (the number of registered {{RouterRule}}s). It is initially set to zero.
```
yoshisatoyanagisawa, line 215: I guess we do not need a dedicated variable for this because [=list=] has the number of items.
https://infra.spec.whatwg.org/#list-size

docs/index.bs
```diff
@@ -3409,6 +3418,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
           1. If running the [=Verify Router Condition=] algorithm with |condition|["{{RouterCondition/not}}"] and |serviceWorker| returns false, return false.
           1. Set |hasCondition| to true.
+      1. If |hasCondition| is true, then:
+          1. Increament |serviceWorker|'s [=service worker/router rule count=] by one.
```
yoshisatoyanagisawa, line 3422: It sounds like the router rule count kept on incremented even if verify fails later. i.e. addRoutes() failed.
Also, if or/not is nested, they may cause +1 per nest.  Is it intended?

docs/index.bs
yoshisatoyanagisawa: A stepping back, in spite of the explanation, this variable does not match with the number of rules.  Then, the variable might be needed.
I feel the limitation in this way counterintuitive because the limit does not depend on the number of rules but depends on the number of rules or nests while it is saying limiting with the number of router rules.

docs/index.bs
```diff
@@ -1614,6 +1624,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
             1. If running the [=Verify Router Condition=] algorithm with |rule|["{{RouterRule/condition}}"] and |serviceWorker| returns false, return [=a promise rejected with=] a {{TypeError}}.
             1. Append |rule| to |routerRules|.
         1. If |routerRules| [=list/contains=] a {{RouterRule}} whose {{RouterRule/source}} is either of "{{RouterSourceEnum/fetch-event}}" or "{{RouterSourceEnum/race-network-and-fetch-handler}}", and |serviceWorker|'s [=set of event types to handle=] does not [=set/contain=] {{ServiceWorkerGlobalScope/fetch!!event}}, return [=a promise rejected with=] a {{TypeError}}.
+        1. If running the [=Check Router Registration Limit=] with |serviceWorker| returns false, return [=a promise rejected with=] a {{TypeError}}.
```
yoshisatoyanagisawa, line 1627: You might want to take |routerRules| instead of |serviceWorker| because |serviceWorker| should not have the latest |routerRules| yet.  It will be updated in L1628.

To avoid `addRoutes()` failure as a permanent failure, it is intended to verify before update, right?

docs/index.bs
sisidovski: Discussed offline. We don't need it anymore.

We have a rough agreement on the new approach, having two limits and validates them in between the step 5 and 6 in addRoutes().
https://w3c.github.io/ServiceWorker/#dom-installevent-addroutes

docs/index.bs
sisidovski: Discussed offline. We don't need it anymore.

We have a rough agreement on the new approach, having two limits and validates them in between the step 5 and 6 in addRoutes().
https://w3c.github.io/ServiceWorker/#dom-installevent-addroutes

docs/index.bs
sisidovski: Good catch! Updated to put |routerRules|.

docs/index.bs
```diff
@@ -1600,6 +1600,10 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       };
     </pre>
 
+    A <dfn export id="dfn-count-router-condition-result">count router condition result</dfn> is a [=struct=] that consists of:
```
yoshisatoyanagisawa, line 1603: nit but the struct might currently be used within this specification, and I guess we do not need to "export" them, unless we want other spec to refer this.

sisidovski: Done.

docs/index.bs
```diff
@@ -3456,6 +3461,57 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+      : Input
+      :: |routerRules|, a [=list of router rules=]
+      : Output
+      :: a boolean
+
+      Note: Router conditions can be complex and nested using {{RouterCondition/_or}} and {{RouterCondition/not}}. To prevent excessive processing, this algorithm introduces two limits. First, the total number of conditions, counting all nested conditions, cannot exceed 1024. Second, the nesting depth is limited to 10 levels to avoid exponential computation.
+
+      1. Let |result| be a [=count router condition result=].
+      1. Set |result|'s [=count router condition result/total count=] to 0.
```
yoshisatoyanagisawa, line 3475: What is the intention to set "total count" to 0 here instead of 1, and omit increment one in count-router-inner-conditions?

docs/index.bs
```diff
@@ -3456,6 +3461,57 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      1. Set |result|'s [=count router condition result/depth=] to 1.
+      1. Let |maxCount| be 1024.
```
yoshisatoyanagisawa, line 3477: I guess setting |maxCount| is to use the same value not only in check-router-registration-limit, but also in count-router-inner-conditions.
However, do we need to ask the browser implementor to implement the early return?

docs/index.bs
```diff
@@ -3456,6 +3461,57 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      :: |maxDepth|, a number
+      : Output
+      :: |result|, a [=count router condition result=]
+
+      1. Increment |result|'s [=count router condition result/total count=] by one.
+      1. If |result|'s [=count router condition result/total count=] exceeds |maxCount|, return |result|.
+      1. If |result|'s [=count router condition result/depth=] exceeds |maxDepth|, return |result|.
+      1. If |condition|["{{RouterCondition/_or}}"] [=map/exists=], then:
+          1. Increment |result|'s [=count router condition result/depth=] by one.
+          1. For each |orCondition| of |condition|["{{RouterCondition/_or}}"]:
```
yoshisatoyanagisawa, line 3503: I feel that the number of conditions under |orCondition| may not be counted?
e.g.
or: [{singleRule}, {singleRule}, {singleRule}]
|result| might be overwritten by the result for {singleRule} for three times, but we want |result|'s [=total count=] to be three.

docs/index.bs
sisidovski: IIUC, this line is the initializing part for "total count". The actual condition node counting is done inside count-router-inner-conditions. We don't start counting yet, so starting from zero before calling count-router-inner-conditions makes sense to me. Also, I don't think incrementing in count-router-inner-conditions? is not omitted. The returned value was used in below step.

```
 Set |result|'s [=count router condition result/total count=] to |currentResult|'s [=count router condition result/total count=].
 ```

File: docs/index.bs (hunk 6 above)
Comment (sisidovski) [reply docs/index.bs:3477]: Without early returns, the algorithm can easily introduces a stack overflow due to its recursion. Should we have a mechanism to prevent that?

File: docs/index.bs (hunk 7 above)
Comment (sisidovski) [reply docs/index.bs:3503]: I believe those are correctly counted? Let's walk through the current algorithm. btw, under the current design, `or: [{singleRule}, {singleRule}, {singleRule}]`, this total count is expected to be four (`or` should be counted as one condition).

1. We call count-router-inner-conditions with the root `or` condition. The total count will be one.
2. Starts the for loop with the first {singleRule}.
3. In the first sub-step, |result|'s `total count` will be two, because we pass "one" as `total count` to count-router-inner-conditions in this sub-step. `total count` will be incremented inside the algorithm.
4. The second {singleRule} is handled. `total count` will be three.
5. The third {singleRule} is handled. `total count` will be four.

Am I missing something?

File: docs/index.bs [hunk 8]
```diff
@@ -3456,6 +3461,57 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      1. Let |result| be a [=count router condition result=].
+      1. Set |result|'s [=count router condition result/total count=] to 0.
+      1. Set |result|'s [=count router condition result/depth=] to 1.
+      1. Let |maxCount| be 1024.
+      1. Let |maxDepth| be 10.
+      1. [=list/For each=] |rule| of |routerRules|:
+          1. Let |currentResult| be the result of running [=Count Router Inner Conditions=] with |rule|["{{RouterRule/condition}}"], |result|, |maxCount|, and |maxDepth|.
+          1. If |currentResult|'s [=count router condition result/total count=] exceeds |maxCount|, return false.
+          1. If |currentResult|'s [=count router condition result/depth=] exceeds |maxDepth|, return false.
+          1. Set |result|'s [=count router condition result/total count=] to |currentResult|'s [=count router condition result/total count=].
```
sisidovski, line 3483: Maybe we need to have a `Note:` why this step is needed?

docs/index.bs
yoshisatoyanagisawa: Upon the sync, I think we agreed to:
1. do the "early return" by checking conditions and width,
2. and avoid checking the value twice by making count-router-inner-conditions to return an error in addition.  check-router-registration-limit just checks the error is set or not, and it won't check the number.

Stepping back, we do want to know the number of conditions but we do not need to know the depth.  Can we move |depth| to the argument of count-router-inner-conditions instead of the struct?

docs/index.bs
yoshisatoyanagisawa: Upon the sync, I understand how this works.
The last result will be the |currentResult|'s total count, and it will be passed to the coming count-router-inner-conditions.  Therefore, the value will be updated like:
1. |currentResult| is passed to count-router-inner-conditions's argument
2. in the count-router-inner-conditions, the given argument will be updated
3. the updated value will be returned
4. |currentResult|'s total count will be set to the returned result's total count.

Therefore, even if the algorithm does not have add x + y like operation, the count will be updated.
I feel the data flow is difficult to track at once, and hope some notes to explain that as you have already noted below.

docs/index.bs
sisidovski: I think the latest commit simplified the flow and addressed the comment.

docs/index.bs
sisidovski: I've updated to set |result| to be the result of count-router-inner-conditions directly, and now we don't have to set `total count` explicitly. Hence note is not required anymore. WDYT?

docs/index.bs
yoshisatoyanagisawa: I guess the sentence does not exist any more.
Let me mark this resolved.

docs/index.bs
yoshisatoyanagisawa: lgtm. thank you.

docs/index.bs
yoshisatoyanagisawa: Maybe?
After reviewing this, I got familiar with the algorithm well and can easily read this like so.
However, I am not sure how people first look feel on this.  Let me delegate the decision to other reviewers on this.

yoshisatoyanagisawa:
Sorry for the recent update, but will you resolve the conflicts?

sisidovski:
> Sorry for the recent update, but will you resolve the conflicts?

Done

yoshisatoyanagisawa:
Thank you. lgtm.

yoshisatoyanagisawa:
@youennf @asutherland @domenic Will you take a look at the PR?

yoshisatoyanagisawa (approved):

yoshisatoyanagisawa:
Let me wait for a few days to gather thoughts from who joined https://github.com/w3c/ServiceWorker/issues/1746 before merge.

docs/index.bs
```diff
@@ -3473,6 +3478,51 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+      Note: Router conditions can be complex and nested using {{RouterCondition/_or}} and {{RouterCondition/not}}. To prevent excessive processing, this algorithm introduces two limits. First, the total number of conditions, counting all nested conditions, cannot exceed 1024. Second, the nesting depth is limited to 10 levels to avoid exponential computation.
+
+      1. Let |result| be a [=count router condition result=].
+      1. Set |result|'s [=count router condition result/condition count=] to 1024.
+      1. Set |result|'s [=count router condition result/quota exceeded=] to false.
+      1. [=list/For each=] |rule| of |routerRules|:
+          1. Set |result| to be the result of running [=Count Router Inner Conditions=] with |rule|["{{RouterRule/condition}}"], |result|, and 10.
+          1. If |result|'s [=count router condition result/quota exceeded=] is true, return false.
+      1. return true.
```
domenic, line 3497:
```suggestion
      1. Return true.
```

docs/index.bs
domenic: The algorithm seems pretty intuitive to me now without needing a note. It is clear that _result_ is mutated and the for-loop over all `_or` conditions means it will be mutated once for each sub-condition.

domenic (approved):

asutherland (approved):
Thank you!

Issue #1172, opened 2017-07-19
Carefully audit all uses of "in parallel" in the spec


bzbarsky:
I'm only partway through respondWith, and there's all sorts of things in there that should not be happening "in parallel", and other parts of this spec have similar issues (e.g hthttps://w3c.github.io/ServiceWorker/v1/#dom-cache-put does mutations on a shared data structure "in parallel"; it's not clear how that's supposed to work exactly).

Please audit all uses of "in parallel" to ensure they actually make sense.

@domenic @wanderview

jungkees:
It's true that I didn't fully consider that point. After seeing https://github.com/whatwg/html/pull/2774, I thought I should review all the cases indeed.

monica-ch:
Bug triage: Editors to prioritize

PR #1755, opened 2025-02-14
Editorial: Queue a task to resolve/reject promise or when fire an event.

monica-ch:
This is for addressing #1740.

***
<a href="https://pr-preview.s3.amazonaws.com/monica-ch/ServiceWorker/pull/1755.html" title="Last updated on Jul 15, 2026, 7:31 PM UTC (e14fa07)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1755/e91ddff...monica-ch:e14fa07.html" title="Last updated on Jul 15, 2026, 7:31 PM UTC (e14fa07)">Diff</a>


yoshisatoyanagisawa:
In https://github.com/w3c/ServiceWorker/pull/1744, I was asked to introduce the parallel queue.  I am revisiting it to see if the case also need to be applied here or not.

docs/index.bs
```diff
@@ -809,8 +809,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         1. Let |promise| be a new <a>promise</a>.
         1. Run the following substeps <a>in parallel</a>:
             1. Let |registration| be the result of running <a>Match Service Worker Registration</a> given |storage key| and |clientURL|.
```
yoshisatoyanagisawa, line 811: I am trying to understand the algorithm and just a question but, why you want to run "Let |registration| ..." step in parallel?
If it is within a queue, once |registration| is not null, it should be kept on non-null.  However, the current algorithm put it outside, during race condition, |promise| may be resolved undefined even after |promise| is resolved with the result.

domenic:
This diff looks locally reasonable to me. However, I am not caught up on the discussions about parallel queues vs. using the generic "in parallel", so I cannot say which is is best here. Maybe merging this first is good and then doing some overall work on using parallel queues later is the right approach.

---

To try to answer the more generic questions:

Ideally, specifications are written to match implementations:

- There are completely separate threads, which mostly do not touch.
  - There is one event loop thread per agent
  - There is a thread for each specific parallel queue
  - Every invocation of the generic "in parallel" can be thought of as spawning a new separate thread
- In particular, it's important to be very clear what objects and data structures are touchable from each spec thread
  - Objects with a strong connection to JavaScript, like promises or Web IDL interface objects, must only be touched from their corresponding event loop thread.
    - But see below for how we don't always perfectly follow this advice.
  - Simple primitive values like booleans or strings, or more complex "primitive-ish" values like sets of integers or Web IDL dictionaries/Infra maps whose values are only primitives, can be touched from any thread. However, you need to take care to avoid race conditions.
    - We usually pass these objects across threads implicitly, with no clear indication that we're doing so. E.g., by having "in parallel" steps close over a variable that was introduced in the event loop portion of an algorithm.
    - This is often fine because 80% of the time, these values just make a one-way trip from the event loop (where they came from the web developer as method arguments or similar) to the in parallel thread (where they get processed).
    - Sometimes we need to do something more complex. Then it's best to add informative notes that are explicit about how you're avoiding race conditions. E.g. see [this step in a spec I recently wrote](https://webmachinelearning.github.io/writing-assistance-apis/#create-an-ai-model-object:~:text=Let%20abortedDuringDownload%20be%20false.,but%20read%20from%20in%20parallel.).
  - Some objects represent "browser process" data structures, and as such you want to ensure you only touch them from "in parallel", or from a specific parallel queue.
    - Usually a specific parallel queue is better, but probably there are cases where it's unnecessary.
    - For example, the user agent's list of all shared worker global scopes is only accessed from the [shared worker manager](https://html.spec.whatwg.org/#shared-worker-manager) parallel queue. (But see below on `SharedWorkerGlobalScope`.)
- Sometimes there is an unfortunate mixing up between JavaScript objects and their "browser process" counterparts, which causes specs to end up touching those JS objects across multiple threads, even though this violates the above rules.
  - For example, there is a lot of information stored on the `Document` object in specifications, like URL or origin, which needs to be accessed from tons of different threads.
  - Or, the example linked above about shared worker manager is specified to look at the "list of all `SharedWorkerGlobalScope` objects", which violates the above rules, because `SharedWorkerGlobalScope` is a JavaScript object that should only exist on the event loop.
  - A fully rigorous spec for this would match implementations better. For example, it would have separate spec objects for `Document` and "browser-process document-related info". It would then need to specify the exact way in which this information is synchronized across the two representations. However, this is a lot of work, for unclear benefit.
  - An example of a spec which does this more correctly is Fetch, which separates [requests](https://fetch.spec.whatwg.org/#concept-request) from [`Request`s](https://fetch.spec.whatwg.org/#request-class).
    - However, the synchronization is kind of magical there, instead of explicit, so it's possible there are spec bugs or mismatches with the implementation, because the "requests" are accessed from multiple threads without explicit transfer algorithms, or explicit guards against racy multiple-writes.
    - But this is probably our best model today of how to separate things, while not making the spec excessively complicated.
    - The service worker spec seems to mostly follow this model.

---

How does this apply to the service worker spec, and to this PR?

Well, I think the main thing is to identify from which thread you intend "service worker registration" objects to be accessed. I get the impression it should be in parallel, not on an event loop thread. So that's why it's worth keeping something parallel-ish in `getRegistration()`, whether it's "in parallel" or a specific parallel queue. And then, you need to queue a task to resolve the promise. So that's why this PR looks reasonable to me.

Note that in this PR there is some thread crossing, where _registration_ is implicitly passed from "in parallel" (where it was created in step 8.1) to the event loop thread (in step 8.2.1 and 8.2.2). This is a bit sketchy, but mostly OK. In practice, what's probably happening here is a copy and snapshot: when "get the service worker registration object" consults _registration_'s installing worker in step 2.6, it's probably not doing some sort of cross-thread lookup at the actual _registration_ object, but instead looking at some key that was passed across threads in `getRegistration()` step 8.2.

A more fully rigorous version of this whole setup would have explicitly-obvious primitive keys (like string UUIDs, or something) that allow us to correspond between browser-process "service worker"/"service worker registration" objects and event loop `ServiceWorker`/`ServiceWorkerRegistration` objects. Then algorithms like `getRegistration()` would pass across those keys in a way that makes it clear we're taking a snapshot. Then back on the event loop thread we'd look up the right `ServiceWorker`/`ServiceWorkerRegistration` objects and use those to construct the return value we resolve the promise with. But that's a lot of work, for unclear gain. The current model of just using the identity of the "service worker"/"service worker registration" objects, instead of explicit primitive keys, and understanding that it's implicitly snapshotted when crossing thread boundaries, is reasonable enough.

If I were to suggest the most useful improvements to improving the spec's cross-thread hygiene, here would be my priority list:

1. PRs like this, which clarify which task source is used to resolve promises.
2. Notate every algorithm in Appendix A with whether they run "in parallel" or on the event loop. This might uncover some problems. (I've been doing this recently, e.g. [here](https://webmachinelearning.github.io/writing-assistance-apis/#supporting-creation:~:text=Assert:%20these%20steps%20are%20running%20on%20realm%E2%80%99s%20surrounding%20agent%E2%80%99s%20event%20loop.) and [here](https://webmachinelearning.github.io/writing-assistance-apis/#initialize-the-summarization-model:~:text=an%20AISummarizerCreateOptions%20options:-,Assert:%20these%20steps%20are%20running%20in%20parallel.,-Perform%20any%20necessary).)
3. Audit all "in parallel" algorithms to ensure they aren't touching JS objects.
4. Use parallel queues appropriately to avoid race conditions from separate "in parallel" situations.
5. Figure out the relationship between parallel queues and the spec's existing idea of "jobs". (Maybe @jakearchibald can comment on this, as I think he might have invented both concepts.) Potentially rebase "jobs" on parallel queues?
6. Consider some mechanism for being more clear about how and when things cross threads. (This is generally not a solved problem in the spec ecosystem. Above I discussed one method, of using explicitly primitive keys, but maybe there is something more lightweight.)

I think it is totally fine if only item 1 on this list gets done. Items like 2-4 can just be things you keep in mind when you write new spec text.

docs/index.bs
```diff
@@ -970,9 +971,10 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         1. Let |promise| be [=a new promise=].
         1. Run the following steps [=in parallel=]:
             1. Let |registration| be [=this=]'s associated [=/service worker registration=].
-            1. If |registration|'s [=active worker=] is null, [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}}, and abort these steps.
-            1. Set |registration|'s [=navigation preload enabled flag=].
-            1. Resolve |promise| with undefined.
+            1. [=Queue a task=] on |promise|'s [=relevant settings object=]'s [=responsible event loop=], using the [=DOM manipulation task source=], to run the following steps:
+                1. If |registration|'s [=active worker=] is null, [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}}, and abort these steps.
+                1. Set |registration|'s [=navigation preload enabled flag=].
```
yoshisatoyanagisawa, line 976: If this is for protecting [=navigation preload enabled flag=] to be updated in parallel, I guess we also need the code to read [=navigation preload enabled flag=] in Handle Fetch to be executed with queue?

docs/index.bs
```diff
@@ -1015,7 +1019,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
             1. Let |state| be a new {{NavigationPreloadState}} dictionary.
             1. If |registration|'s [=navigation preload enabled flag=] is set, set |state|["{{NavigationPreloadState/enabled}}"] to true.
             1. Set |state|["{{NavigationPreloadState/headerValue}}"] to |registration|'s [=navigation preload header value=].
-            1. Resolve |promise| with |state|.
+            1. [=Queue a task=] to resolve |promise| with |state|.
```
yoshisatoyanagisawa, line 1022: Will you specify which event loop to use?

docs/index.bs
```diff
@@ -1126,7 +1130,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         1. Run the following substeps <a>in parallel</a>:
             1. Set [=ServiceWorkerGlobalScope/service worker=]'s <a>skip waiting flag</a>.
             1. Invoke [=Try Activate=] with [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=].
-            1. Resolve |promise| with undefined.
+            1. [=Queue a task=] to resolve |promise| with undefined.
```
yoshisatoyanagisawa, line 1133: Will you specify which event loop to use?

docs/index.bs
```diff
@@ -1271,10 +1275,10 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
                 1. Let |source| be the result of [=getting the service worker object=] that represents |contextObject|'s [=relevant global object=]'s [=ServiceWorkerGlobalScope/service worker=] in |targetClient|.
                 1. Let |deserializeRecord| be <a abstract-op>StructuredDeserializeWithTransfer</a>(|serializeWithTransferResult|, |destination|'s [=relevant Realm=]).
 
-                    If this throws an exception, catch it, [=fire an event=] named {{ServiceWorkerContainer/messageerror!!event}} at |destination|, using {{MessageEvent}}, with the {{MessageEvent/origin}} attribute initialized to |origin| and the {{MessageEvent/source}} attribute initialized to |source|, and then abort these steps.
+                    If this throws an exception, catch it, [=queue a task=] to [=fire an event=] named {{ServiceWorkerContainer/messageerror!!event}} at |destination|, using {{MessageEvent}}, with the {{MessageEvent/origin}} attribute initialized to |origin| and the {{MessageEvent/source}} attribute initialized to |source|, and then abort these steps.
```
yoshisatoyanagisawa, line 1278: I feel this [=queue a task=] ambiguous.  It may not tell whose queue is used for the purpose.
I suppose it targetClient's what do you think?

docs/index.bs
```diff
@@ -1271,10 +1275,10 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
                 1. Let |messageClone| be |deserializeRecord|.\[[Deserialized]].
                 1. Let |newPorts| be a new [=frozen array type|frozen array=] consisting of all {{MessagePort}} objects in |deserializeRecord|.\[[TransferredValues]], if any.
-                1. [=Dispatch|Dispatch an event=] named {{Window/message!!event}} at |destination|, using {{MessageEvent}}, with the {{MessageEvent/origin}} attribute initialized to |origin|, the {{MessageEvent/source}} attribute initialized to |source|, the {{MessageEvent/data}} attribute initialized to |messageClone|, and the {{MessageEvent/ports}} attribute initialized to |newPorts|.
+                1. [=Queue a task=] to [=Dispatch|Dispatch an event=] named {{Window/message!!event}} at |destination|, using {{MessageEvent}}, with the {{MessageEvent/origin}} attribute initialized to |origin|, the {{MessageEvent/source}} attribute initialized to |source|, the {{MessageEvent/data}} attribute initialized to |messageClone|, and the {{MessageEvent/ports}} attribute initialized to |newPorts|.
```
yoshisatoyanagisawa, line 1281: ditto.

docs/index.bs
```diff
@@ -1384,8 +1388,8 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
             1. For each [=/service worker client=] |client| where the result of running [=obtain a storage key=] given |client| [=storage key/equals=] the associated [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=]'s [=service worker registration/storage key=]:
                 1. If |client|'s [=environment/id=] is not |id|, [=continue=].
                 1. Wait for either |client|'s [=environment/execution ready flag=] to be set or for |client|'s [=discarded flag=] to be set.
-                1. If |client|'s [=environment/execution ready flag=] is set, then invoke [=Resolve Get Client Promise=] with |client| and |promise|, and abort these steps.
-            1. Resolve |promise| with undefined.
+                1. If |client|'s [=environment/execution ready flag=] is set, [=queue a task=] on |promise|'s [=relevant settings object=]'s [=responsible event loop=], using the [=DOM manipulation task source=], to invoke [=Resolve Get Client Promise=] with |client| and |promise|, and abort these steps.
+            1. [=Queue a task=] to resolve |promise| with undefined.
```
yoshisatoyanagisawa, line 1392: Why not queue in Line 1391, and run invocation and resolve in its substeps?

yoshisatoyanagisawa: Also, is it guaranteed that client state wont change during the parallel execution, right?

docs/index.bs
```diff
@@ -1479,20 +1483,21 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
-                1. If |registration| is not the [=ServiceWorkerGlobalScope/service worker=]'s <a>containing service worker registration</a>, [=continue=].
-
-                    Note: |registration| will be null if the [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=] is [=service worker registration/unregistered=].
-
-                1. If |client|'s <a>active service worker</a> is not the [=ServiceWorkerGlobalScope/service worker=], then:
-                    1. Invoke <a>Handle Service Worker Client Unload</a> with |client| as the argument.
-                    1. Set |client|'s <a>active service worker</a> to [=ServiceWorkerGlobalScope/service worker=].
-                    1. Invoke <a>Notify Controller Change</a> algorithm with |client| as the argument.
-            1. Resolve |promise| with undefined.
+            1. [=Queue a task=] to run the following steps on the [=ServiceWorkerGlobalScope/service worker=]'s [=responsible event loop=] using the [=DOM manipulation task source=]:
```
yoshisatoyanagisawa, line 1486: Why this case put "For each [=/service worker client=] |client| where..." inside the queue while the `get(id)` is not?

docs/index.bs
```diff
@@ -2008,15 +2013,16 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
-            1. Let |p| be the result of running the algorithm specified in {{Cache/matchAll(request, options)}} method with |request| and |options|.
-            1. Wait until |p| settles.
-            1. If |p| rejects with an exception, then:
-                1. Reject |promise| with that exception.
-            1. Else if |p| resolves with an array, |responses|, then:
-                1. If |responses| is an empty array, then:
-                    1. Resolve |promise| with undefined.
-                1. Else:
-                    1. Resolve |promise| with the first element of |responses|.
+            1. [=Queue a task=] on |promise|'s [=relevant settings object=]'s [=responsible event loop=], using the [=DOM manipulation task source=], to run the following steps:
```
yoshisatoyanagisawa, line 2016: ditto

docs/index.bs
```diff
@@ -2044,7 +2050,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
                 1. [=list/For each=] |requestResponse| of |requestResponses|:
                     1. Add a copy of |requestResponse|'s response to |responses|.
             1. [=list/For each=] |response| of |responses|:
-                1. If |response|'s [=response/type=] is "`opaque`" and [=cross-origin resource policy check=] with |promise|'s [=relevant settings object=]'s [=environment settings object/origin=], |promise|'s [=relevant settings object=], "", and |response|'s [=filtered response/internal response=] returns <b>blocked</b>, then reject |promise| with a `TypeError` and abort these steps.
+                1. If |response|'s [=response/type=] is "`opaque`" and [=cross-origin resource policy check=], [=queue a task=] on |promise|'s [=relevant settings object=]'s [=environment settings object/origin=], |promise|'s [=relevant settings object=], "", and |response|'s [=filtered response/internal response=] returns <b>blocked</b>, then reject |promise| with a `TypeError` and abort these steps.
```
yoshisatoyanagisawa, line 2053: What does this mean?

docs/index.bs
```diff
@@ -2261,10 +2267,11 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       The <dfn method for="CacheStorage"><code>match(|request|, |options|)</code></dfn> method steps are:
 
         1. If |options|["{{MultiCacheQueryOptions/cacheName}}"] [=map/exists=], then:
-            1. Return [=a new promise=] |promise| and run the following substeps [=in parallel=]:
+            1. Let |promise| be [=a new promise=].
+            1. Run the following substeps [=in parallel=]:
```
yoshisatoyanagisawa, line 2271: Considering the possible modification of the map, I guess we want to execute followings inside a parallel queue.
It might be good to have a dedicated parallel queue for the Cache API.

docs/index.bs
```diff
@@ -2284,8 +2291,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         1. Let |promise| be [=a new promise=].
         1. Run the following substeps [=in parallel=]:
```
yoshisatoyanagisawa, line 2292: ditto.

docs/index.bs
```diff
@@ -2298,11 +2306,12 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
```
yoshisatoyanagisawa, line 2306: ditto.

docs/index.bs
```diff
@@ -2318,7 +2327,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
             1. Let |cacheJobPromise| be [=a new promise=].
             1. Run the following substeps [=in parallel=]:
```
yoshisatoyanagisawa, line 2328: ditto

docs/index.bs
```diff
@@ -2336,7 +2345,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
                 Note: The [=list/items=] in the result [=ordered set=] are in the order that their corresponding entry was added to the [=name to cache map=].
 
-            1. Resolve |promise| with |cacheKeys|.
+            1. [=Queue a task=] to resolve |promise| with |cacheKeys|.
```
yoshisatoyanagisawa, line 2348: ditto

docs/index.bs
monica-ch: In `Handle Fetch` algorithm I see [=navigation preload enabled flag=] is being used to check and also don't see they are running the steps are in [=in parallel=]

docs/index.bs
monica-ch: Yes you are right, it should be in a queue. Updated!

docs/index.bs
monica-ch: Updated the queue to resolve in substeps.

@yoshisatoyanagisawa Can you please elaborate on your third comment in this thread please

`For Client API, it might affect not only service worker but also clients. Do we need a parallel queue that covers both instead of the event loop for ServiceWorker itself?`

docs/index.bs
monica-ch: This shouldn't be in queue like get(id), reverted and update the queue task.

docs/index.bs
monica-ch: updated!

docs/index.bs
monica-ch: Can you please also elaborate your suggestion

docs/index.bs
yoshisatoyanagisawa: Stepping back and see if it is possible to update [=navigation prelaod flag=] from arbitrary location.
Then, I totally got confused.

`ServiceWorkerRegistration` has `NavigationPreloadManager` as a readonly object.  Then, I assume it is not possible to call `enable()`/`disable()` because it updates a field inside `NavigationPreloadManager`.  If it is allowed, then I suppose it technically possible to execute `enable()`/`disable()` from arbitrary ServiceWorker clients?  I suppose each client is loaded and executed asynchronously, and we might need to assume the case that the flag is changed during `Handle Fetch`.

Stepping back again, `Handle Fetch` just read a boolean flag, I guess the operation here is atomic.  Then, it might be fine to read/write it at the same time.

docs/index.bs
yoshisatoyanagisawa: I assume that Cache API can be executed in parallel.  Then, there can be a case that a |cacheName| map is deleted while a |cacheName|'s map is being accessed.
To prevent that happen, I guess access to the [=relevant name to cache map=] must not happen in parallel.
For the purpose, there might need to be a dedicated parallel queue in https://w3c.github.io/ServiceWorker/#cache-constructs, and all actions to the CacheStorage may need to be happen within the parallel queue.
I mean for `match()`, `has()`, `open()`, and `delete()`, they should be executed within the cache constructs's parallel queue.

The situation should also be the same for the [=cache=] object.  I guess each [=cache=] object will have the dedicated parallel queue, and `match()`, `matchAll()`, `add()`, `addAll()`, `put()`, `delete()`, and `keys()` might also executed within the parallel queue for the [=cache=] object.

What do you think?

docs/index.bs
yoshisatoyanagisawa: Sorry, I overlooked this.
I am reading this as multi thread programming, and data structure can be changed except it is executed within the queue that serialize the data structure manipulation.
When I read

> For each [service worker client](https://w3c.github.io/ServiceWorker/#dfn-service-worker-client) client where the result of running [obtain a storage key](https://storage.spec.whatwg.org/#obtain-a-storage-key) given client [equals](https://storage.spec.whatwg.org/#storage-key-equal) the associated [service worker](https://w3c.github.io/ServiceWorker/#serviceworkerglobalscope-service-worker)’s [containing service worker registration](https://w3c.github.io/ServiceWorker/#dfn-containing-service-worker-registration)’s [storage key](https://w3c.github.io/ServiceWorker/#service-worker-registration-storage-key):

I just wondered what happens if one of the service worker clients has been removed or gets execution ready flag during the sub step execution, and suggest to run Clients API algorithm within the dedicated parallel queue to prevent unexpected modifications to clients.

gterzian:
@domenic nice write-up at https://github.com/w3c/ServiceWorker/pull/1755#issuecomment-2661941708; consider expanding on it in some form that will see it being shared more widely and survive this PR.

monica-ch:
Thanks @yoshisatoyanagisawa for the thorough review, and thanks @domenic for the
[threading-model write-up](https://github.com/w3c/ServiceWorker/pull/1755#issuecomment-2661840247)
— that framing is really helpful.

I've just rebased this PR onto the latest `main`. Following Domenic's guidance
that *"it is totally fine if only item 1 on this list gets done"*, I'd like to
keep this PR scoped to **item 1** (queuing tasks for promise resolution / event
firing per #1740) and split out the larger parallel-queue work into follow-ups.

Proposed plan:

- **This PR (#1755):** land the task-queuing fixes as-is
- **NavigationPreloadManager:** reading a boolean flag is atomic (per your own
  follow-up conclusion) — I can add a short editorial note here, or defer.
  Preference?
- **Clients API — dedicated parallel queue:** file as follow-up (Domenic's item 4)
- **Cache / CacheStorage — dedicated parallel queue:** file as follow-up
  (Domenic's item 4; also related to #1740 and #1831)
- **Appendix A "in parallel" annotations:** file as follow-up (Domenic's item 2;
  umbrella #1172)

Per @gterzian's suggestion, I'll also mirror Domenic's write-up into #1172 so
it survives this PR and can be referenced from the follow-ups.

If this split works for you, I'll file the follow-ups and address the remaining
in-scope comments here.

index.bs
```diff
@@ -2281,12 +2290,14 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
         1. If |options|["{{MultiCacheQueryOptions/cacheName}}"] [=map/exists=], then:
-            1. Return [=a new promise=] |promise| and run the following substeps [=in parallel=]:
+            1. Let |promise| be [=a new promise=].
+            1. Run the following substeps [=in parallel=]:
                 1. [=map/For each=] |cacheName| → |cache| of the [=relevant name to cache map=]:
                     1. If |options|["{{MultiCacheQueryOptions/cacheName}}"] matches |cacheName|, then:
-                        1. Resolve |promise| with the result of running the algorithm specified in {{Cache/match(request, options)}} method of {{Cache}} interface with |request| and |options| (providing |cache| as thisArgument to the `\[[Call]]` internal method of {{Cache/match(request, options)}}.)
-                        1. Abort these steps.
-                1. Resolve |promise| with undefined.
+                        1. [=Queue a task=] on |promise|'s [=relevant settings object=]'s [=responsible event loop=], using the [=DOM manipulation task source=], to run the following steps:
```
yoshisatoyanagisawa, line 2299: We might need a queue for cache storage or data corruption may happen.  We might need to choose a queue carefully.

monica-ch: Added a dedicated parallel queue on name to cache map and routed CacheStorage.match/has/open/delete/keys through it, so all reads/writes on the shared map are serialized. Cache-object-level parallel queues (Cache.match/matchAll/add/addAll/put/delete/keys) and a Clients-API parallel queue remain deferred to the follow-ups per item 4.

yoshisatoyanagisawa:
Thanks for the updates, especially for addressing the CacheStorage race conditions!
Sorry, I understand @domenic suggested moving forward with item 1 here, but I feel the PR size is too large for me to approve with full confidence. Would it be possible to split the PR by feature? For example, having separate PRs for Cache API, NavigationPreloadManager, Clients API, and ServiceWorkerGlobalScope, and so on.

domenic:
Folks, just a quick request to stop @-mentioning me during my retirement :). I keep unsubscribing from this thread but then getting pulled back in.

monica-ch:
> Thanks for the updates, especially for addressing the CacheStorage race conditions! Sorry, I understand domenic suggested moving forward with item 1 here, but I feel the PR size is too large for me to approve with full confidence. Would it be possible to split the PR by feature? For example, having separate PRs for Cache API, NavigationPreloadManager, Clients API, and ServiceWorkerGlobalScope, and so on.

Make sense, split into 6 focused PRs off main, one per feature area. Please take a look when you get a chance!

getRegistration → [https://github.com/w3c/ServiceWorker/pull/new/queue-task-serviceworkercontainer](vscode-file://vscode-app/c:/Users/monicach/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
NavigationPreloadManager (4 methods) → [https://github.com/w3c/ServiceWorker/pull/new/queue-task-navigationpreload](vscode-file://vscode-app/c:/Users/monicach/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
skipWaiting → [https://github.com/w3c/ServiceWorker/pull/new/queue-task-serviceworkerglobalscope](vscode-file://vscode-app/c:/Users/monicach/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Client / Clients API → [https://github.com/w3c/ServiceWorker/pull/new/queue-task-clients-api](vscode-file://vscode-app/c:/Users/monicach/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Cache API → [https://github.com/w3c/ServiceWorker/pull/new/queue-task-cache-api](vscode-file://vscode-app/c:/Users/monicach/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
CacheStorage API (with the parallel queue) → [https://github.com/w3c/ServiceWorker/pull/new/queue-task-cachestorage-api](vscode-file://vscode-app/c:/Users/monicach/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

PR #1756, opened 2025-03-03
ServiceWorkerAutoPreload

sisidovski:
This is the PR to support [ServiceWorkerAutoPreload](https://github.com/explainers-by-googlers/service-worker-auto-preload).

ServiceWorkerAutoPreload is defined as an optional browser optimization so that the browser minimizes the cost of the bootstrap by dispatching the navigational network request before starting the service worker. The dispatched network request is consumed by `fetch(e.request)` in the fetch handler after the bootstrap, or consumed by the client as a fallback request if the fetch handler doesn't call `respondWith()`.

The developer can disable this optimization by opting out via setting the router rule provided by the static routing API.  Specifically, the router source "network" works as an opt-out signal.

Please read [the explainer](https://github.com/WICG/service-worker-auto-preload) for more detail.

***
<a href="https://pr-preview.s3.amazonaws.com/sisidovski/ServiceWorker/pull/1756.html" title="Last updated on Jul 28, 2025, 3:15 AM UTC (3daef71)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1756/a402c93...sisidovski:3daef71.html" title="Last updated on Jul 28, 2025, 3:15 AM UTC (3daef71)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3228,6 +3228,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       1. Let |shouldSoftUpdate| be true if any of the following are true, and false otherwise:
           * |request| is a [=non-subresource request=].
           * |request| is a [=subresource request=] and |registration| is [=stale=].
+      1. Let |raceResponse| be null. 
       1. Let |source| be null.
       1. If |activeWorker|'s [=service worker/list of router rules=] [=list/is not empty=]:
           1. Set |timingInfo|’s [=service worker timing info/worker router evaluation start=] to the [=coarsened shared current time=] given |useHighResPerformanceTimers|.
@@ -3263,7 +3264,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
                   1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
                   1. Let |queue| be an empty [=queue=] of [=race result=].
                   1. Let |raceFetchController| be null.
-                  1. Let |raceResponse| be a [=race response=] whose [=race response/value=] is "<code>pending</code>".
+                  1. Set |raceResponse| to be a [=race response=] whose [=race response/value=] is "<code>pending</code>".
                   1. Run the following substeps [=in parallel=], but [=abort when=] |fetchController|'s [=fetch controller/state=] is "<code>terminated</code>" or "<code>aborted</code>":
                       1. Set |raceFetchController| to the result of calling [=fetch=] given |request|, with [=fetch/processResponse=] set to the following steps given a [=/response=] |raceNetworkRequestResponse|:
                           1. Set |raceResponse|'s [=race response/value=] to |raceNetworkRequestResponse|.
@@ -3316,7 +3317,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               Note: A user agent may speculatively dispatch a network request in parallel with creating a fetch event in order to minimize the bootstrap cost.
 
               1. Let |autoPreloadFetchController| be null.
-              1. Let |raceResponse| be a [=race response=] whose [=race response/value=] is "<code>pending</code>".
+              1. Set |raceResponse| to be a [=race response=] whose [=race response/value=] is "<code>pending</code>".
               1. Run the following substeps [=in parallel=], but [=abort when=] |fetchController|'s [=fetch controller/state=] is "<code>terminated</code>" or "<code>aborted</code>":
                   1. Set |autoPreloadFetchController| to the result of calling [=fetch=] given |request|, with [=fetch/processResponse=] set to the following steps given a [=/response=] |autoPreloadRequestResponse|:
                   1. If |autoPreloadRequestResponse|'s [=response/status=] is [=ok status=], set |raceResponse|'s [=race response/value=] to |autoPreloadRequestResponse|.
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3228,10 +3228,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       1. Let |shouldSoftUpdate| be true if any of the following are true, and false otherwise:
           * |request| is a [=non-subresource request=].
           * |request| is a [=subresource request=] and |registration| is [=stale=].
-      1. Let |source| be null.
       1. If |activeWorker|'s [=service worker/list of router rules=] [=list/is not empty=]:
           1. Set |timingInfo|’s [=service worker timing info/worker router evaluation start=] to the [=coarsened shared current time=] given |useHighResPerformanceTimers|.
-          1. Set |source| to be the result of running the [=Get Router Source=] algorithm with |registration|'s <a>active worker</a> and |request|.
+          1. Let |source| be the result of running the [=Get Router Source=] algorithm with |registration|'s <a>active worker</a> and |request|.
           1. If |source| is non-null, then:
               1. Set |timingInfo|'s [=service worker timing info/worker matched router source=] be set to |source|, and [=service worker timing info/worker final router source=] be set to {{RouterSourceEnum/"network"}}.
               1. If |source| is {{RouterSourceEnum/"network"}}:
@@ -3312,7 +3311,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               1. [=If aborted=], then:
                   1. Let |deserializedError| be the result of [=deserialize a serialized abort reason=] given null and |workerRealm|.
                   1. [=fetch controller/Abort=] |preloadFetchController| with |deserializedError|.
-          1. Else if |source| is not "{{RouterSourceEnum/fetch-event}}", a user agent may run the following substeps:
+          1. Else if |timingInfo|'s [=service worker timing info/worker matched router source=] is not "{{RouterSourceEnum/fetch-event}}", a user agent may run the following substeps:
 
               Note: A user agent may speculatively dispatch a network request in parallel with creating a fetch event in order to minimize the bootstrap cost.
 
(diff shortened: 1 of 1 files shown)


sisidovski:
@yoshisatoyanagisawa @youennf I think now it's ready for the review. Could you take a look? Thank you!

docs/index.bs
```diff
@@ -3228,9 +3228,11 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       1. Let |shouldSoftUpdate| be true if any of the following are true, and false otherwise:
           * |request| is a [=non-subresource request=].
           * |request| is a [=subresource request=] and |registration| is [=stale=].
+      1. Let |raceResponse| be null.
```
yoshisatoyanagisawa, line 3231: I guess |raceResponse| can go to the line before L3293 to make its lifetime shorter?
We do not need to share |raceResponse| between "race-network-and-fetch-event" and SWAutpPreload.

yoshisatoyanagisawa:
No need to do that in this PR, but I feel steps to create |raceResponse| can be merged with "race-network-and-fetch-event" case.

docs/index.bs
sisidovski: That's true. Moved this to L3319.

sisidovski: Ah no, we have to pass `raceResponse` in L3328.

> 1. Let |fetchResult| be the result of [=Create Fetch Event and Dispatch=] with |request|, |registration|, |useHighResPerformanceTimers|, |timingInfo|, |workerRealm|, |reservedClient|, |preloadResponse|, and |raceResponse|.

So I think the current position is reasonable?

sisidovski: Discussed offline, we don't have to use the same variable name.

docs/index.bs
```diff
@@ -3288,29 +3288,44 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+                      To [=fetch/processResponse=] for |navigationPreloadResponse|, run these substeps:
+
+                      1. If |navigationPreloadResponse|'s [=response/type=] is "`error`", reject |preloadResponse| with a `TypeError` and terminate these substeps.
+                      1. Associate |preloadResponseObject| with |navigationPreloadResponse|.
+                      1. Resolve |preloadResponse| with |preloadResponseObject|.
+              1. [=If aborted=], then:
+                  1. Let |deserializedError| be the result of [=deserialize a serialized abort reason=] given null and |workerRealm|.
+                  1. [=fetch controller/Abort=] |preloadFetchController| with |deserializedError|.
+          1. Else if |timingInfo|'s [=service worker timing info/worker matched router source=] is not "{{RouterSourceEnum/fetch-event}}", a user agent may run the following substeps:
```
sisidovski, line 3314: Updated not to introduce |source| here, use |timingInfo|'s [=service worker timing info/worker matched router source=] instead.

yoshisatoyanagisawa: Thanks.

yoshisatoyanagisawa:
The specification change itself looks good to me.  However, let me wait for other vendors to agree the change.

docs/index.bs
```diff
@@ -3288,29 +3288,44 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+                  1. If |autoPreloadRequestResponse|'s [=response/status=] is [=ok status=], set |responseForAutoPreload|'s [=race response/value=] to |autoPreloadRequestResponse|.
+              1. [=If aborted=] and |autoPreloadFetchController| is not null, then:
+                  1. [=fetch controller/Abort=] |autoPreloadFetchController|.
+                  1. Set |responseForAutoPreload| to a [=race response=] whose [=race response/value=] is null.
+              1. Resolve |preloadResponse| with undefined.
       1. Else, resolve |preloadResponse| with undefined.
-      1. Let |fetchResult| be the result of [=Create Fetch Event and Dispatch=] with |request|, |registration|, |useHighResPerformanceTimers|, |timingInfo|, |workerRealm|, |reservedClient|, |preloadResponse|, and null.
+      1. Let |fetchResult| be the result of [=Create Fetch Event and Dispatch=] with |request|, |registration|, |useHighResPerformanceTimers|, |timingInfo|, |workerRealm|, |reservedClient|, |preloadResponse|, and |responseForAutoPreload|.
```
youennf, line 3328: In the current WebKit implementation, if service worker does not call respondWith, the process doing networking is notified and it will reuse the responseForAutoPreload directly.
AIUI, the mechanism as described here will likely reuse the response in the renderer process, which might a bit slower. I am not sure whether this is observable so maybe the spec is fine as is.

yoshisatoyanagisawa: I feel you assume https://w3c.github.io/ServiceWorker/#create-fetch-event-and-dispatch-algorithm is handled in the renderer process.  I understand that the Chromium implementation is actually not.  Most of Step 17 is handled in the renderer process, but other steps in the algorithm would be handled in the browser process.

docs/index.bs
sisidovski: We frame the autopreload is an optional optimization that the browser can apply at its choosing. So we want to use the static routing API as an opt-out mechanism. If we only invoke the autopreload only when the matched router source is `fetch-event`, that is more like an opt-in, not opt-out.

docs/index.bs
```diff
@@ -3288,29 +3288,44 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+                      1. If |navigationPreloadResponse|'s [=response/type=] is "`error`", reject |preloadResponse| with a `TypeError` and terminate these substeps.
+                      1. Associate |preloadResponseObject| with |navigationPreloadResponse|.
+                      1. Resolve |preloadResponse| with |preloadResponseObject|.
+              1. [=If aborted=], then:
+                  1. Let |deserializedError| be the result of [=deserialize a serialized abort reason=] given null and |workerRealm|.
+                  1. [=fetch controller/Abort=] |preloadFetchController| with |deserializedError|.
+          1. Else if |timingInfo|'s [=service worker timing info/worker matched router source=] is not "{{RouterSourceEnum/fetch-event}}", a user agent may run the following substeps:
+
+              Note: A user agent may speculatively dispatch a network request in parallel with creating a fetch event in order to minimize the bootstrap cost.
```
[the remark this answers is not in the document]
sisidovski: In the explainer, we mention two possible scenarios that the browser may invoke the autopreload.
https://github.com/WICG/service-worker-auto-preload?tab=readme-ov-file#eligibility-criteria

I don't think we can fully specify when it's used since some conditions are not deterministic, but do you think we should say more here?

docs/index.bs
```diff
@@ -3288,29 +3288,44 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+
+              1. Let |autoPreloadFetchController| be null.
+              1. Set |responseForAutoPreload| to be a [=race response=] whose [=race response/value=] is "<code>pending</code>".
+              1. Run the following substeps [=in parallel=], but [=abort when=] |fetchController|'s [=fetch controller/state=] is "<code>terminated</code>" or "<code>aborted</code>":
+                  1. Set |autoPreloadFetchController| to the result of calling [=fetch=] given |request|, with [=fetch/processResponse=] set to the following steps given a [=/response=] |autoPreloadRequestResponse|:
+                  1. If |autoPreloadRequestResponse|'s [=response/status=] is [=ok status=], set |responseForAutoPreload|'s [=race response/value=] to |autoPreloadRequestResponse|.
```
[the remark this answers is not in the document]
sisidovski: Thanks! Updated to set the response to `responseForAutoPreload` regardless of the status.

sisidovski:
@youennf Could you take a look this again? Thanks!

docs/index.bs
youennf: Oh, that is right, the algorithm is perfectly fine as is.

youennf:
LGTM overall. Two suggestions as detailed in the comments, but they are not blockers.

docs/index.bs
sisidovski: Thank you! That sounds reasonable. Added `Assert: [=service worker timing info/worker matched router source=] is null.`.

docs/index.bs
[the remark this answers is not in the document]
sisidovski: Yes. Let's keep it as is for now.

Just to note: we're happy with rephrasing it with SHOULD.

sisidovski (commented):
Thanks!

docs/index.bs
```diff
@@ -3288,29 +3288,45 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+          1. Else if |timingInfo|'s [=service worker timing info/worker matched router source=] is not "{{RouterSourceEnum/fetch-event}}", a user agent may run the following substeps:
+
+              Note: A user agent may speculatively dispatch a network request in parallel with creating a fetch event in order to minimize the bootstrap cost.
+
+              1. Assert: [=service worker timing info/worker matched router source=] is null.
+              1. Let |autoPreloadFetchController| be null.
+              1. Set |responseForAutoPreload| to be a [=race response=] whose [=race response/value=] is "<code>pending</code>".
+              1. Run the following substeps [=in parallel=], but [=abort when=] |fetchController|'s [=fetch controller/state=] is "<code>terminated</code>" or "<code>aborted</code>":
+                  1. Set |autoPreloadFetchController| to the result of calling [=fetch=] given |request|, with [=fetch/processResponse=] set to the following steps given a [=/response=] |autoPreloadRequestResponse|:
+                  1. Set |responseForAutoPreload|'s [=race response/value=] to |autoPreloadRequestResponse|.
```
yoshisatoyanagisawa, line 3323: You may need another indent here?

yoshisatoyanagisawa:
ok, I think all comments are addressed.

PR #1760, opened 2025-03-18
Editorial:  Fix broken links/fragments

monica-ch:
This is for addressing https://github.com/w3c/ServiceWorker/issues/1758 that fixes broken links fragments and updates redirects as a result of [validator](https://validator.w3.org/checklink?uri=https%3A%2F%2Fwww.w3.org%2FTR%2F2025%2FCRD-service-workers-20250306%2F&hide_type=all&recursive=on&depth=&check=Check)

***
<a href="https://pr-preview.s3.amazonaws.com/monica-ch/ServiceWorker/pull/1760.html" title="Last updated on Mar 19, 2025, 8:25 PM UTC (376e647)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1760/b8255db...monica-ch:376e647.html" title="Last updated on Mar 19, 2025, 8:25 PM UTC (376e647)">Diff</a>

PR #1762, opened 2025-03-24
Editorial: Revise reference links with bikeshed

monica-ch:
This is the continuation of PR https://github.com/w3c/ServiceWorker/pull/1760 to fix reference links with dfn's.

This PR revise the links for HTML and RFC2397.

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -88,6 +88,12 @@ spec: ecma-262; urlPrefix: https://tc39.es/ecma262/
             text: statement
             text: declaration
 
+spec: html; urlPrefix: https://html.spec.whatwg.org/
+    type: dfn
+        text: HTML Standard;
+        text: set up; url:set-up-a-window-environment-settings-object
+        text: created; url:creating-a-new-browsing-context
+
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     type: attribute
         urlPrefix: comms.html
@@ -116,6 +122,10 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
         text: response; for: Response; url: concept-response-response
         text: request; for: Request; url: concept-request-request
         text: HTTP fetch; for: /; url: concept-http-fetch
+
+spec: rfc2397; urlPrefix: https://datatracker.ietf.org/doc/html/rfc2397
+    type: dfn
+        text: data: URL; url: section-2
 
 spec: rfc8288; urlPrefix: https://tools.ietf.org/html/rfc8288
     type: dfn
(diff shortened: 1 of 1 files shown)


docs/index.bs
```diff
@@ -322,33 +332,33 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     *The rest of the section is non-normative.*
 
-    Issue: The behavior in this section is not fully specified yet and will be specified in [HTML Standard](https://html.spec.whatwg.org). The work is tracked by the [issue](https://github.com/w3c/ServiceWorker/issues/765) and the [pull request](https://github.com/whatwg/html/pull/2809).
+    Issue: The behavior in this section is not fully specified yet and will be specified in [=HTML Standard=]. The work is tracked by the [issue](https://github.com/w3c/ServiceWorker/issues/765) and the [pull request](https://github.com/whatwg/html/pull/2809).
```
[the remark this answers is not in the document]
yoshisatoyanagisawa: I believe it is fine to leave github links as-is.

docs/index.bs
```diff
@@ -88,6 +88,12 @@ spec: ecma-262; urlPrefix: https://tc39.es/ecma262/
             text: statement
             text: declaration
 
+spec: html; urlPrefix: https://html.spec.whatwg.org/
+    type: dfn
+        text: HTML Standard;
+        text: set up; url:set-up-a-window-environment-settings-object
```
yoshisatoyanagisawa, line 94: I understand that you do not want to write a long name to represent it inside the document.  However, I feel referring this with [=set up=] too ambiguous.  Can this be [=set up a window environment settings object=] instead?
So do [=created=].

yoshisatoyanagisawa:
Thanks for working on this.
I understand you prefer to set shorter-names, but  I am afraid of the name conflicts.

docs/index.bs
monica-ch: Updated, agree with future conflicts.

docs/index.bs
```diff
@@ -88,6 +88,12 @@ spec: ecma-262; urlPrefix: https://tc39.es/ecma262/
             text: statement
             text: declaration
 
+spec: html; urlPrefix: https://html.spec.whatwg.org/
```
mkruisselbrink, line 91: I would strongly discourage adding references to other specs in the "anchors" block like this. If there are definitions in other specs that you need to reference they should be exported from the other spec, rather than worked around like this (and I think even if they aren't exported, you can still link to them without adding them to anchors by explicitly mentioning the spec).

And the same probably also applies to all the existing html fetch and storage references below.

yoshisatoyanagisawa: There are actually four ways to create a link to other specification:
1. W3C standard exports, and use it directly `[=something=]`, which might be what you recommend.
2. the "anchors" block to explain the combination of spec and text to be linked.  Then, use it in the .bs file.
3. `[=something=](https://example.com)` style link to directly add a link.
4. `<a href="https://example.com">something</a>` to directly add a link.

How do you use them differently?  Or, are there a style guide or so to explain the recommended way?  In https://github.com/w3c/ServiceWorker/pull/1760#issuecomment-2735224987, I suggested to modify 3 and 4 to either of 1 or 2.  I am quite sure that brought this change.

mkruisselbrink: As https://speced.github.io/bikeshed/#custom-dfns sort of hints at, the "anchors" block exists to link to specs that are not in the autolinking database. If a spec is in the autolinking database (as w3c specs all should be) you should never need to use it. You can use an explicit `<l spec='html'>` indication to link to unexported definitions (although you should probably still work with the target spec to make sure the definition does get exported). Maybe link-defaults also work to explicitly link to an unexported definition. But in either case using anchors to link to specs means for one that you won't get any bikeshed warnings if the destination doesn't exist (or stops to exist), and in general just makes the spec way less maintainable.

monica-ch: I created https://github.com/w3c/ServiceWorker/issues/1763 to track this work. Thanks for bringing this.

yoshisatoyanagisawa: Thanks for the reply and sorry for the delay, let me dump what I understood:
1. use the W3C spec exported dfns (i.e. autolinking database) as much as possible.  It allows bikeshed to find out the broken links when it happens.
2. if it is not possible (e.g. not W3C spec or so), the "anchors" block or the explicit indication can be used as a fallback, but makes maintainability bad because of bikeshed does not find the broken link.

Let me discuss the next actions in the #1763.

PR #1764, opened 2025-04-16
Add race-network-and-cache source to static routing api. 

monica-ch:
This PR adds new source `race-network-and-cache` to the SW static routing api per https://github.com/WICG/service-worker-static-routing-api/blob/main/final-form.md.

***
<a href="https://pr-preview.s3.amazonaws.com/monica-ch/ServiceWorker/pull/1764.html" title="Last updated on Aug 18, 2026, 8:18 PM UTC (e0f2990)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1764/3ff30ad...monica-ch:e0f2990.html" title="Last updated on Aug 18, 2026, 8:18 PM UTC (e0f2990)">Diff</a>


docs/index.bs
```diff
@@ -3258,6 +3260,38 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+                          1. Let |requestResponse| be the first element of |requestResponses|.
+                          1. Let |response| be |requestResponse|'s response.
+                          1. If |response|'s [=response/type=] is "`opaque`", and |globalObject| is null:
+                              1. Set |globalObject| to the result of running [=Setup ServiceWorkerGlobalScope=] with |activeWorker|.
+                          1. If |globalObject| is null, [=continue=].
+                          1. If |response|'s [=response/type=] is "`opaque`", and [=cross-origin resource policy check=] with |globalObject|'s [=environment settings object/origin=], |globalObject|, "", and |response|'s [=filtered response/internal response=] returns <b>blocked</b>, then [=continue=].
+                          1. [=queue/Enqueue=] |response| to |queue|.
+                          1. If |raceFetchController| is not null, [=fetch controller/abort=] |raceFetchController|.
+              1. Wait until |queue| is not empty.
+              1. If |queue| is empty, return null.
```
yoshisatoyanagisawa, line 3293: What does this step mean?
L3292 says "Wait until |queue| is not empty.", then L3293 should not happen?

yoshisatoyanagisawa: I guess L3292 might wait for:
1. |queue| is fulfilled
2. or, both of network fetch and cache lookup failure.

But stepping back, race-network-and-fetch-event may also need both failure case.

yoshisatoyanagisawa:
@sisidovski Will you take a look?
I assume it mostly fine but I feel all failure scenarios may not work well for both race-network-and-fetch-event and race-network-and-cache.

docs/index.bs
monica-ch: Added flags to track if they are finished and Wait until |queue| is not empty step is updated so it won't wait indefinitely if both fails. The next step handles if they fail and |queue| is empty our algorithm returns.

Is this change, okay? Feel free to suggest for corrections or if it is unclear

docs/index.bs
```diff
@@ -3256,7 +3261,45 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
                   1. Let |fetchHandlerResponse| be the result of [=Create Fetch Event and Dispatch=] with |request|, |registration|, |useHighResPerformanceTimers|, |timingInfo|, |workerRealm|, |reservedClient|, |preloadResponse|, and |raceResponse|.
                   1. If |fetchHandlerResponse| is not null and not a [=network error=], and |raceFetchController| is not null, [=fetch controller/abort=] |raceFetchController|.
                   1. [=queue/Enqueue=] |fetchHandlerResponse| to |queue|.
-              1. Wait until |queue| is not empty.
+                  1. Set |fetchHandlerCompleted| to true.
+              1. Wait until |queue| is not empty or (|networkFetchCompleted| is true and |fetchHandlerCompleted| is true).
```
yoshisatoyanagisawa, line 3265: @domenic Can I ask your advice on how this kinds of thing is usually written in the spec?
The explanation sounds reasonable as code, but I am afraid it is not usual way of writing spec.

docs/index.bs
```diff
@@ -3242,12 +3244,15 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
               1. Run the following substeps [=in parallel=], but [=abort when=] |fetchController|'s [=fetch controller/state=] is "<code>terminated</code>" or "<code>aborted</code>":
                   1. Set |raceFetchController| to the result of calling [=fetch=] given |request|, with [=fetch/processResponse=] set to the following steps given a [=/response=] |raceNetworkRequestResponse|:
                       1. If |raceNetworkRequestResponse|'s [=response/status=] is [=ok status=], then:
                           1. Set |raceResponse|'s [=race response/value=] to |raceNetworkRequestResponse|.
                           1. [=queue/Enqueue=] |raceNetworkRequestResponse| to |queue|.
                       1. Otherwise, set |raceResponse|'s [=race response/value=] to a [=network error=].
+                  1. Set |networkFetchCompleted| to true.
               1. [=If aborted=] and |raceFetchController| is not null, then:
                   1. [=fetch controller/Abort=] |raceFetchController|.
                   1. Set |raceResponse| to a [=race response=] whose [=race response/value=] is null.
```
yoshisatoyanagisawa, line 3258: I guess you might also need to set |networkFetchCompleted| to true.  That is because network fetch will not move forward on abort.

docs/index.bs
```diff
@@ -3256,7 +3261,45 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+                  1. Set |raceFetchController| to the result of calling [=fetch=] given |request|, with [=fetch/processResponse=] set to the following steps given a [=/response=] |raceNetworkRequestResponse|:
+                      1. If |raceNetworkRequestResponse|'s [=response/status=] is [=ok status=], then:
+                          1. Set |raceResponse|'s [=race response/value=] to |raceNetworkRequestResponse|.
+                          1. [=queue/Enqueue=] |raceNetworkRequestResponse| to |queue|.
+                      1. Otherwise, set |raceResponse|'s [=race response/value=] to a [=network error=].
+                  1. Set |networkFetchCompleted| to true.
+              1. [=If aborted=] and |raceFetchController| is not null, then:
+                  1. If |raceFetchController|'s [=fetch controller/state=] is not "<code>terminated</code>", [=fetch controller/abort=] |raceFetchController|.
+                  1. Set |raceResponse| to a [=race response=] whose [=race response/value=] is null.
```
yoshisatoyanagisawa, line 3284: Also |networkFetchComplete| for this.

docs/index.bs
```diff
@@ -3256,7 +3261,45 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+              1. Resolve |preloadResponse| with undefined.
+              1. Run the following substeps [=in parallel=]:
+                  1. [=map/For each=] |cacheName| → |cache| of the |registration|'s [=service worker registration/storage key=]'s [=name to cache map=]:
```
yoshisatoyanagisawa, line 3287: I feel this as a duplication of the cache lookup process above.  Can we factor this procedure out to the independent algorithm?

docs/index.bs
yoshisatoyanagisawa: Thanks for the update.
The algorithm looks good except for the abort network scenario.
I want a spec expert to check the way to write the procedure down to specification.

docs/index.bs
domenic: Are you concerned mainly about the parentheses? I think they are OK since they are easy to understand, but yes, it is unusual. The pattern I would use would be:

> Wait until any of the following are true:
> 
> * |queue| is not empty; or
> * |networkFetchCompleted| is true and |fetchHandlerCompleted| is true.

yoshisatoyanagisawa: Yes, I wanted to know how people usually write this because I have not seen specs using parentheses.
And, thank you for the advice.

@monica-ch Will you revise the sentences as @domenic's way?

docs/index.bs
```diff
@@ -1591,14 +1591,16 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
       dictionary RouterSourceDict {
         DOMString cacheName;
+        DOMString raceNetworkAndCacheCacheName;
```
sisidovski, line 1594: I'm not sure if we need a dedicated field to specify a cache name for race-network-and-cache. Also, could you give some example of how the final API shape will look like?

docs/index.bs
sisidovski: I'm not sure if we really really need to introduce `networkFetchCompleted` and `fetchHandlerCompleted`. For `race-network-and-fetch-handler`, the race is expressed by `queue`. Why we need these two flags?

It looks `networkFetchCompleted` doesn't wait for the promise resolve. So assuming `queue` can be still empty even when `networkFetchCompleted` is true. This will return null, which means the fetch event fallback in the handle fetch algorithm.

Also not sure why we should wait for `fetchHandlerCompleted` as well. This means we have to wait for both the network and fetch handler completion.

IIUC `race-network-and-fetch-handler` won't run the race anymore after this change.

yoshisatoyanagisawa: This change should be due to https://github.com/w3c/ServiceWorker/pull/1764#discussion_r2046691943.
I concerned that if both network and cache failed, nothing is enqueued, and waiting for queue won't finish.  I stepped back and saw the `race-network-and-fetch-handler`, and felt it may also have the same issue.
Having the way to know network and cache lookup completion should be needed to avoid the infinite queue waiting.

docs/index.bs
monica-ch: using a cacheName in the source indicate the source as "cache".  It is hard to identify the source if cacheName is used for both sources.

  e.addRoutes([
    {
      condition: {
        urlPattern: {pathname: "/race_network_and_cache_test_1"}
      },
      source: "race-network-and-cache" // Race b/w network and cache and without cache name indicates search all caches.
    },
     {
      condition: {
        urlPattern: "/race_network_and_cache_test_2",
      },
      source: {raceNetworkAndCacheCacheName: "test"} // Having `raceNetworkAndCacheCacheName` tells us to use race b/w network and cache source with the provided cacheName
     condition: {
        urlPattern: "/cache",
      },
      source: {cacheName: "test"} // existing example for cache source
  },
  ]);

This is in-line with the final form for Race-Network-and-cache - https://github.com/WICG/service-worker-static-routing-api/blob/main/final-form.md#race-network-and-cache-storage

@yoshisatoyanagisawa Correct me if this isn't expected

yoshisatoyanagisawa: Yes, that is my understanding.

docs/index.bs
monica-ch: Updated, please take a look and correct me if anything looks wrong.

docs/index.bs
```diff
@@ -3220,43 +3222,67 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
               1. Run the following substeps [=in parallel=]:
                   1. Let |fetchHandlerResponse| be the result of [=Create Fetch Event and Dispatch=] with |request|, |registration|, |useHighResPerformanceTimers|, |timingInfo|, |workerRealm|, |reservedClient|, |preloadResponse|, and |raceResponse|.
                   1. If |fetchHandlerResponse| is not null and not a [=network error=], and |raceFetchController| is not null, [=fetch controller/abort=] |raceFetchController|.
                   1. [=queue/Enqueue=] |fetchHandlerResponse| to |queue|.
-              1. Wait until |queue| is not empty.
+                  1. Set |fetchHandlerCompleted| to true.
+              1. Wait until any of the following are true:
+                  1. |queue| is not empty.
+                  1. Both |networkFetchCompleted| and |fetchHandlerCompleted| are true.
+              1. If |queue| is empty, return null.
```
yoshisatoyanagisawa, line 3255: Is there a way to reuse raceResponse if possible?
The reason of making [=Create Fetch Event and Dispatch=] to take [=raceResponse=] is that we wanted to avoid sending a duplicated request to the server.  If we follow that thought, it might be reasonable to reuse network error on both complete but queue empty case.

docs/index.bs
sisidovski: Thanks. RouterSourceDict can be a RouterSource so we need a dedicated field for the cache name of race-network-and-cache. That is reasonable.

docs/index.bs
monica-ch: Updated, does it looks better?

monica-ch:
@yoshisatoyanagisawa Can you please review the recent changes - removed networkFetchCompleted and fetchHandlerCompleted that was introduced in the PR earlier and incorporated timingInfo per the recent changes in the cache/race-network-and-cache algorithm.

yoshisatoyanagisawa:
Thanks for the contribution.  Overall it looks good except the raceNetworkAndCacheCacheName source support.
To reduce the algorithm duplication, I hope cache lookup algorithms are factored out.

yoshisatoyanagisawa:
@youennf @asutherland FYI
We aim to add the race-network-and-cache source to address the slow storage access.  Please let us know your thoughts.

Thanks @monica-ch for addressing the comments, the PR looks good to me.

monica-ch:
@youennf @asutherland What do you think of this change?

youennf:
I haven't looked at the PR, a few thoughts/questions:
1. How is a web application knowledgeable enough to properly choose cache vs. race-cache-and-network? 
2. A UA can implement race-network-and-cache by using the existing cache route implementation. This approach might actually be useful in some conditions (metered connection e.g.). AFAIUI, the web application will have no way of identifying the UA did an actual race or not.
3. A UA can already start the network load without waiting for the cache. This is a partial race (it cannot use the network response before knowing whether there is a cache entry), but can it solve the perf issue? Should we consider allowing the cache route to be fully raceable with network by default?
4. Are there privacy concerns? For instance, can a web application infer, from the selected route, new information (slow hard drive vs. networking)? Does this reveal new HW/config information? Could this be used by two pages running on the device as a side-information channel (one using intensively hard drive for instance, and the other one checking whether network or cache is used)?

yoshisatoyanagisawa:
Thank you for sharing your thoughts/questions @youennf 
Let me answer one by one.

I think the Resource Timing API should also be updated to help web developers to use the API properly.

> How is a web application knowledgeable enough to properly choose cache vs. race-cache-and-network?

If [the elapsed time](https://github.com/WICG/service-worker-static-routing-api/blob/main/resource-timing-api.md#cache-rule-specified) between `cacheLookupStart` and `responseStart` is large, the cache storage access can be slow.  The `race-network-and-cache` can be chosen for the case.

> A UA can implement race-network-and-cache by using the existing cache route implementation. This approach might actually be useful in some conditions (metered connection e.g.). AFAIUI, the web application will have no way of identifying the UA did an actual race or not.

I think this can be guessed from the resource timing API.  Especially seeing the [workerFinalRouterSource](https://www.w3.org/TR/resource-timing/#dom-performanceresourcetiming-workerfinalroutersource) ratio.

> A UA can already start the network load without waiting for the cache. This is a partial race (it cannot use the network response before knowing whether there is a cache entry), but can it solve the perf issue? Should we consider allowing the cache route to be fully raceable with network by default?

Since [ServiceWorkerAutoPreload](https://github.com/WICG/service-worker-auto-preload) has been around for a while, there may not actually be much resistance to racing, but I think the current cache behavior will leave web developers with an explicit choice to prioritize the cache. i.e. no network access if the cache is available.

> Are there privacy concerns? For instance, can a web application infer, from the selected route, new information (slow hard drive vs. networking)? Does this reveal new HW/config information? Could this be used by two pages running on the device as a side-information channel (one using intensively hard drive for instance, and the other one checking whether network or cache is used)?

I think this is the equivalent attack surface with the fetch handler.  The service worker static routing API is offloading of the traditional fetch handler.  We keep this offloading principle as much as possible.  With the race-network-and-cache, the web application can guess the user's device situation like disk access performance and network access performance.  However, the same thing can be done with the fetch handler.

youennf:
> If [the elapsed time](https://github.com/WICG/service-worker-static-routing-api/blob/main/resource-timing-api.md#cache-rule-specified) between `cacheLookupStart` and `responseStart` is large, the cache storage access can be slow. The `race-network-and-cache` can be chosen for the case.

If decision is made this way, that would require reinstalling the service worker whenever a website wants to opt-in/out of racing. This seems inadequate as the usefulness of racing may depend on varying network conditions.

I would have expected that racing could be enabled/disabled without the need to reinstall the service worker.
Similarly, is there a use case for having different race strategies for different cache routes of a single service worker?

> I think this can be guessed from the resource timing API. Especially seeing the [workerFinalRouterSource](https://www.w3.org/TR/resource-timing/#dom-performanceresourcetiming-workerfinalroutersource) ratio.

It can be guessed but not known for sure, especially if the UA is disabling racing in some specific cases (5G metered vs. wifi for instance).

> However, the same thing can be done with the fetch handler.

Not with the same precision. A fetch handler adds JS, different IPC messages and so on.

yoshisatoyanagisawa:
Thanks for the follow-up. I feel I might not have addressed your points clearly in my previous response, so let me clarify my perspective on each.

**Regarding the opt-in/out of racing and SW reinstallation**: In my view, the decision to use racing is less about varying network conditions and more about whether the server-side is willing to accept the additional load. The motivation of the feature is that developers opt for `race-network-and-cache` because disk/storage access can be slow, rather than simply because the network is fast. However, providing the option not to race is important for those who need to prioritize minimizing server load.

**Regarding different strategies for different routes**: A potential use case for route-specific strategies would be a "niche" optimization where a developer races resources critical to LCP (Largest Contentful Paint) to ensure they arrive as fast as possible, while relying solely on the cache for other resources to reduce server pressure—even if those resources take slightly longer to load.

**Regarding the UA disabling racing (e.g., 5G metered vs. Wi-Fi)**: I understand your point now. I was previously thinking about how developers use the Resource Timing API to send data to the server for research and statistics. From that perspective, it's discernible on the server side. However, if you are referring to the web app knowing this on the client side, I’m not sure what behavior the app would change based on that information. Beyond server-side statistics, what specific use cases do you have in mind for the client-side web app?

**Regarding precision**: While I agree that bypassing the fetch handler reduces "noise" from JavaScript and IPC, I’m not sure how impactful that is compared to the much larger noise inherent in network or HDD access. In terms of measurement precision, both approaches use the same underlying timers, so the results should be comparable. If you believe the precision difference is significant, I would appreciate more details on that.

asutherland:
This seems like a good change; we always wanted this capability.  I do like that this provides UAs the ability to potentially bias towards always going to cache based on user preference but also having the latitude to use the network for latency reasons.

One question I think this capability inherently raises is "If we are fetching something from the network and it's not in the cache, does it get put in the cache? Or can the cache entry be updated with the network result?".  The current answer is  "no" and the existing [understanding cache lifetimes](https://w3c.github.io/ServiceWorker/#cache-lifetimes) prose already incidentally answers the question, but it could be nice to add some non-normative text around [InstallEvent](https://w3c.github.io/ServiceWorker/#installevent-interface) to maybe reiterate that point.

Refreshing my memory from https://github.com/WICG/service-worker-static-routing-api/blob/main/final-form.md I do see that there is an updatedCacheName mechanism for stale-while-revalidate and it does seem desirable that any cache updating would be quite explicit in that way.  But since we also don't really have any automatic eviction mechanism, leaving it to the SW itself to perform any content updates does seem appropriate rather than risking making it easy to create an ever-growing cache.

monica-ch:
> since we also don't really have any automatic eviction mechanism, leaving it to the SW itself to perform any content updates does seem appropriate rather than risking making it easy to create an ever-growing cache.

added a Note under InstallEvent.addRoute clarifying that neither "cache" nor "race-network-and-cache" populates the Cache, the network-wins case just delivers the response with no cache write. PTAL.

PR #1767, opened 2025-05-01
Correct ServiceWorkerRegistration Web IDL

yoshisatoyanagisawa:
As mentioned in https://github.com/w3c/ServiceWorker/issues/1472, https://w3c.github.io/ServiceWorker/#resolve-job-promise-algorithm is written to return Promise<ServiceWorkerRegistration>. Also, according to https://wpt.fyi/results/service-workers/service-worker/update-result.https.html?label=experimental&label=master&aligned, all implementations are actually returning a `Promise<ServiceWorkerRegistration>`.

***
<a href="https://pr-preview.s3.amazonaws.com/yoshisatoyanagisawa/ServiceWorker/pull/1767.html" title="Last updated on May 1, 2025, 2:24 AM UTC (6495b0f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1767/479b8ae...yoshisatoyanagisawa:6495b0f.html" title="Last updated on May 1, 2025, 2:24 AM UTC (6495b0f)">Diff</a>

Issue #1472, opened 2019-09-16
The return value for update() is wrong in WebIDL


philipwalton:
The [current WebIDL for the `ServiceWorkerRegistration`](https://w3c.github.io/ServiceWorker/#serviceworkerregistration-interface) is incorrect; it shows `update()` returning `Promise<void>`.

<img width="857" alt="Screen Shot 2019-09-16 at 4 18 56 PM" src="https://user-images.githubusercontent.com/326742/64940018-d9f9c280-d89d-11e9-98d7-37703a8a5b79.png">

I believe this is a mistake since implementations all return the registration object, and the [`update()` algorithm](https://w3c.github.io/ServiceWorker/#dom-serviceworkerregistration-update) says to return the registration as well.

yoshisatoyanagisawa:
Let me check what is written in WebIDLs.
Chromium: 
> `[CallWith=ScriptState, RaisesException] Promise<ServiceWorkerRegistration> update();`

(https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/service_worker/service_worker_registration.idl;l=26;drc=047c7dc4ee1ce908d7fea38ca063fa2f80f92c77)

Gecko:
>  ` [Throws, NewObject]`
>  `Promise<undefined> update();`

(https://github.com/mozilla/gecko-dev/blob/fefa0c86cf24e6afa9f4dbb3823d4d16b2713dc7/dom/webidl/ServiceWorkerRegistration.webidl#L28)

WebKit:
>  `[NewObject] Promise<undefined> update();`

(https://github.com/WebKit/WebKit/blob/f8eb4cf035d529c6e55ebfd28c92fdc9468981d5/Source/WebCore/workers/service/ServiceWorkerRegistration.idl#L46)

As @saschanaz pointed out, Gecko and WebKit returns `Promise<undefined>`.
However, although the WebIDL looks invalid, Chromium seems to follow the step by step description in [Resolve Job Promise algorithm](https://w3c.github.io/ServiceWorker/#resolve-job-promise-algorithm).

If mismatch of WebIDL and [Resolve Job Promise algorithm](https://w3c.github.io/ServiceWorker/#resolve-job-promise-algorithm) matters, adjusting the WebIDL can be a choice.  What do you think?

saschanaz:
Depends on whether there's enough implementation interest for that, otherwise I'd vote for following the majority. cc @asutherland

annevk:
[...]

PR #1771, opened 2025-05-20
Update index.bs

preitinger:
Corrected return type of ServiceWorkerRegistration.update()

***
<a href="https://pr-preview.s3.amazonaws.com/preitinger/ServiceWorker/pull/1771.html" title="Last updated on May 20, 2025, 1:25 PM UTC (da4c389)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1771/b7bbcea...preitinger:da4c389.html" title="Last updated on May 20, 2025, 1:25 PM UTC (da4c389)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -556,7 +556,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
         readonly attribute USVString scope;
         readonly attribute ServiceWorkerUpdateViaCache updateViaCache;
 
-        [NewObject] Promise&lt;ServiceWorkerRegistration&gt; update();
+        [NewObject] Promise&lt;void&gt; update();
         [NewObject] Promise&lt;boolean&gt; unregister();
 
         // event


yoshisatoyanagisawa:
Will you elaborate more on why you think `void` is correct?
It became `ServiceWorkerRegistration` in https://github.com/w3c/ServiceWorker/pull/1767.  The change is based on the discussion in https://github.com/w3c/ServiceWorker/issues/1472.  Chromium, Mozilla, and WebKit agreed to use this.

preitinger:
> Will you elaborate more on why you think `void` is correct? It became `ServiceWorkerRegistration` in #1767. The change is based on the discussion in #1472. Chromium, Mozilla, and WebKit agreed to use this.

First of all, I noticed that the typescript signature of `ServiceWorkerRegistration.update()` told me that it returns `Promise<void>`.
Originally, I created an issue in the MDN web docs which I use for my daily work. There, they told me that this spec says that `ServiceWorkerRegistration.update()` should return `Promise<ServiceWorkerRegistration>`. Now, considering the lifecycle of `ServiceWorkerRegistration`, I wonder what it should return, actually. (The detail description for update() says btw nothing about the return value, just that an activity is started and when it is done the promise returns. Not that with a value. Only in fail cases it describes which exception is thrown. And for me, just as an interested newbie user (switched to typescript 2 years ago) and not used to this specification process, it seems to be very logical that `update()` just fulfills its `Promise<void>` when the `update` is done. And then, the user knows that he will find the new service worker in `ServiceWorkerRegistration.installing` (or if some strange super-quick happens) already in `ServiceWorkerRegistration.waiting` of the `ServiceWorkerRegistration` object which he owns anyway because he called `update()` on it. In my opinion, if `update` should return a value at all except for the fact that it has completed which means `void`, then not an object of `ServiceWorkerRegistration`, but a `ServiceWorker` which must then be probably the same as referenced by `ServiceWorkerRegistration.installing` (or if that is possible at all that the installation is so fast) in `ServiceWorkerRegistration.waiting`. It makes no sense at all imo to return a `ServiceWorkerRegistration` except for itself for some function chaining or the like.

yoshisatoyanagisawa:
Please see https://github.com/w3c/ServiceWorker/issues/1472#issuecomment-2897117405 why it is `promise<ServiceWorkerRegistration>`.  By the way, if you think it is very logical that `update()` just fulfills its `Promise<void>` when the update is done, I want you to get consensus in the github issue I linked.  Also, I want the algorithm consistent with the IDL change.

preitinger:
OK, I am not used to the process here, but I think I have understood the following facts:

- The typescript definition (at least in VS Code which I am using) wrongly tells that Promise<void> is returned by `update()`.
- Almost all browsers return currently the same object on which `update()` is called.
- The algorithms that are triggered by `update()` are consistently defined to return `Promise<ServiceWorkerRegistration>`.
- The effort for changing the return value to `Promise<void>` would be huge in several areas.
- Millions of JavaScript projects (not Typescript) might break if the return value would be changed.
- Users who are confused at first glance, as me, can just ignore the return value.

Have I understood it halfways? ;-)

Thank you for your explanations! Have I to do anything like withdraw my pull request?

yoshisatoyanagisawa:
Yes, your understanding is correct.
Let me close the PR without merging.

PR #1774, opened 2025-05-30
Fix: Ensure timingInfo is returned on network fallback if SW static routing API is used

yoshisatoyanagisawa:
According to the [explainer](https://github.com/WICG/service-worker-static-routing-api/blob/main/resource-timing-api.md#correspondence-of-the-matched-source-type-and-the-actual-source-type), the `worker final router source` should be `'network'` for a network fallback.  The |timingInfo| is returned to make that happen.


yoshisatoyanagisawa:
@quasi-mod I also found that "network" should be set if "fetch-event" is a matched router source, and fallback according to the explainer.  Let me revise the algorithm like so.

PR #1775, opened 2025-06-03
Add new environment settings object field "cross site ancestry" for SameSite cookies work

bvandersloot-mozilla:
To define how cookies work correctly, particularly around the SameSite attribute, we need to know whether or not an environment settings object has any cross-site ancestors. This is making sure the ESO definition in ServiceWorkers also defines this member algorithm. 

We will have to land [the PR that adds this to HTML](https://github.com/whatwg/html/pull/11133) before landing this, but it'd be nice to get this approved before we alter HTML so everything goes smoothly.

See https://github.com/whatwg/html/issues/9000 for even broader context.

***
<a href="https://pr-preview.s3.amazonaws.com/bvandersloot-mozilla/ServiceWorker/pull/1775.html" title="Last updated on Jul 29, 2025, 6:09 PM UTC (433306a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1775/ef029c5...bvandersloot-mozilla:433306a.html" title="Last updated on Jul 29, 2025, 6:09 PM UTC (433306a)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3065,8 +3065,8 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               :: Return |serviceWorker|'s [=service worker/script url=].
               : The [=environment settings object/origin=]
               :: Return its registering [=/service worker client=]'s [=environment settings object/origin=].
-              : The [=environment settings object/cross site ancestry=]
-              :: Return its registering [=/service worker client=]'s [=environment settings object/cross site ancestry=].
+              : The [=environment settings object/has cross-site ancestor=]
+              :: Return its registering [=/service worker client=]'s [=environment settings object/has cross-site ancestor=].
               : The [=environment settings object/policy container=]
               :: Return |workerGlobalScope|'s [=WorkerGlobalScope/policy container=].
               : The [=environment settings object/time origin=]


docs/index.bs
```diff
@@ -3065,6 +3065,8 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               :: Return |serviceWorker|'s [=service worker/script url=].
               : The [=environment settings object/origin=]
               :: Return its registering [=/service worker client=]'s [=environment settings object/origin=].
+              : The [=environment settings object/cross site ancestry=]
+              :: Return its registering [=/service worker client=]'s [=environment settings object/cross site ancestry=].
```
yoshisatoyanagisawa, line 3069: I assume that `cross site ancestry` is used for noticing 3rd party site ancestry to avoid unexpected cookie access there.  Then, I wonder if it is valid to use the registration time client for decision making.
Assuming that the registration has been done by Origin A, but actual usage is fetching from ServiceWorker-controlled iframe Origin A inside Origin B.  For that case, I guess fetching inside iframe Origin A can be captured by ServiceWorker registered by the Origin A (not iframe's) if the storage partitioning is disabled.  Then, the iframe itself might be `cross site ancestry` == true but the ServiceWorker intercepting the request might be `cross site ancestry` == false.  Is it intended behavior?

bvandersloot-mozilla: Okay, that makes sense! Thank you for clearing that up. 

I think this actually works fine in the case that storage partitioning is enabled because the cross-site ancestry will be the same between the frame initiating the fetch and the worker handling it.

Currently there are some plans to specify all of the storage partitioning implementation changes, particularly from @kyraseevers. 

The spec works as-is because of some hand-waving in Fetch in how `SameSite` is handled. Would it be reasonable to add an inline issue somewhere in this spec pointing out that if storage partitioning is disabled, the invocation of fetch should be done with a client with `cross site ancestry` matching the initiating client?

yoshisatoyanagisawa: Thanks for the reply.

Could you please add a note mentioning that `cross-site ancestry` may not be set correctly if Storage Partitioning is disabled? I think we should strongly recommend enabling it for any algorithm that relies on `cross-site ancestry`.

Other than that, looks good to me.

bvandersloot-mozilla: I've added a note in the list. Broken references should resolve once we merge https://github.com/whatwg/html/pull/11133.

Thank you for the feedback!

yoshisatoyanagisawa (approved):

docs/index.bs
```diff
@@ -3065,6 +3065,11 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+                <div class="note">
+                  Note: If service workers are not partitioned by the [=environment settings object/cross site ancestry=], clients must include logic to use the initial  [=/http fetch=]'s [=/request=]'s [=request/client=]'s [=environment settings object/cross site ancestry=] when determining the "<code>SameSite</code>" mode.
```
annevk, line 3071: We cannot have requirements in notes.

annevk (commented):
For some reason I thought service workers had to be more complicated, but I suspect that's mainly when Storage Access comes into play.

docs/index.bs
annevk: I'm not sure it makes a whole lot of sense to have the note? It's essentially saying you have to do something else if you don't implement the specification, which is, hmm, weird?

docs/index.bs
```diff
@@ -3065,6 +3065,11 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               :: Return |serviceWorker|'s [=service worker/script url=].
               : The [=environment settings object/origin=]
               :: Return its registering [=/service worker client=]'s [=environment settings object/origin=].
+              : The [=environment settings object/cross site ancestry=]
```
annevk, line 3068: This will have to be renamed now.

docs/index.bs
yoshisatoyanagisawa: Hi @annevk, Thank you for the review and sorry for the delay in my response.

You're correct, and I agree with removing this note. It's not appropriate to define requirements within a spec note.

Let me clarify the uncertainty that led me to suggest it in the first place. My core question is: Is storage partitioning considered a normative requirement for a compliant ServiceWorker implementation?

If all compliant user agents must implement storage partitioning, then as you said, we don't need to consider the non-partitioned case, and this note is simply noise.

However, if partitioning is not a hard requirement, I am concerned about the scenario where a non-partitioned ServiceWorker intercepts a fetch from a cross-site iframe. It might incorrectly determine the `cross-site ancestry` to be `false`, leading to an unintended relaxation of `SameSite` cookie protections. The note was a flawed attempt to bridge that potential gap.

Given your expertise on the platform, I'd appreciate your guidance here. Should the ServiceWorker spec be written with the assumption that storage partitioning is present? Or should we add normative text (not a note) to ensure the `cross-site ancestry` is handled safely even in a non-partitioned context?

bvandersloot-mozilla: Given this conclusion, I'm happy to close without merging.

annevk: I don't understand. Don't we still need to forward the field?

yoshisatoyanagisawa: I've reopened this pull request, and I apologize if my earlier comments led to the misunderstanding and its closure.

To clarify, my intention was only to discuss the necessity of the note I had previously requested. I agree with @annevk's final assessment that we should still forward the field, and the conclusion of our discussion was that only the note should be removed, not the entire change in this PR.

The original purpose of this PR—adding the field for cross-site ancestry—is still necessary, though it may require renaming as @annevk pointed out separately.

bvandersloot-mozilla: Ah, I missed that part of the patch- sorry for closing prematurely. I've renamed it and reverted the note.

yoshisatoyanagisawa:
https://github.com/whatwg/html/pull/11133 looks merged.
I am also going to merge this.

docs/index.bs
yoshisatoyanagisawa: lgtm.

Issue #1788, opened 2025-09-04
Revise cache storage lookup using the "obtain a local storage bottle map" algorithm


yoshisatoyanagisawa:
In https://github.com/w3c/ServiceWorker/issues/1784, @annevk pointed out that "A storage key doesn't hold a name to cache map."
However, In https://w3c.github.io/ServiceWorker/#cache-constructs, [name to cache map](https://w3c.github.io/ServiceWorker/#dfn-name-to-cache-map) is associated with the [storage key](https://storage.spec.whatwg.org/#storage-key).

The part should be updated to use the storage standard.

yoshisatoyanagisawa:
It has been addressed in https://github.com/w3c/ServiceWorker/pull/1789.
Let me close this.

Issue #1784, opened 2025-07-30
What does "the registration’s storage key’s name to cache map" mean?


annevk:
A storage key doesn't hold a name to cache map.

yoshisatoyanagisawa:
Thanks for pointing this out.

The phrase "the registration’s storage key’s name to cache map" is based on the current definitions in the specification.

[Section 5.1 of the specification](https://w3c.github.io/ServiceWorker/#ref-for-dfn-name-to-cache-map) states, "Each storage key has an associated name to cache map". Since a `ServiceWorkerRegistration` has an associated storage key, the wording resulted in the current chain of possessives.

That said, I agree that this phrasing can be a bit confusing. I'm open to suggestions for a clearer alternative. For instance, we could rephrase it to "the name to cache map associated with the registration's storage key".

What do you think?

annevk:
I don't think it makes sense for the abstract concept of a storage key to have an associated name to cache map. It makes a bit more sense if we're talking about instances of storage keys, but that would really be clarified by a map of its own.

Perhaps

> (storage key, name) -> cache

or some such?

mkruisselbrink:
Ideally I think this should all defer to concepts in the whatwg storage spec? I.e. get a storage bottle map for "caches" and look up the named cache in there?

yoshisatoyanagisawa:
I follow @mkruisselbrink's idea.

To implement this, what if we add a new algorithm to the Storage spec like "To obtain a local storage bottle map given a storage key"? It could live in section 4.6 "Storage bottles".

The ServiceWorker specification could then call this new algorithm with the registration's storage key and the identifier "caches". This would cleanly replace the current "name to cache map" concept.

Does this sound like a good approach?

annevk:
We don't have an environment we can use? If not, I guess we indeed need to refactor Storage so yo can pass a storage key directly.

yoshisatoyanagisawa:
> We don't have an environment we can use? 

That is my understanding.

> If not, I guess we indeed need to refactor Storage so yo can pass a storage key directly.

sure.

wanderview:
Can you not use request's reserved client as the environment?

https://fetch.spec.whatwg.org/#concept-request-reserved-client

yoshisatoyanagisawa:
Thank you for your suggestion.

Using the request's reserved client as the environment would indeed work for the fetch event. However, as you know, the Cache API is also used in other contexts, such as the install event, where there is no associated request or reserved client.

[...]

PR #1789, opened 2025-09-04
Fix: Align Cache API definitions with Storage spec

yoshisatoyanagisawa:
This commit clarifies how the Cache API interacts with the underlying storage mechanism, resolving ambiguity that stemmed from previous specification language.

Key changes:
- Removes the direct association between "storage key" and "name to cache map".
- Defines "relevant name to cache map" as being retrieved via the `obtain a local storage bottle map` algorithm from the Storage specification, ensuring alignment.
- Updates the definition of `CacheStorage` to represent the `name to cache map` associated with its `environment settings object`, rather than the `origin`.

This makes the relationship between the ServiceWorker Cache API and the Storage specification explicit and consistent, addressing the core issue raised in https://github.com/w3c/ServiceWorker/issues/1784 and the follow-up work in https://github.com/w3c/ServiceWorker/issues/1788.

***
<a href="https://pr-preview.s3.amazonaws.com/yoshisatoyanagisawa/ServiceWorker/pull/1789.html" title="Last updated on Sep 9, 2025, 2:24 AM UTC (45b3847)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1789/3ab2b8a...yoshisatoyanagisawa:45b3847.html" title="Last updated on Sep 9, 2025, 2:24 AM UTC (45b3847)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -1952,9 +1952,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     A <dfn id="dfn-name-to-cache-map">name to cache map</dfn> is an <a>ordered map</a> whose [=map/entry=] consists of a [=map/key=] (a string that represents the name of a [=request response list=]) and a [=map/value=] (a [=request response list=]).
 
-    Each [=/storage key=] has an associated [=name to cache map=].
-
-    The <dfn id="dfn-relevant-name-to-cache-map">relevant name to cache map</dfn> is the [=name to cache map=] associated with the [=/storage key=] [=obtain a storage key|obtained=] from [=this=]'s associated [=CacheStorage/global object=]'s [=environment settings object=].
+    The <dfn id="dfn-relevant-name-to-cache-map">relevant name to cache map</dfn> is a [=name to cache map=] obtained as the result of running [=obtain a local storage bottle map=] with [=this=]'s associated [=CacheStorage/global object=]'s [=environment settings object=] and "<code>caches</code>".
   </section>
 
   <section>
@@ -2262,7 +2260,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     The user agent *must* create a {{CacheStorage}} object when a {{Window}} object or a {{WorkerGlobalScope}} object is created and associate it with that <dfn for="CacheStorage">global object</dfn>.
 
-    A {{CacheStorage}} object represents a <a>name to cache map</a> of its associated [=CacheStorage/global object=]'s <a>environment settings object</a>'s [=environment settings object/origin=]. Multiple separate objects implementing the {{CacheStorage}} interface across documents and workers can all be associated with the same <a>name to cache map</a> simultaneously.
+    A {{CacheStorage}} object represents the [=name to cache map=] associated with its associated [=CacheStorage/global object=]'s [=environment settings object=]. Multiple separate objects implementing the {{CacheStorage}} interface across documents and workers can all be associated with the same <a>name to cache map</a> simultaneously.
 
     <section algorithm="cache-storage-match">
       <h4 id="cache-storage-match">{{CacheStorage/match(request, options)}}</h4>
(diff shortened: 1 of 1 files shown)


yoshisatoyanagisawa:
@annevk @mkruisselbrink @wanderview FYI.
@domenic @monica-ch Will you review this?

wanderview:
Looks reasonable to me, but my spec knowledge is a bit rusty.

docs/index.bs
```diff
@@ -2262,7 +2260,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     The user agent *must* create a {{CacheStorage}} object when a {{Window}} object or a {{WorkerGlobalScope}} object is created and associate it with that <dfn for="CacheStorage">global object</dfn>.
 
-    A {{CacheStorage}} object represents a <a>name to cache map</a> of its associated [=CacheStorage/global object=]'s <a>environment settings object</a>'s [=environment settings object/origin=]. Multiple separate objects implementing the {{CacheStorage}} interface across documents and workers can all be associated with the same <a>name to cache map</a> simultaneously.
+    A {{CacheStorage}} object represents the [=name to cache map=] associated with its associated [=CacheStorage/global object=]'s [=environment settings object=]. Multiple separate objects implementing the {{CacheStorage}} interface across documents and workers can all be associated with the same <a>name to cache map</a> simultaneously.
```
monica-ch, line 2263: using def ref syntax for <a>name to cache map</a> to [=name to cache map=] and also maintains consistency

docs/index.bs
```diff
@@ -1952,9 +1952,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     A <dfn id="dfn-name-to-cache-map">name to cache map</dfn> is an <a>ordered map</a> whose [=map/entry=] consists of a [=map/key=] (a string that represents the name of a [=request response list=]) and a [=map/value=] (a [=request response list=]).
 
-    Each [=/storage key=] has an associated [=name to cache map=].
-
-    The <dfn id="dfn-relevant-name-to-cache-map">relevant name to cache map</dfn> is the [=name to cache map=] associated with the [=/storage key=] [=obtain a storage key|obtained=] from [=this=]'s associated [=CacheStorage/global object=]'s [=environment settings object=].
+    The <dfn id="dfn-relevant-name-to-cache-map">relevant name to cache map</dfn> is a [=name to cache map=] obtained as the result of running [=obtain a local storage bottle map=] with [=this=]'s associated [=CacheStorage/global object=]'s [=environment settings object=] and "<code>caches</code>".
```
monica-ch, line 1955: nit: Isn't below better describes indirect relationship which we want to establish

The <dfn id="dfn-relevant-name-to-cache-map">relevant name to cache map</dfn> is the [=name to cache map=] associated with the result of running [=obtain a local storage bottle map=] with [=this=]'s associated [=CacheStorage/global object=]'s [=environment settings object=] and "<code>caches</code>".

docs/index.bs
monica-ch, line 2263: and also, to maintain consistency here? 

A {{CacheStorage}} object represents the [=name to cache map=] associated with the result of running [=obtain a local storage bottle map=] with its associated [=CacheStorage/global object=]'s [=environment settings object=]....

What do you think?

docs/index.bs
domenic, line 1955: Nit: it should not be necessary to give your `<dfn>`s `id=""` values. (The main use is if you are changing the definition text and want to preserve any links to the old definition.)

Another nit (and it is a preexisting issue): `[=this=]` can only be used inside Web IDL algorithms. It looks like "relevant name to cache map" is only used inside such algorithms, but it is still confusing to see `[=this=]` used indirectly in such a way. The more traditional thing to do would be to say "the **relevant name to cache map** of a `CacheStorage` _obj_ is... _obj_'s associated..."

domenic (commented):
This looks substantively right to me.

docs/index.bs

docs/index.bs
yoshisatoyanagisawa: sure. updated. also mixed with domenic's comment.

PR #1790, opened 2025-09-05
Handle Fetch: Use correct storage key for subresource requests

yoshisatoyanagisawa:
This change corrects the cache resolution logic within the Handle Fetch algorithm when the router source is RouterSourceEnum/"cache".

The Problem:
Previously, the algorithm always used reservedClient to resolve the storage key, regardless of the request type. However, reservedClient is only valid for top-level navigation requests and is not available for subresource requests. This could lead to accessing the incorrect cache storage during subresource fetches.

The Fix:
Logic has been added to differentiate between a non-subresource request and a subresource request.
- For a non-subresource request: It continues to use reservedClient as before.
- For a subresource request: It now uses the client associated with the request object.

This ensures that the correct storage key, based on the requesting client, is used for subresource requests.

Fixes: https://github.com/w3c/ServiceWorker/issues/1784

***
<a href="https://pr-preview.s3.amazonaws.com/yoshisatoyanagisawa/ServiceWorker/pull/1790.html" title="Last updated on Sep 5, 2025, 2:09 AM UTC (d0832ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1790/953b6d3...yoshisatoyanagisawa:d0832ba.html" title="Last updated on Sep 5, 2025, 2:09 AM UTC (d0832ba)">Diff</a>

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -3241,7 +3241,12 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
               1. Else if |source| is {{RouterSourceEnum/"cache"}}, or |source|["{{RouterSourceDict/cacheName}}"] [=map/exists=], then:
                   1. If |shouldSoftUpdate| is true, then [=in parallel=] run the [=Soft Update=] algorithm with |registration|.
                   1. Set |timingInfo|’s [=service worker timing info/worker cache lookup start=] to the [=coarsened shared current time=] given |useHighResPerformanceTimers|.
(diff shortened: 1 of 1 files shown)


yoshisatoyanagisawa:
@wanderview Will you take a look?  This should address your concern.
@monica-ch Will you review this?
@annevk @mkruisselbrink FYI.

monica-ch:
LGTM, thanks for fixing!

wanderview:
LGTM

PR #1818, opened 2026-03-06
Remove redundant origin check in unregister algorithm

### LINKED ISSUE #1305: how can unregister()'s same-origin check ever fail?
Currently the first step of unregister is:

```
1. If the origin of job’s scope url is not job’s client's origin, then:
  1. Invoke Reject Job Promise with job and "SecurityError" DOMException.
  2. Invoke Finish Job with job and abort these steps.
```

In this case the scope URL is the scope from the ServiceWorkerRegistration.  For this to fail a ServiceWorkerRegistration would have to be exposed in a cross-origin client context.

That doesn't seem possible... And if it is, there are probably bigger problems we need to fix.

Can we remove these steps from the spec?

yoshisatoyanagisawa:
This removes the first step of the Unregister algorithm, which threw a "SecurityError" DOMException if the origin of the job's scope URL did not match the job's client's origin.

This check is redundant because the scope URL is derived directly from the ServiceWorkerRegistration object itself. For this check to fail, the ServiceWorkerRegistration would have had to be exposed in a cross-origin client context, which is already prevented by the platform.

Fixes #1305

***
<a href="https://pr-preview.s3.amazonaws.com/yoshisatoyanagisawa/ServiceWorker/pull/1818.html" title="Last updated on Mar 6, 2026, 9:07 AM UTC (93d977e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1818/5614d6f...yoshisatoyanagisawa:93d977e.html" title="Last updated on Mar 6, 2026, 9:07 AM UTC (93d977e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3723,9 +3723,6 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       : Output
       :: none
 
-      1. If the [=environment settings object/origin=] of |job|'s [=job/scope url=] is not |job|'s [=job/client=]'s [=environment settings object/origin=], then:
-          1. Invoke [=Reject Job Promise=] with |job| and "{{SecurityError}}" {{DOMException}}.
-          1. Invoke <a>Finish Job</a> with |job| and abort these steps.
       1. Let |registration| be the result of running <a>Get Registration</a> given |job|'s [=job/storage key=] and |job|'s [=job/scope url=].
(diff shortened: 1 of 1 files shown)


yoshisatoyanagisawa:
@monica-ch Will you review this?

PR #1821, opened 2026-03-10
Teach ExtendableMessageEvent to work with origins.

### LINKED ISSUE #1820: Origin API: need to change ExtendableMessageEvent
The change we made to MessageEvent per discussion in https://github.com/whatwg/html/issues/11993 also needs to be made for ExtendableMessageEvent.

cc @mikewest

mikewest:
Currently, ExtendableMessageEvent's origin holds a serialized origin. In order for Origin.from(...) to function as intended for sources with opaque origins, we need to hold an origin on ExtendableMessageEvent instead.

This patch adds an "origin" concept to `ExtendableMessageEvent`, and shifts call sites to initialize that value with an origin whenever possible. It also adds "extract an origin" steps to return the stored origin when present, and to return null in cases where the object was initialized with an untrusted origin representation (e.g., when its constructor is called with a ExtendableMessageEventInit dictionary).

This replicates whatwg/html#11993's work on `MessageEvent`, and should fix w3c/ServiceWorker#1820.

***
<a href="https://pr-preview.s3.amazonaws.com/mikewest/ServiceWorker/pull/1821.html" title="Last updated on Mar 11, 2026, 9:37 AM UTC (c890bb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1821/4a9e044...mikewest:c890bb1.html" title="Last updated on Mar 11, 2026, 9:37 AM UTC (c890bb1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1847,6 +1847,8 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       <h4 id="extendablemessage-event-lasteventid">{{ExtendableMessageEvent/lastEventId|event.lastEventId}}</h4>
 
       The <dfn attribute for="ExtendableMessageEvent">lastEventId</dfn> attribute *must* return the value it was initialized to. When the object is created, this attribute *must* be initialized to the empty string.
+
+      Objects implementing the {{ExtendableMessageEvent}} interface's [=extract an origin=] steps are to return [=this=]'s [=ExtendableMessageEvent/origin=] if it is an [=/origin=]; otherwise null.
     </section>
 
     <section>
--- a/index.bs
+++ b/index.bs
@@ -100,6 +100,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         urlPrefix: interaction.html
             text: visibilityState; for: Document; url: page-visibility
     type: dfn
+        urlPrefix: comms.html
+            text: concept-MessageEvent-origin; url: concept-messageevent-origin
         urlPrefix: syntax.html
             text: delay the load event; for: document; url: delay-the-load-event
         urlPrefix: browsers.html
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -100,6 +100,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         urlPrefix: interaction.html
             text: visibilityState; for: Document; url: page-visibility
     type: dfn
+        urlPrefix: comms.html
+            text: concept-MessageEvent-origin; url: concept-messageevent-origin
```
annevk: We should probably export this.

annevk (approved):
This looks good to me. A Service Worker editor should probably do final approval.

index.bs
mikewest: I didn't expect anyone outside HTML would need it. Silly me. I'll send you a PR in a minute to add the attribute.

mikewest: Ah. Actually we did export it. I guess it just hasn't landed in the references DB yet. We should be able to drop this ~tomorrow.

yoshisatoyanagisawa:
Thanks for the update. To ensure full parity with the changes made to `MessageEvent` in whatwg/html#12235, I was looking for the "extract an origin" steps for `ExtendableMessageEvent`.

Since `ExtendableMessageEvent` doesn't inherit from `MessageEvent`, we'll need an explicit definition to support the `Origin.from()` API. Should we add a sentence similar to the one in the [HTML spec](https://html.spec.whatwg.org/#the-messageevent-interface:extract-an-origin) here as well?

mikewest:
Thanks for pointing that out. Added it in https://github.com/w3c/ServiceWorker/pull/1821/commits/c890bb1e5b32455f474c1c0d35485bb177ce63c9.

yoshisatoyanagisawa:
lgtm. Thank you!

yoshisatoyanagisawa (approved):

PR #1832, opened 2026-07-13
Editorial: clarify FetchEvent clientId, resultingClientId, and replacesClientId initialization

### LINKED ISSUE #1267: clarify when FetchEvent.clientId will or will not be set for navigations
Over in #1266 the question of when exactly `FetchEvent.clientId` will or will not be populated for navigations came up.

It seems it must not be populated for:

* If the navigation is initiated from a cross-origin Client
* If there is no Client at all, like the user entering a URL in the tab and pressing enter.

We could potentially should a clientId for:

* Same-origin client setting window.location
* Same-origin client setting iframe.src
* Same-origin client calling window.open()
* Same-origin client calling clients.openWindow()

monica-ch:
## Summary

Aligns the [Create Fetch Event and Dispatch](https://w3c.github.io/ServiceWorker/#create-fetch-event-and-dispatch-algorithm) algorithm and the `FetchEvent.clientId` / `resultingClientId` attribute definitions with **actual browser behavior** (Chrome, Firefox, Safari) and MDN.

Fixes #1267.

### 1. `FetchEvent.clientId` algorithm step

Previously the spec unconditionally initialized `clientId` from `request`'s client, including for navigations. However, all major browsers return an empty string for `clientId` on navigation requests. Per [@wanderview's suggestion](https://github.com/w3c/ServiceWorker/issues/1267#issuecomment-362298501):

> My inclination is to probably implement resultingClientId and keep navigation clientId empty at first.

**Before:**
> If |client| is not null, initialize |e|'s clientId attribute to |client|'s environment id.

**After:**
> If |request| is a subresource request and |client| is not null, initialize |e|'s clientId attribute to |client|'s environment id.

### 2. `FetchEvent.resultingClientId` algorithm step

Adds an explicit `reservedClient` null check and removes the redundant "and to the empty string otherwise" clause (the IDL default already provides `""`).

### 3. `FetchEvent.replacesClientId` algorithm step

Removes the redundant "and to the empty string otherwise" clause for consistency with the sibling attributes.

### 4. Explanatory notes

Added descriptive notes for both `FetchEvent.clientId` and `FetchEvent.resultingClientId` attribute sections, documenting all cases where the value is the empty string. These match MDN's documented behavior.

***
<a href="https://pr-preview.s3.amazonaws.com/monica-ch/ServiceWorker/pull/1832.html" title="Last updated on Jul 17, 2026, 3:59 PM UTC (018917e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1832/e91ddff...monica-ch:018917e.html" title="Last updated on Jul 17, 2026, 3:59 PM UTC (018917e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1706,7 +1706,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
       <dfn attribute for="FetchEvent"><code>clientId</code></dfn> attribute *must* return the value it was initialized to. When an <a>event</a> is created the attribute *must* be initialized to the empty string.
 
-      Note: {{FetchEvent/clientId}} is the [=environment/id=] of |request|'s [=request/client=] when the [=/request=] has an associated initiating [=/service worker client=]. This is the case for [=subresource requests=], and also for [=non-subresource requests=] that were initiated by an existing client (such as worker script fetches, or a [=navigation request=] initiated by an existing document). It is the empty string when |request|'s [=request/client=] is null (e.g., a top-level [=navigation request=] triggered by the user typing a URL in the address bar). See {{FetchEvent/resultingClientId}} for the environment associated with the resulting document of a [=navigation request=].
+      Note: {{FetchEvent/clientId}} is the [=environment/id=] of |request|'s [=request/client=] when the [=/request=] has an initiating [=/service worker client=], and the empty string otherwise (e.g., a top-level [=navigation request=] with no initiator). See {{FetchEvent/resultingClientId}} for the environment associated with the resulting document of a [=navigation request=].
     </section>
 
     <section>
--- a/index.bs
+++ b/index.bs
@@ -1706,7 +1706,7 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
       <dfn attribute for="FetchEvent"><code>clientId</code></dfn> attribute *must* return the value it was initialized to. When an <a>event</a> is created the attribute *must* be initialized to the empty string.
 
-      Note: {{FetchEvent/clientId}} is the [=environment/id=] of |request|'s [=request/client=] when the [=/request=] has an associated initiating [=/service worker client=]. This is the case for [=subresource requests=], and also for [=non-subresource requests=] that were initiated by an existing client (such as [=worker=] script fetches, or a [=navigation request=] initiated by an existing document). It is the empty string when |request|'s [=request/client=] is null (e.g., a top-level [=navigation request=] triggered by the user typing a URL in the address bar). See {{FetchEvent/resultingClientId}} for the environment associated with the resulting document of a [=navigation request=].
+      Note: {{FetchEvent/clientId}} is the [=environment/id=] of |request|'s [=request/client=] when the [=/request=] has an associated initiating [=/service worker client=]. This is the case for [=subresource requests=], and also for [=non-subresource requests=] that were initiated by an existing client (such as worker script fetches, or a [=navigation request=] initiated by an existing document). It is the empty string when |request|'s [=request/client=] is null (e.g., a top-level [=navigation request=] triggered by the user typing a URL in the address bar). See {{FetchEvent/resultingClientId}} for the environment associated with the resulting document of a [=navigation request=].
     </section>
 
     <section>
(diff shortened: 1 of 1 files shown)


monica-ch:
@yoshisatoyanagisawa Could you review these changes and let me know if this is the right approach for updating the spec?

index.bs
```diff
@@ -1705,12 +1705,16 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       <h4 id="fetch-event-clientid">{{FetchEvent/clientId|event.clientId}}</h4>
 
       <dfn attribute for="FetchEvent"><code>clientId</code></dfn> attribute *must* return the value it was initialized to. When an <a>event</a> is created the attribute *must* be initialized to the empty string.
+
+      Note: {{FetchEvent/clientId}} is |request|'s [=request/client=]'s [=environment/id=] for [=subresource requests=], and the empty string otherwise (including for [=non-subresource requests=] such as [=navigation requests=], or when |request|'s [=request/client=] is null). See {{FetchEvent/resultingClientId}} for the environment associated with the resulting document of a [=navigation request=].
```
yoshisatoyanagisawa: Especially for Chrome, I did not know #1267 when I worked for PlzDedicatedWorker or so, and I intentionally plumbed clientId for navigation request.  I am not so convinced the clientId would be an empty string.
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/service_worker/service_worker_global_scope.cc;l=1554;drc=9437ad2580d9d89469a5f572fef982ab6b245115 (it guarantees resultingClientId will be an empty string for subresource, but clientId might be set for non-subresource cases)

index.bs
```diff
@@ -3410,9 +3414,9 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
               1. Initialize |e|’s {{FetchEvent/request}} attribute to |requestObject|.
               1. Initialize |e|’s {{FetchEvent/preloadResponse}} to |preloadResponse|.
-              1. Initialize |e|'s {{FetchEvent/clientId}} attribute to |client|'s [=environment/id=].
-              1. If |request| is a <a>non-subresource request</a> and |request|'s [=request/destination=] is not {{RequestDestination/"report"}}, initialize |e|'s {{FetchEvent/resultingClientId}} attribute to |reservedClient|'s [=environment/id=], and to the empty string otherwise.
-              1. If |request| is a <a>navigation request</a>, initialize |e|'s {{FetchEvent/replacesClientId}} attribute to |request|'s [=request/replaces client id=], and to the empty string otherwise.
+              1. If |client| is not null, initialize |e|'s {{FetchEvent/clientId}} attribute to |client|'s [=environment/id=].
+              1. If |request| is a <a>non-subresource request</a>, |request|'s [=request/destination=] is not {{RequestDestination/"report"}}, and |reservedClient| is not null, initialize |e|'s {{FetchEvent/resultingClientId}} attribute to |reservedClient|'s [=environment/id=].
```
yoshisatoyanagisawa, line 3418: Is it relevant if the `resultingClientId` is an empty before this step?

monica-ch: Yes the step relies on it. resultingClientId is guaranteed to be "" before this step (IDL default FetchEventInit.resultingClientId = "" and the attribute dfn "must be initialized to the empty string" on event creation). That's why the step only conditionally sets a non-empty value, the "empty otherwise" case is covered by the pre-step default, so no explicit else clause is needed. Same pattern for clientId and replacesClientId.

index.bs
monica-ch: Thanks @yoshisatoyanagisawa, you're right, my earlier text was too narrow. 

For context, Gecko and WebKit currently leave clientId empty for navigation, but [Gecko's own code](https://searchfox.org/firefox-main/source/dom/serviceworkers/ServiceWorkerOp.cpp) has a TODO calling this a temporary limitation (bug 1487534)  so the intended direction across engines is Blink's. Aligning the spec there.

Changes pushed yesterday on fetch-event-client-id-fixes:

Algorithm step: dropped the subresource restriction now reads "If client is not null, initialize e's clientId to client's environment id."
clientId Note: rewritten to say clientId reflects the initiating client whenever one exists, empty only when request's client is null.
Minor editorial cleanup on the clientId and resultingClientId Notes.

resultingClientId and replacesClientId algorithm steps untouched they already match Blink's behavior per your description. PTAL.

yoshisatoyanagisawa (approved):
lgtm!
Thank you.

index.bs
yoshisatoyanagisawa: lgtm

Issue #1841, opened 2026-08-04
Use a per-cache dedicated parallel queue for Cache API methods


monica-ch:
Follow-up to #1755 (see [split plan comment](https://github.com/w3c/ServiceWorker/pull/1755#issuecomment-4963401716)) and [threading-model write-up](https://github.com/w3c/ServiceWorker/pull/1755#issuecomment-2661941708) (item 4: "Use parallel queues appropriately to avoid race conditions from separate 'in parallel' situations").

#1838 does a dedicated parallel queue on the [=name to cache map=] so [CacheStorage.match](vscode-file://vscode-app/c:/Users/monicach/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/has/open/delete/keys no longer race on the shared map. The sibling problem on each individual [=cache=] object is not yet addressed. @yoshisatoyanagisawa flagged this during #1755 review:

The situation should also be the same for the [=cache=] object. I guess each [=cache=] object will have the dedicated parallel queue, and match(), matchAll(), add(), addAll(), put(), delete(), and keys() might also executed within the parallel queue for the [=cache=] object.

#1837 does the queue-a-task-for-resolve pattern for these methods but does not add a parallel queue.

**Proposal**
Give each [=cache=] object its own dedicated parallel queue and route the following algorithms through it, so reads and writes on the same cache are serialized:

{{Cache/match(request, options)}}
{{Cache/matchAll(request, options)}}
{{Cache/add(request)}}
{{Cache/addAll(requests)}}
{{Cache/put(request, response)}}
{{Cache/delete(request, options)}}
{{Cache/keys(request, options)}}

**Related**
#1755 — parent (queue-a-task refactor)
#1740 — original missing-tasks-in-parallel issue
#1831 — cache/cache-storage forgot to queue a task to resolve promise
#1838 — sibling PR (CacheStorage-level parallel queue)
#1172 — umbrella "Carefully audit all uses of 'in parallel' in the spec"

monica-ch:
FYI @yoshisatoyanagisawa

PR #1836, opened 2026-07-16
Editorial: queue a task in Client / Clients API

monica-ch:
Updates to:
- [client-postmessage-options](https://w3c.github.io/ServiceWorker/#dom-client-postmessage) — queue tasks to fire `messageerror` and dispatch `message` on the target client's responsible event loop.
- [clients-get](https://w3c.github.io/ServiceWorker/#dom-clients-get) — restructured per @yoshisatoyanagisawa's suggestion to queue once and run "invoke Resolve Get Client Promise" + "abort" as substeps; also queues the fall-through resolve.
- [clients-claim](https://w3c.github.io/ServiceWorker/#dom-clients-claim) — queue the final resolve with undefined on the promise's responsible event loop.

All of these resolved/rejected promises or fired events directly from an "in parallel" block, which violates the "don't touch JS objects from parallel" rule.

This is part 4/6 of #1740. Split out from #1755 for focused review.

***
<a href="https://pr-preview.s3.amazonaws.com/monica-ch/ServiceWorker/pull/1836.html" title="Last updated on Jul 16, 2026, 9:02 PM UTC (284e9d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1836/e91ddff...monica-ch:284e9d0.html" title="Last updated on Jul 16, 2026, 9:02 PM UTC (284e9d0)">Diff</a>

PR #1842, opened 2026-08-04
Editorial: use a dedicated parallel queue for the Clients API

### LINKED ISSUE #1840: Use a dedicated parallel queue for the Clients API
Follow-up to #1755 (see [split plan comment](https://github.com/w3c/ServiceWorker/pull/1755#issuecomment-4963401716)) and [threading-model write-up](https://github.com/w3c/ServiceWorker/pull/1755#issuecomment-2661941708) (item 4: "Use parallel queues appropriately to avoid race conditions from separate 'in parallel' situations").

#1836 landed the queue-a-task fix for the promise-resolution side of the Clients API, but the algorithms still use generic [=in parallel=] on the browser-process side. @yoshisatoyanagisawa pointed this out during #1755 review:

For Client API, it might affect not only service worker but also clients. Do we need a parallel queue that covers both instead of the event loop for ServiceWorker itself?

I just wondered what happens if one of the service worker clients has been removed or gets execution ready flag during the sub step execution, and suggest to run Clients API algorithm within the dedicated parallel queue to prevent unexpected modifications to clients.

**Proposal**
Introduce a dedicated parallel queue for the Clients API and route the following algorithms through it, so iteration over [=service worker clients=] cannot race with client removal or [=environment/execution ready flag=] flips:

{{Clients/get(id)}}
{{Clients/matchAll(options)}}
{{Clients/openWindow(url)}}
{{Clients/claim()}}
{{Client/postMessage(message, options)}}
{{WindowClient/focus()}}
{{WindowClient/navigate(url)}}

**Related**
#1755 — parent (queue-a-task refactor)
#1740 — original missing-tasks-in-parallel issue
#1172 — umbrella "Carefully audit all uses of 'in parallel' in the spec"

monica-ch:
Closes #1840. Follow-up to #1755.

Give each {{Clients}} object a dedicated parallel queue and route the operations on {{Clients}} through it, so operations on the same {{Clients}} object do not race with each other (e.g. `matchAll` iterating over service worker clients while `claim` is mutating their [=active service worker=]).

## Changes

- Add a `<dfn export for="Clients">Clients parallel queue</dfn>` (a [=parallel queue=]) attached to each {{Clients}} object, following the same pattern as the `name to cache map parallel queue` introduced in #1838.
- Route the following method algorithms through it, replacing their existing top-level `Run … in parallel` block:
    - {{Clients/get(id)}}
    - {{Clients/matchAll(options)}}
    - {{Clients/openWindow(url)}}
    - {{Clients/claim()}}
- Wrap the two remaining bare `Resolve |promise| with undefined.` steps (in `get()` and `claim()`) in `Queue a task` on `|promise|`'s [=responsible event loop=] using the [=DOM manipulation task source=], matching the pattern already used in `matchAll()` and consistent with the queue-a-task refactor from #1755.

Total change: **+10 / −6** in `index.bs`, one commit.

## Rationale

Yoshi flagged this concern during #1755 review:

> I just wondered what happens if one of the service worker clients has been removed or gets execution ready flag during the sub step execution, and suggest to run Clients API algorithm within the dedicated parallel queue to prevent unexpected modifications to clients.

#1836 landed the queue-a-task-for-resolve fixes for these methods. This PR delivers the sibling parallel-queue work that was explicitly deferred in the [split plan](https://github.com/w3c/ServiceWorker/pull/1755#issuecomment-4963401716).

## Out of scope

- {{Client/postMessage(message, options)}} — lives on {{Client}}, not {{Clients}}, and does a single-client lookup rather than iterating the full client list. Can be a further follow-up if needed.
- {{WindowClient/focus()}} and {{WindowClient/navigate(url)}} — already skip the `in parallel` block entirely and use `Queue a task` on the client's own event loop, so no parallel-queue treatment is meaningful for them.

## Related

- #1755 — parent (queue-a-task refactor)
- #1836 — Clients API queue-a-task-for-resolve (merged)
- #1740 — original missing-tasks-in-parallel issue
- #1172 — umbrella "Carefully audit all uses of 'in parallel' in the spec"
- #1838 — sibling PR (CacheStorage-level parallel queue), same dfn pattern
- #1841 — sibling PR (per-cache parallel queue), same dfn pattern

***
<a href="https://pr-preview.s3.amazonaws.com/monica-ch/ServiceWorker/pull/1842.html" title="Last updated on Aug 18, 2026, 9:16 PM UTC (73a7c0b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1842/9e9fe27...monica-ch:73a7c0b.html" title="Last updated on Aug 18, 2026, 9:16 PM UTC (73a7c0b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1386,13 +1386,22 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     The user agent *must* create a {{Clients}} object when a {{ServiceWorkerGlobalScope}} object is created and associate it with that object.
 
+    The user agent has an associated <dfn>service worker clients parallel queue map</dfn>, a [=map=] whose [=map/keys=] are [=storage key=]s and whose [=map/values=] are [=parallel queue=]s. Each such [=parallel queue=] is used to serialize {{Clients}} operations that access [=/service worker clients=] with the corresponding [=storage key=].
+
+    The <dfn>service worker clients parallel queue</dfn> for a [=storage key=] |key| is the result of running the following steps:
+
+        1. If [=service worker clients parallel queue map=][|key|] does not [=map/exist=], then [=map/set=] [=service worker clients parallel queue map=][|key|] to a new [=parallel queue=].
+        1. Return [=service worker clients parallel queue map=][|key|].
+
+    The <dfn>relevant service worker clients parallel queue</dfn> for a {{Clients}} object is the [=service worker clients parallel queue=] for the {{Clients}} object's [=relevant global object=]'s [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=]'s [=service worker registration/storage key=].
+
     <section algorithm="clients-get">
       <h4 id="clients-get">{{Clients/get(id)}}</h4>
 
       The <dfn method for="Clients"><code>get(|id|)</code></dfn> method steps are:
 
         1. Let |promise| be a new <a>promise</a>.
-        1. Run these substeps <a>in parallel</a>:
+        1. [=queue/Enqueue=] the following steps to [=this=]'s [=relevant service worker clients parallel queue=]:
             1. For each [=/service worker client=] |client| where the result of running [=obtain a storage key=] given |client| [=storage key/equals=] the associated [=ServiceWorkerGlobalScope/service worker=]'s [=containing service worker registration=]'s [=service worker registration/storage key=]:
                 1. If |client|'s [=environment/id=] is not |id|, [=continue=].
                 1. Wait for either |client|'s [=environment/execution ready flag=] to be set or for |client|'s [=discarded flag=] to be set.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1386,20 +1385,21 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
 
     The user agent *must* create a {{Clients}} object when a {{ServiceWorkerGlobalScope}} object is created and associate it with that object.
 
+    Each {{Clients}} object has an associated <dfn>Clients parallel queue</dfn> (a [=parallel queue=]) used to serialize the {{Clients}} object's operations.
+
```
asutherland: Recognizing this patch is a WIP, I think this is currently under-constrained and it probably makes sense to do this like you've done it for the Cache API in https://github.com/w3c/ServiceWorker/pull/1838 which is basically one-per-storage-key.  (Whereas this currently seems to be 1-per-ServiceWorker instance.)

monica-ch: Agreed, 1-per-`Clients` is under-constrained. Will rework to 1-per-storage-key matching #1838.

Quick clarification before I push: #1838 hangs the queue off `[=name to cache map=]` because that map is itself per-storage-key. Clients don't have an equivalent map — the [=/service worker clients=] list is UA-wide, filtered by storage key. Would you prefer a UA-level `<dfn>Clients parallel queue</dfn>` map keyed by storage key, or a prose "for each storage key, the UA has an associated Clients parallel queue" form?

asutherland: (this got cross-referenced below, but the right answer definitely is not obvious to me, so I asked in the storage spec at https://github.com/whatwg/storage/issues/194)

monica-ch: Thanks for kicking off the discussion. Reworked per Anne's suggestion, UA now owns a map keyed by storage key, per-Clients methods enqueue via a `relevant service worker clients parallel queue` helper (mirrors #1838's [relevant name to cache map](https://github.com/w3c/ServiceWorker/pull/1838). PTAL.

Issue #1846, opened 2026-09-02
Export SW primitives needed by FedCM and Payment Handler for JIT registration


monica-ch:
FedCM (see [w3c-fedid/FedCM#842](https://github.com/w3c-fedid/FedCM/pull/842)) and Payment Handler both need to perform just-in-time (JIT) service worker registration/unregistration from their own spec algorithms. The current SW spec exposes only a handful of exported algorithms — [`Handle Fetch`](https://w3c.github.io/ServiceWorker/#on-fetch-request-algorithm), [`Fire Functional Event`](https://w3c.github.io/ServiceWorker/#fire-functional-event-algorithm), [`Match Service Worker Registration`](https://w3c.github.io/ServiceWorker/#scope-match-algorithm), [`Request Matches Cached Item`](https://w3c.github.io/ServiceWorker/#request-matches-cached-item-algorithm), and [`Lookup Race Response`](https://w3c.github.io/ServiceWorker/#lookup-race-response-algorithm). The register/update/unregister primitives are all internal, so external specs currently have no stable way to programmatically manage service worker registrations.

@yoshisatoyanagisawa flagged this on the FedCM PR ([comment](https://github.com/w3c-fedid/FedCM/pull/842#issuecomment-3159938167)), and @Brandr0id proposed three additions ([comment](https://github.com/w3c-fedid/FedCM/pull/842#issuecomment-5149038173)).

This issue tracks adding three exported primitives to the SW spec so external specifications can integrate cleanly.

### 1. Export `Soft Update`

Add the `export` attribute to the existing [`Soft Update`](https://w3c.github.io/ServiceWorker/#soft-update-algorithm) dfn, plus small clarifying Notes for external callers (fire-and-forget contract; job-equivalence collapses duplicate calls). No algorithm behavior change.

### 2. New exported `Register a Service Worker`

A thin wrapper around `Create Job` + `Schedule Job` with `job type = register`, taking (storage key, script URL, scope URL, worker type, update via cache mode). Fire-and-forget; passes `null` for both client and job promise, so the resulting job has no referrer. Includes a Note that the same-origin gate against the referrer is skipped, so callers must satisfy the same-origin precondition themselves.

### 3. New exported `Unregister Service Workers`

Bulk unregister by storage key: iterate the `registration map` for matching storage key entries, parse the serialized scope URL, and schedule an unregister job for each.

## Rationale

[...]

PR #1847, opened 2026-09-02
Add Request Soft Update algorithm for use by other specifications

monica-ch:
Part of #1846.

## Change

Introduces a new exported [`Request Soft Update`](https://w3c.github.io/ServiceWorker/#request-soft-update-algorithm) algorithm. The existing `Soft Update` remains **unexported** — external specifications must go through the wrapper.

## Design decisions

- **Storage key as input, not environment.** Yoshi's original suggestion allowed either. FedCM has no `environment settings object` (it operates from UA-side code), so `storage key` is the only workable input type.
- **"Coalesce, throttle, or ignore" language.** Explicit UA discretion. "Coalesce" allows deduplication of redundant requests; "throttle" allows rate-limiting; "ignore" is the escape hatch.
- **`Soft Update` is not exported.** Callers must go through the wrapper. This ensures the storage-key check and UA-discretion clause always apply to external calls.
- **`forceBypassCache` is optional.** It defaults to `false`, but calling specifications can request cache bypass when their use case requires it.

## Adoption implications

- Signature `(registration, requestingStorageKey, forceBypassCache = false)` becomes an implicit stability commitment.
- FedCM (w3c-fedid/FedCM#842) and Payment Handler can bind to `Request Soft Update` directly.
- UAs gain explicit authority to throttle or ignore, so implementations can add reasonable rate-limiting without violating the spec.

## Related

- #1846 — umbrella issue
- w3c-fedid/FedCM#842 — motivating consumer; confirmed compatible with this design
- #1848 — sibling PR (`Register a Service Worker`)
- #1849 — sibling PR (`Unregister Service Workers`)

***
<a href="https://pr-preview.s3.amazonaws.com/monica-ch/ServiceWorker/pull/1847.html" title="Last updated on Sep 9, 2026, 5:04 AM UTC (49711d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1847/92aba3b...monica-ch:49711d1.html" title="Last updated on Sep 9, 2026, 5:04 AM UTC (49711d1)">Diff</a>


monica-ch:
@yoshisatoyanagisawa Will you take a look at this change?

@Brandr0id FYI

yoshisatoyanagisawa:
I have two main concerns regarding exporting the `Soft Update` algorithm directly:

1. **DoS and Lifecycle Thrashing:** Exposing this allows arbitrary triggers that could lead to excessive resource usage for both the browser and the site (essentially a form of DoS or lifecycle thrashing).
2. **Privacy / Security:** We need to prevent arbitrary sites or contexts from triggering an update if they somehow obtain a registration that doesn't belong to their partition.

Because external callers like FedCM genuinely need to fetch the latest script, relying on `updateViaCache` is not a viable workaround here. 

Therefore, I suggest we introduce and export a **wrapper algorithm** (e.g., `Request Soft Update`) instead of directly exporting the raw `Soft Update`.

This wrapper algorithm could do the following:
- Require the calling `environment` (or storage key) as an argument to strictly enforce a storage key match against the registration.
- Explicitly state that the browser *may* ignore or throttle the update request. This gives User Agents the authority to prevent excessive resource consumption and mitigate DoS attacks.

What do you think?

monica-ch:
Thanks for the review — I reworked this around the `Request Soft Update` wrapper and tightened the user-agent discretion wording in d583ee4.

- The wrapper requires the requesting storage key and returns when it does not match the registration's storage key.
- The user agent may return before running `Soft Update`, allowing it to decline requests to protect against excessive resource use.
- The user agent may coalesce concurrent invocations for the same registration.
- The underlying `Soft Update` algorithm remains unexported.

I used `return` deliberately: WHATWG's algorithm-writing guidance uses it to exit an entire algorithm, whereas “abort these steps” is intended for nested or parallel substeps. This also avoids leaving “throttle” and “ignore” as undefined behavior.

FedCM can use this wrapper with its UA-derived identity-handler storage key; the request remains intentionally best-effort rather than guaranteeing that every invocation starts an update.

monica-ch:
.

> Thanks for the review — I reworked this around the `Request Soft Update` wrapper and tightened the user-agent discretion wording in [49711d1](https://github.com/w3c/ServiceWorker/commit/49711d11d01c07975febf0ec21e9c9ee7c568c0b).
> 
> * The wrapper requires the requesting storage key and returns when it does not match the registration's storage key.
> * The user agent may return before running `Soft Update`, allowing it to decline requests to protect against excessive resource use.
> * A note explains that implementations can rate-limit requests or treat several nearby invocations for the same registration as one.
> * The note also makes the best-effort contract explicit: callers cannot rely on an update check occurring.
> * The underlying `Soft Update` algorithm remains unexported.
> 
> I used `return` deliberately: WHATWG's algorithm-writing guidance uses it to exit an entire algorithm, whereas “abort these steps” is intended for nested or parallel substeps. Keeping coalescing and rate limiting in the explanatory note avoids making an undefined implementation strategy look like an executable algorithm step.
> 
> FedCM can use this wrapper with its UA-derived identity-handler storage key; the request remains intentionally best-effort rather than guaranteeing that every invocation starts an update.

Agreed and make sense. Thanks for the feedback @yoshisatoyanagisawa. I reworked this to export a  `Request Soft Update`  wrapper instead of exposing Soft Update directly.

It takes a storage key because callers such as FedCM may not have an environment available, but can supply the UA-derived storage key for the request. The wrapper enforces a match with the registration and gives the UA discretion to decline, rate-limit, or coalesce update requests. Callers therefore cannot assume every invocation triggers an update check.

@Brandr0id could you confirm that this best-effort contract works for the FedCM identity-handler flow?

index.bs
```diff
@@ -2928,6 +2928,24 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
       1. Invoke <a>Schedule Job</a> with |job|.
   </section>
 
+  <section algorithm>
+    <h3 id="request-soft-update-algorithm"><dfn export>Request Soft Update</dfn></h3>
+
+    This algorithm allows other specifications to request that the user agent check a [=/service worker registration=] for updates.
+
+      : Input
```
yoshisatoyanagisawa, line 2936: Considering the use case, you may want to have `forceBypassCache` as Input?

index.bs
```diff
@@ -2928,6 +2928,24 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      :: |registration|, a [=/service worker registration=]
+      :: |requestingStorageKey|, a [=/storage key=]
+      : Output
+      :: None
+
+      1. If |registration|'s [=service worker registration/storage key=] is not [=storage key/equal=] to |requestingStorageKey|, return.
+      1. The user agent *may* return.
```
yoshisatoyanagisawa: Expect Note here to tell what situation the browser may return here, or write the algorithm like: `If the user agent decides to limit the frequency of update checks, return.` to clarify the condition.

(Yeah, I mean that instead of having one large note below, I suggest to bring relevant part here.)

yoshisatoyanagisawa (approved):
LGTM with comments.

index.bs
monica-ch: Good point. I added  forceBypassCache  as an optional input, defaulting to false, and pass it through to  Soft Update . This preserves the default behavior while allowing callers that need it to request cache bypass.

index.bs
monica-ch: Updated. The step now explicitly returns when the user agent decides to limit update-check frequency, and I moved the relevant rate-limiting and coalescing note directly below that step.

yoshisatoyanagisawa:
still lgtm.

PR #1848, opened 2026-09-02
Add Register a Service Worker algorithm for use by other specifications

monica-ch:
Part of #1846.

This introduces a new exported algorithm — a stable public surface that external specifications can bind to.

## Change

Adds an exported [`Register a Service Worker`](https://w3c.github.io/ServiceWorker/#register-a-service-worker-algorithm) algorithm so other specifications such as FedCM and Payment Handler can register a service worker without a client-driven flow.

It also extracts the URL checks from [`Start Register`](https://w3c.github.io/ServiceWorker/#start-register-algorithm) into a shared `Validate and Normalize a Service Worker URL` algorithm. Both JavaScript registration and browser-initiated registration now use the same rules:

- remove URL fragments;
- require an HTTP or HTTPS scheme; and
- reject paths containing encoded `/` or `\` characters.

The existing JavaScript behavior is preserved: invalid URLs reject the `register()` promise with `TypeError`, and a missing scope still defaults to the script URL's directory. The exported algorithm is fire-and-forget and returns when validation fails.

## Exported algorithm

**Input:** storage key, script URL, scope URL, worker type, and update-via-cache mode  
**Output:** none

The algorithm validates and normalizes the URLs, requires the script and scope to be same-origin, checks that the script origin is potentially trustworthy, creates a register job with a null client and promise, and schedules it.

Because there is no calling service worker client, the algorithm sets the job's referrer to the normalized script URL. This allows the existing [`Register`](https://w3c.github.io/ServiceWorker/#register-algorithm) same-origin checks to remain the canonical enforcement point.

## Adoption implications

- FedCM (w3c-fedid/FedCM#842) and Payment Handler can use this instead of invoking non-exported job algorithms directly.
- The `(storageKey, scriptURL, scopeURL, workerType, updateViaCache)` signature becomes a stability commitment.
- Registration remains asynchronous and fire-and-forget; no completion or failure signal is returned.

## Related

- #1846 — umbrella issue
- w3c-fedid/FedCM#842 — motivating PR
- #1847 — `Request Soft Update`
- #1849 — `Unregister Service Workers`

***
<a href="https://pr-preview.s3.amazonaws.com/monica-ch/ServiceWorker/pull/1848.html" title="Last updated on Sep 9, 2026, 4:57 AM UTC (e053de5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ServiceWorker/1848/92aba3b...monica-ch:e053de5.html" title="Last updated on Sep 9, 2026, 4:57 AM UTC (e053de5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2777,6 +2777,29 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
   </section>
 
   <section algorithm>
+    <h3 id="register-a-service-worker-algorithm"><dfn export>Register a Service Worker</dfn></h3>
+
+    Note: Because this algorithm has no calling [=/service worker client=], the [=job/referrer=] is not set by [=Create Job=]. To satisfy the same-origin checks in the [=Register=] algorithm (steps 2 and 3), this algorithm explicitly sets the job's [=job/referrer=] to |scriptURL|. Combined with the same-origin precondition asserted in step 1, this ensures both same-origin checks in [=Register=] pass.
+
+      : Input
+      :: |storageKey|, a [=/storage key=]
+      :: |scriptURL|, a [=/URL=]
+      :: |scopeURL|, a [=/URL=]
+      :: |workerType|, a [=job/worker type=]
+      :: |updateViaCache|, a [=service worker registration/update via cache mode=]
+      : Output
+      :: None
+
+      1. Assert: |scriptURL|'s [=url/origin=] and |scopeURL|'s [=url/origin=] are [=same origin=].
+      1. If the result of running <a>potentially trustworthy origin</a> with |scriptURL|'s [=url/origin=] as the argument is <code>Not Trusted</code>, return.
+      1. Let |job| be the result of running <a>Create Job</a> with *register*, |storageKey|, |scopeURL|, |scriptURL|, null, and null.
+      1. Set |job|'s [=job/worker type=] to |workerType|.
+      1. Set |job|'s [=job/update via cache mode=] to |updateViaCache|.
+      1. Set |job|'s [=job/referrer=] to |scriptURL|.
+      1. Invoke <a>Schedule Job</a> with |job|.
+  </section>
+
+  <section algorithm>
     <h3 id="update-algorithm"><dfn>Update</dfn></h3>
 
       : Input


monica-ch:
@yoshisatoyanagisawa CYTAL, thanks!

index.bs
```diff
@@ -2776,6 +2776,29 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      :: |storageKey|, a [=/storage key=]
+      :: |scriptURL|, a [=/URL=]
+      :: |scopeURL|, a [=/URL=]
+      :: |workerType|, a [=job/worker type=]
+      :: |updateViaCache|, a [=service worker registration/update via cache mode=]
+      : Output
+      :: None
+
+      1. Assert: |scriptURL|'s [=url/origin=] and |scopeURL|'s [=url/origin=] are [=same origin=].
+      1. If the result of running <a>potentially trustworthy origin</a> with |scriptURL|'s [=url/origin=] as the argument is <code>Not Trusted</code>, return.
```
yoshisatoyanagisawa, line 2806: I assume that `Start Register` (https://w3c.github.io/ServiceWorker/#start-register) does more validation inside.  However, such validation looks not implemented here.  For example, a scheme is limited to HTTP/HTTPS, a hash will be omitted, etc.

yoshisatoyanagisawa:
Is it possible to have a common algorithm between an algorithm call via the JavaScript `register() `and the registration initiated by the browser (this case), and make the latter exported? I assume that the script URL validation can be common between the JavaScript and this.

index.bs
monica-ch: Addressed in e053de5. I extracted the fragment removal, HTTP/HTTPS scheme check, and encoded-slash path check into a shared `Validate and Normalize a Service Worker URL` algorithm. Both `Start Register` and the exported `Register a Service Worker` algorithm now use it, while `Start Register` retains its existing `TypeError` rejection and scope-defaulting behavior.

monica-ch:
Thanks, @yoshisatoyanagisawa — updated in e053de5. The URL validation and normalization previously embedded in `Start Register` is now a shared algorithm used by both the JavaScript and browser-initiated registration paths. Client-specific behavior such as promise rejection and storage-key acquisition remains in `Start Register`.

The shared helper operates on a copy of each URL, so the exported algorithm does not mutate URL records owned by its caller. PTAL when you have a chance.

index.bs
```diff
@@ -2776,6 +2784,33 @@ spec: storage; urlPrefix: https://storage.spec.whatwg.org/
...
+      :: |updateViaCache|, a [=service worker registration/update via cache mode=]
+      : Output
+      :: none
+
+      1. Set |scriptURL| to the result of running [=Validate and Normalize a Service Worker URL=] with |scriptURL|.
+      1. If |scriptURL| is failure, return.
+      1. Set |scopeURL| to the result of running [=Validate and Normalize a Service Worker URL=] with |scopeURL|.
+      1. If |scopeURL| is failure, return.
+      1. If |scriptURL|'s [=url/origin=] and |scopeURL|'s [=url/origin=] are not [=same origin=], return.
+      1. If the result of running <a>potentially trustworthy origin</a> with |scriptURL|'s [=url/origin=] as the argument is <code>Not Trusted</code>, return.
```
yoshisatoyanagisawa: nit but `Not Trusted` might be `Not Trustworthy`  now upon https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin

monica-ch:
@yoshisatoyanagisawa I've addressed the comment. Do you think it's in good shape now?

